PR #11, opened 2014-11-29
Canon PowerShot SX 260 HS ccd width

dbaldwin:


floscher:
Was it intentional, that the file was made executable (100644 → 100755) in this commit?

smathermather:
It's going to be an ongoing issue. The install script currently doesn't use a separate build directory, and also makes most everything executable, which is a bit of a problem. I'm guessing this is how it happened.

smathermather:
Hmm, looking at the install script, it no longer does this, but just runs `chmod 700 run.pl` on , so I retract the above... .

dbaldwin:
Let me know if there's something I can do differently on this end. The edit was made in TextMate on OS X. I didn't mess with permissions at all.

smathermather:
No worries. I should have checked it before I merged.

PR #41, opened 2015-01-23
Cleanup/git ignore

JesseCrocker:
Currently after running ./install the there are lots of changes to commited files, and lots of untracked files. These changes leave the repository clean after running ./install.sh
There were also several ~ files in the repository, I have deleted and ignored those.


smathermather:
Oh thank goodness. Thanks @JesseCrocker.

PR #44, opened 2015-02-03
test alternate install rules for 12.04 & 14.04

smathermather:
if [lsb_release -a | grep Release | awk '{print $2}' = "12.04" ];
...


smathermather:
Nope. Better pull request from @bgirardot...

PR #50, opened 2015-02-07
integrate OpenSfM as alternative to getkeypoints match and bundler

smathermather:
This is an optional replacement, but may become default once OpenSfM matures.


smathermather:
This is to address this item on the wishlist: https://github.com/OpenDroneMap/OpenDroneMap/issues/19

smathermather:
See:

PR #80, opened 2015-03-13
added unit "Panasonic DMC-TS4" @ 6.08 mm H

GeospatialDaryl:
lovin it


smathermather:
Dig the comment. Thanks @GeospatialDaryl.

PR #90, opened 2015-04-01
adding iPhone 5s data to ccd_defs.json

aaronpdennis:


smathermather:
Hi Aaron, does it give ok results with the iPhone 5s?

aaronpdennis:
I ran a test on it earlier today just with a few photos I took of a statue on campus and was pretty happy with the result, especially the point cloud, which I felt was very accurate.

It would be interesting to see the results with more images and from an aerial perspective. I'm planning on testing that in the next few weeks, taking some pictures with the iPhone from a drone 25-50 feet in the air. When I get those results I'll share those here, too. Below are the results from the quick test I did today.

screen shot 2015-04-01 at 3 25 11 pm
screen shot 2015-04-01 at 3 25 29 pm
screen shot 2015-04-01 at 3 24 54 pm

And in Sketchfab: https://skfb.ly/DGOT

smathermather:
Sounds great. I'll merge the pull request.

thomaspurchas:
Hey @aaronpdennis how did you calculate the CCD width for the iPhone 5S. Trying to do the samething for the iPhone 5, but I can't find any spec sheets.

aaronpdennis:
@thomaspurchas it's difficult for me to recall, I know I found the same problem... I believe I used [xiftool](https://en.m.wikipedia.org/wiki/ExifTool) to pull out the metadata tags from a sample image I took. I think the answer was in the exif tags somewhere but unfortunately I don't still have the files I worked with

PR #133, opened 2015-07-21
Update ccd_defs.json

LeithHawkins:
Adding GoPro Hero S.

Not sure wether to add S or Silver maybe ill ad both ???


LeithHawkins:
@smathermather   sorry to be a pain but can I add this for the Go pro hero as well as some of the photos are coming in as an S and some as Silver...  Thanks for providing an awesome platform :+1:

dakotabenjamin:
Added to recent merge

PR #160, opened 2015-08-27
Workflow

paulinus:
This PR changes the way tasks are called in the python script.

Right now calls are nested so that each task calls the following task before finishing.

With the changes all tasks are called from the main so that they become independent and "start-with" and "end-with" have only to be checked in one place.

### DIFF
--- a/run.py
+++ b/run.py
@@ -470,9 +470,6 @@ def resize():
                 fileObject["height"] = int(match.group(2).strip())
             print "\t (" + str(fileObject["width"]) + " x " + str(fileObject["height"]) + ")"
 
-    if args.end_with != "resize":
-        getKeypoints()
-
 
 def getKeypoints():
     """Run vlsift to create keypoint files for each image"""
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -470,9 +470,6 @@ def resize():
                 fileObject["height"] = int(match.group(2).strip())
             print "\t (" + str(fileObject["width"]) + " x " + str(fileObject["height"]) + ")"
 
-    if args.end_with != "resize":
-        getKeypoints()
-
```
dakotabenjamin, line 475: According to [PEP8](http://legacy.python.org/dev/peps/pep-0008/#blank-lines): 
Surround top-level function and class definitions with two blank lines.

paulinus, line 475: I think the two blank lines are there. At least on the new version on the right >>>

paulinus, line 475: Note the blank line in the top of the removed chunk

dakotabenjamin, line 475: Thanks, didn't notice that.

paulinus, line 475: No problem, good that you are enforcing PEP8!

dakotabenjamin:
Looks good. I'm running some quick tests now. 

On a somewhat related note, I think --run-only is non-functional.

PR #165, opened 2015-09-03
Do not change file extension when resizing images

paulinus:
odm_texture uses the list.txt file created for bundler. The file contains the list of resized images.
odm_texture uses this filenames to get the original images. If the name of the original images and the name of the resized images is not the same, odm_texture does not find the original images.

This was giving problems when the original files had any extension different than "jpg". For example "JPG".

### DIFF
--- a/run.pl
+++ b/run.pl
@@ -497,7 +497,7 @@ sub prepareObjects {
         
     foreach $fileObject (@objects) {
         if($fileObject->{isOk}){
-            $fileObject->{step_0_resizedImage}  = "$jobOptions{jobDir}/$fileObject->{base}.jpg";
+            $fileObject->{step_0_resizedImage}  = "$jobOptions{jobDir}/$fileObject->{src}";
             
             $fileObject->{step_1_pgmFile}       = "$jobOptions{jobDir}/$fileObject->{base}.pgm";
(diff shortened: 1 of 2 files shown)


paulinus:
Made a bad fix. Do not merge, please. I'm fixing it.

paulinus:
It is good to go now.

dakotabenjamin:
with run.pl we had issues with capitalized file extensions, is that fixed now in the python port?

paulinus:
I have not experimented the issue with capitalized letters, so i don't know if it is fixed.

The change is only so that `odm_texturing` get the correct filenames. That could be fixed in by other means if we want to avoid capital letters.

smathermather:
This sounds like it will address the file name extension issue. It will be interesting to see.

paulinus:
Ok, i've found where the problem with extensions other than "jpg" is. Bundler assumes files are jpg or bmp here https://github.com/snavely/bundler_sfm/blob/master/src/ImageData.cpp#L285-L296

I will let the extension change to "jpg" and fix the problem with `odm_texturing` by other means

paulinus:
I've did the fix in another way here https://github.com/OpenDroneMap/OpenDroneMap/pull/168
Closing this one.

PR #167, opened 2015-09-04
Enable the use of OpenSfM as an alternative to Bundler

paulinus:
This PR enables the use of OpenSfM for feature extraction, matching and structure from motion (bundling).

It adds the option `--use-opensfm` option to `run.py`. When used, the tasks `getKeypoints`, `match` and `bundler` are skipped and the new `opensfm` task is run.

The `install.sh` script is updated to install the Ceres solver and OpenSfM.


dakotabenjamin:
Two things: 
1. Ceres now is installed twice, once as a submodule, and once from a download. 
2. I got some runtime errors. First I got stuck printing a huge fileObject, so I commented that print line out. Then I got this error:

```
  - running OpenSfM - Fri Sep 04 15:21:18  2015
  File "/odm/src/OpenSfM/bin/import_bundler", line 6, in <module>
    import cv2
ImportError: No module named cv2

quitting cause:
        "/odm/src/OpenSfM/bin/import_bundler" opensfm --list list.txt
returned with code 256.
```

i tried `pip install cv2` but that didn't solve it.

paulinus:
1. Did not realize that ceres was in a submodule. I though ceres had been removed on previous commit. Will fix that.  Do you prefer having it as submodule or from a download?
2. Just added the missing dependencies. Thanks for checking

dakotabenjamin:
Confession: I was testing that commit I reverted on the python port, then I accidentally slipped it back in. It wasn't causing any problems, so I just left it. 

As to whether it should be a submodule or not, I'm not sure. @smathermather?

smathermather:
I don't have any strong leanings as to whether it should be a download or submodule.

paulinus:
I've let it as a download just to control which version we fetch.

smathermather:
That makes sense. Thanks!

dakotabenjamin:
run.py still crashes while printing fileObject while running vagrant.

paulinus:
That is `import cv2` halting. Let me investigate.

paulinus:
@dakotabenjamin i'm not able to reproduce the bug. It happened to me once that python halted after importing cv2. I've restarted the virtual machine and it did not happen anymore.

Since I've seen ubuntu's opencv being unstable before, I've added the installation of opencv from sources. That should fix the problem at the price of longer installation time.

dakotabenjamin:
Why is it necessary to print the fileObject on line 674? That is the specific line that is causing the problem. When run.py reaches that line, it gets stuck printing a huge object, then I cannot event use CTRL-C or any keys to cancel the process. I have to close Git Bash and ssh back into vagrant. One thing to note is that now I have a zombie process each time it happens.

paulinus:
Just deleted the print. That was just for debugging and forgot to clean it. It has no relation to the problem. Sorry.

The terminal hanging is python dying when importing cv2 in `import_bundler`.  CTRL-C does not work when python is running native libraries. It is a pain.

Did you run the install with opencv? does it still halt?

dakotabenjamin:
When I had commented out the line, everything ran fine. I'm gonna do a fresh install and one more quick run, then merge the PR if there are no issues. Thanks for all the hard work!

dakotabenjamin:
To be clear, I never had any install issues after 234f308

dakotabenjamin:
I've been stuck on building opencv for about 20 minutes. 
EDIT: opencv building took 26 minutes

dakotabenjamin:
opencv building took 26 minutes

PR #168, opened 2015-09-06
Fix name of original images for odm_texturing

paulinus:
odm_texturing was using `pmvs/list.rd.txt` to get the list of image names. This file contains the names of the renamed images, which can be different from the original ones. Thus, odm_texturing was trying to open the images from the original folder, but with the modified name.

The fix creates an `image_list.txt` file with the list of original image names.

### DIFF
--- a/run.py
+++ b/run.py
@@ -689,7 +689,13 @@ def odm_texturing():
     except:
         pass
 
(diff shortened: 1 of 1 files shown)


paulinus:
Need to fix the order when some images don't get reconstructed. Will reopen after fixing.

paulinus:
The image list with the original file names, now contains only the reconstructed images in the same order as the `pmvs/list.rd.txt`.

PR #196, opened 2015-11-03
Update opensfm

paulinus:
This updates ODM to use a recent version of OpenSfM, which improves the way radial distortion is handled and permits to match only the k-nearest images.

Install is modified to use Ubuntu's OpenCV. It is not giving problems in my tests using clean ubuntu 14.04 boxes. It also adds OpenGV which OpenSfM now depends on.


dakotabenjamin:
1. I had to do a clean install of Ubuntu or it failed. This should be noted
2. Do we know if problems occur with other versions of OpenCV being installed elsewhere (e.g. 3.0)
3. @smathermather and I discussed making OpenSfM the default over bundler, with the possibility of removing bundler completely. How do you feel about that?

dakotabenjamin:
The large datasets are processing completely. I'm pretty happy with this.

paulinus:
@dakotabenjamin for 1 and 2 i think we need a better way of installing things than a single shell script. For 3, no problem making it the default. Let's try it for a while to see how it compares with bundler in different datasets.

Issue #204, opened 2015-11-12
T/F parameters always set to True


dakotabenjamin:
run the following: 

```
vagrant@vagrant-ubuntu-trusty-64:/vagrant_data/odm_data/seneca$ /odm/run.py --resize-to 1600 -e cmvs --use-opensfm Fals
e --odm_georeferencing-useGcp False

  - configuration:
{'start_with': 'resize', 'end_with': 'cmvs', 'force_focal': None, 'odm_texturing_textureWithSize': 3600, 'matcher_thresh
old': 2.0, 'cmvs_maxImages': 500, 'zip_results': False, 'odm_orthophoto_resolution': 20.0, 'pmvs_threshold': 0.7, 'match
er_preselect': False, 'resize_to': 1600, 'matcher_k': 8, 'odm_meshing_solverDivide': 9, 'min_num_features': 4000, 'use_o
pensfm': True, 'pmvs_wsize': 7, 'odm_texturing_textureResolution': 4096, 'matcher_ratio': 0.6, 'odm_meshing_samplesPerNo
de': 1, 'odm_meshing_maxVertexCount': 100000, 'pmvs_level': 1, 'run_only': None, 'odm_meshing_octreeDepth': 9, 'pmvs_csi
ze': 2, 'matcher_kDistance': 20, 'odm_georeferencing_useGcp': True, 'odm_georeferencing_gcpFile': 'gcp_list.txt', 'match
er_useKnn': True, 'pmvs_minImageNum': 3, 'force_ccd': None}
```

No matter what input, the boolean parameters always set to True when called.

lupus78:
Seams like there is already the correct solution in the code:

--zip-results works as if added, means True, otherwise False. I think this should be applied for --use-opensfm and --odm_georeferencing-useGcp

```
parser.add_argument('--zip-results',
                    action='store_true',
                    default=False,
                    help='compress the results using gunzip')

parser.add_argument('--use-opensfm',
                    type=bool,
                    default=False,
                    help='use OpenSfM instead of Bundler to find the camera positions '
                         '(replaces getKeypoints, match and bundler steps)')
```

lupus78:
I have just made a pull request with a possible solution. #205 

Looks like argparse don't support the --parameter True/False format. What you can do is make a "--use-xyz" and "--dont-use-xyz" parameters.

I did this for the --use-opensfm and --use-bundler parameters. They are linked, only 1 can specified at a time and opensfm is the default.

[...]

PR #205, opened 2015-11-13
Argument parser and ccd_def_check fixes

lupus78:
Shows a nice list of Configuration parameters, something like it was in the Perl version.

Adds new argument: "--job-dir-name <string>". You can specify the output folder name and not use the default "reconstruction-with-image-size-[resize-to]".
One limitation however, it only works with ASCII characters.

Tweaked the ccd_defs_check python port to work with the pre-hook commit.

Fixes the Boolean parameters as discussed in https://github.com/OpenDroneMap/OpenDroneMap/issues/204  I'll add more detail in the issue.


lupus78:
Added 2 more commits. First solves the matcher preselect arguments as we discussed in #204 . The second fixes the start/end-with and run-only arguments.

PR #206, opened 2015-11-15
Python port

GeospatialDaryl:


smathermather:
Excited for this work, as is @dakotabenjamin.

PR #216, opened 2015-11-30
use GDAL Coordinate System format specification, replace cs2cs with gdaltransform

lupus78:
Keeps the old format "WGS84 UTM 17N" also working, but prefers the GDAL strandard CS definitions
Replaces cs2cs with gdaltransform.

### DIFF
--- a/run.py
+++ b/run.py
@@ -10,7 +10,6 @@ import subprocess
 import shutil
 import shlex
 # import collections  # Never used
-import fractions
 import argparse
 import errno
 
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
So anything from [here](http://www.gdal.org/classOGRSpatialReference.html#aec3c6a49533fe457ddc763d699ff8796) will work?

lupus78:
Yes. I would suggest to use EPSG:n codes, those are the most simple form.

dakotabenjamin:
I tested it with merkato/boruszyn_kap after changing the first line to EPSG:2180 and it worked. 
I created a gcp file for odm_data_bellus and odm_data_rvnir, both in state plane Ohio North (ft) (EPSG:3734). 

It doesn't look like the code supports altitude in feet though. Do you think that if the coordinate system is in feet in the gcp_list.txt, then the altitude should also be in feet then converted in the code, or should the gcp_list.txt standard require meters? I'm leaning towards the former.

lupus78:
Sure, the altitude has to be in the same unit as the X and Y coordinates.

dakotabenjamin:
Oh ok I see that gdaltransform handles z-coords. Perfect.

PR #219, opened 2015-12-09
Increase default min number of features to extract

dakotabenjamin:
At 4000, there were some issues that compounded into stitching problems while texturing. I tested at 8000 and 6000, and found better results.


smathermather:
:+1:

Issue #18, opened 2014-12-24
Add digital surface model output


smathermather:
Current outputs from ODM include point cloud, mesh, textured mesh, georeferenced textured mesh, and orthophoto. It'd be just dandy if the mesh could also be output as a digital surface model, as a simple raster.

lupus78:
Is there any update on this? I am keen to get an elevation raster out of ODM.

smathermather:
Sadly, no progress yet on this. But, with the python port, you'll get georeferenced point clouds from which you can create a surface model in other software.

lupus78:
Which software can do that and how?

smathermather:
QGIS & GRASS are two good candidates. I'd start with QGIS (https://qgis.org).

lupus78:
Yep, with the `option-0000_georef.ply` I had to make a CSV, with the X,Y shifted by the offsets, and I could import in into QGIS with GRASS plugin `r.in.xyz` command.

I'm thinking to add this to run.py to make that CSV.

However there is also an other format LIDAR (las) outputed by run.py, but I had no success using that yet in QGIS.

smathermather:
@lupus78, adding this CSV output would be very welcome.

KommandorKeen:
Cloud Compare with the Cloth Simulation Filter does a great job (with tweeking) of separating the ground from above ground features, then you can use the raster generation tool to create the DSM, or DEM as required.  All part of an opensource toolchain.

yjmenezes:
@KommandorKeen 

Ground point classification is a challenge. I will try Cloth Simulation, thx.
take a look: PKtools is licensed by GPL v3.
http://pktools.nongnu.org/html/md_apps.html#available_tools

KommandorKeen:
Thanks

Installed pktools on ubuntu and the QGIS plugin

Cannot seem to get the raster dataset from LAS(Z) to give any values.

Have tried with both LIDAR las and Drone photogrammetry las files.

 

Otherwise the Raster tools ad functionality to QGIS which was sorely missing.

 

Thanks,

 

Simon

yjmenezes:
@KommandorKeen 

you can run PKtools from Command line.
Check your QGIS menu Processing pktools patch. My Pktools folder is: /usr/bin

I am testing PKtools with my drone dataset.
I have a second dataset with contours compiled, by hand,  with a photogrammetric workstation.
[...]

PR #222, opened 2015-12-12
XYZ point cloud output

lupus78:
This is the CSV output I was talking about in #18 

It is done in the `odm_georeferencing` step, the output file name is `pointcloud_georef.csv`

This GRASS wiki explains how to convert it to DEM.
https://grasswiki.osgeo.org/wiki/LIDAR#Import_ASCII_text_into_a_raster_DEM

### DIFF
--- a/run.py
+++ b/run.py
@@ -12,7 +12,7 @@ import shlex
 # import collections  # Never used
 import argparse
 import errno
-
+import csv
 import knnMatch_exif
 
 # the defs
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
One way we could approach this is to specify and output type ( like `--output` or `-o csv`) so the user is not getting redundant information and to save space. The results folder can get rather huge now.

lupus78:
I'm thinking we should move the DEM into a new task `odm_terrain` (or similar name), and make that and option of that `--odm_terrain-type laz/csv`
This way one can completly skip the DEM generation if it is not needed.

dakotabenjamin:
This has been added to `edgarriba-cmake`, which will be merged soon!

PR #175, opened 2015-09-22
Python port objodm - 1st Commit of 'objRun.py'

GeospatialDaryl:
Hi-

I finished a clean commit of ojbRun.py.  In this commit:
-  Increased PEP-8 compliance for line length
-  Created the ODMJob class
  - instantiates to a single job, and __init__s a folder of jpg into
  - ODMPhoto class objects, which contain all metadata in object attributes.
-  Factored the first two 'steps'/calls into the ODMJob class as methods
-  Cleaned up some functions and a bit of organization.

The code base is cleaner and easier to read as well.

@dakotabenjamin - thanks for the feedback yesterday on the 'unable to merge' - I was having trouble with my local git install. 

ODM Dev Community - I strongly feel this OO approach is the way to go - it facilitates building a GUI and increases the modularity of individual steps in the processing chain.  I will add some additional comments to try to clarify the class structure - right now, the legacy list properties are kept using the original handles/names, and then, when parsed, added to the usually-more-descriptive object attribute name.

Keep flying them kites, and quads, and hot-air balloons-

Daryl

PR #223, opened 2015-12-14
Tweaks to run.py

lupus78:
This improves the code quality and output of run.py.
- 2 new functions print_task_header(), print_task_info(), which are printing the task messages simiallary. (We may add colors later for more readabilty)
- tweaked the run() function to only print the command if error happens, cleans up the output!
- using mkdir_p() everywhere, there we many folder created without it...
- turned the --zip-results argument into a last task. I think it's nicer like this.
- prepare_objects() optimize, got a bit faster, removed redundant and unused code bits
- tweaked the messages of what is happening right now

### DIFF
--- a/run.py
+++ b/run.py
@@ -478,10 +478,7 @@ def prepare_objects():
     jobOptions["step_3_filelist"] = jobOptions["jobDir"] + "/list.txt"
     jobOptions["step_3_bundlerOptions"] = jobOptions["jobDir"] + "/options.txt"
 
-    try:
-        os.mkdir(jobOptions["jobDir"])
-    except:
-        pass
+    mkdir_p(jobOptions["jobDir"])
 
     for fileObject in objects:
         if fileObject["isOk"]:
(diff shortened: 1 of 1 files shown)


lupus78:
pls note! this PR also contains the "XYZ point cloud output" commit, which I have submitted in #222

lupus78:
I have just pushed Object Oriented tweaks to this PR.

There are two Classes
ODMJob: this is the main class, initiated with the job folder, and each task is a method of the class
ODMPhoto: this is for the image, stores the exif infos, and makes the resizes, and other photo related stuff

I have based it on the #175, but this one applies it to the whole run.py.

dakotabenjamin:
Hey @lupus78 you should check out @edgarriba 's fork: https://github.com/edgarriba/OpenDroneMap/tree/cmake

The fork contains some massive changes to the code structure, including OO-ification. I don't mind merging this branch with the new classes, but just know that in the future those changes will be overwritten.

lupus78:
Hm... I didn't know about this fork.... should have save me some work. When will @edgarriba 's fork be merged?

dakotabenjamin:
We're working on testing now. Because of the large rework of the code structure it could take some time

PR #256, opened 2016-02-24
Add gcp_list to test dataset

Fi156:


dakotabenjamin:
I'm gonna merge this, and later we can go through and update the coords. I was gonna do it now with QGIS and Openlayers plugin, but for some reason the plugins repo is down.

dakotabenjamin:
This is taking a bit because I'm running into errors, and not sure if its on my end. 

Something to note here when using the --rerun command, I think some check needs to be in place when deciding between using a gcp file or exif tags.

Issue #237, opened 2016-01-26
ccd_defs is soon to be depreciated


dakotabenjamin:
Since we have more or less removed support for matcher/bundler (it still exists in the code, but OpenSfM is faster and better and therefore runs as the default), ccd_defs.json has been replaced with OpenSfM's own [sensor data file](https://github.com/mapillary/OpenSfM/blob/master/opensfm/data/sensor_data.json).

I will not be merging any ccd_defs.json update PRs anymore.

renzoepolo:
Micasense RedEdge Size CCD?

dakotabenjamin:
I don't know how a multispectral camera would work with ODM. I think you'll have to get in contact with Micasense to get more information on their sensors.

smathermather:
RedEdge has 5 bands, correct? ODM assumes 3 bands at this time, but we're open to contributions which extend that to multispectral.

renzoepolo:
Yes, RedEdge has 5 band.
I going to work with one band each time.

dakotabenjamin:
That being said, I don't think OpenSfM actually 'fails' when there's no ccd width information.

PR #287, opened 2016-04-08
Modify Camera option

JangWooCheon:
About modifying camera ( Sony ILCE-7M2) sensor size


dakotabenjamin:
You are modifying depreciated code that will be removed in the next update. See #237

smathermather:
@dakotabenjamin -- should we include in current code though (with corrections -- looks like there are some typos)?

dakotabenjamin:
No because the current code uses the mapillary json file, unless for some reason you need to use bundler.

smathermather:
:+1:

PR #288, opened 2016-04-17
Altered to work with Pictures without these tags.

Fi156:
From the tutorial of pyexif2:

As a handy shortcut, you can always assign a value for a given key regardless of whether it’s already present in the metadata. If a tag was present, its value is overwritten. If the tag was not present, one is created and its value is set:

> > > metadata[key] = value

http://tilloy.net/dev/pyexiv2/tutorial.html


dakotabenjamin:
Do you mind pushing this to the dev branch?

Fi156:
Thanks for pointing out ;-)

Issue #265, opened 2016-02-29
[edgarriba-cmake] Use of LD_LIBRARY_PATH system variable


dakotabenjamin:
is [generally not recommended](http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html) for production. How should we load the libraries properly?

smathermather:
Now that we have cmake, do we do a full install on everything? Or, if we are trying to co-exist with other installs, e.g. of OpenCV, etc., what is our best approach?

GeospatialDaryl:
It may make sense to leave a default configuration of 'install everything'
for the container folks.  It should be pretty trivial to edit the local
cmake cache to configure for local installs.

Daryl

On Mon, Feb 29, 2016 at 2:30 PM, Stephen Mather notifications@github.com
wrote:

> Now that we have cmake, do we do a full install on everything? Or, if we
> are trying to co-exist with other installs, e.g. of OpenCV, etc., what is
> our best approach?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OpenDroneMap/OpenDroneMap/issues/265#issuecomment-190347375
> .

## 

---

Daryl Van Dyke
GIS Analyst
Klamath Strategic Habitat Conservation Team
US Fish & Wildlife Service - AFWO, R8
(707) 825-5153
https://github.com/GeospatialDaryl

dakotabenjamin:
Do you mind taking a look at it? This is my first project that uses cmake, so I'm not too savvy. 

The relevant files should be here:
https://github.com/OpenDroneMap/OpenDroneMap/blob/edgarriba-cmake/SuperBuild/cmake/External-Ecto.cmake
https://github.com/OpenDroneMap/OpenDroneMap/blob/edgarriba-cmake/SuperBuild/CMakeLists.txt

GeospatialDaryl:
Hi Dakota-

I absolutely would/will - but I'm at a training in W VA and have no access
to a linux box.  I'll take a look but not sure I can work it through
without being able to sandbox and test.

I'll follow up - Daryl

On Mon, Feb 29, 2016 at 3:49 PM, Dakota Benjamin notifications@github.com
wrote:

> Do you mind taking a look at it? This is my first project that uses cmake,
> so I'm not too savvy.
> 
> The relevant files should be here:
> 
> https://github.com/OpenDroneMap/OpenDroneMap/blob/edgarriba-cmake/SuperBuild/cmake/External-Ecto.cmake
[...]

PR #300, opened 2016-05-12
Bash wrapper with environment variables

deltafactory:
Possible fix for #265.


dakotabenjamin:
So simple, I like it.

PR #344, opened 2016-07-19
Fixed link in README.md

dthgeek:
Fixed the link for OpenAerialMap so that it goes to https://openaerialmap.org/ and not to http://opentopography.org/


dakotabenjamin:
Fixed this in the dev branch: https://github.com/OpenDroneMap/OpenDroneMap/blob/dev/README.md#examples

PR #351, opened 2016-08-01
Create build.sh

pierotofy:
Added a script to make building process easier (commands are the same as "Building OpenDroneMap using git" in the readme).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+RUNPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export PYTHONPATH=$RUNPATH/SuperBuild/install/lib/python2.7/dist-packages:$RUNPATH/SuperBuild/src/opensfm:$PYTHONPATH
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$RUNPATH/SuperBuild/install/lib
+bash configure.sh
+mkdir build && cd build && cmake .. && make && cd ..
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
This seems redundant: https://github.com/OpenDroneMap/OpenDroneMap/blob/dev/configure.sh#L80 has the local module compilation, and the export calls will only be contained within the shell created to run build.sh. 

What is your reasoning for including this file?

pierotofy:
I was thinking it could make the build process easier for new users, people have to manually copy/paste the exports directive prior to running configure.sh.

Instead of typing:

```
cd path/to/odm/dir
git clone https://github.com/OpenDroneMap/OpenDroneMap.git .
export PYTHONPATH=$PYTHONPATH:`pwd`/SuperBuild/install/lib/python2.7/dist-packages:`pwd`/SuperBuild/src/opensfm
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:`pwd`/SuperBuild/install/lib
bash configure.sh
```

People could do:

```
cd path/to/odm/dir
git clone https://github.com/OpenDroneMap/OpenDroneMap.git .
bash build.sh
```

Oh, I didn't see that configure.sh was also invoking the cmake && make commands. I removed those from build.sh.

dakotabenjamin:
I agree that we should make it easier for new users. If you look at the README in dev now you will see that it already has been simplified:

https://github.com/OpenDroneMap/OpenDroneMap/tree/dev#build-opendronemap

boils down to: 

```
git clone https://github.com/OpenDroneMap/OpenDroneMap.git
cd OpenDroneMap
bash configure.sh
```

Then you can either write the exports to bashrc or just use `run.sh`

pierotofy:
Did not see that; ok, so it looks like this is redundant. I'll close this pull request. Thank you for the explanations!

Issue #352, opened 2016-08-03
XMP error running with docker


sear:
Hi, 

 I installed the docker image to use on a mac, then tested with a sample dataset (odm_data_benchmark) which worked fine. After that I tried with our own dataset and i got into this error:
$ docker run -v $IMAGES:/images opendronemap/opendronemap
[INFO]    Initializing OpenDroneMap app - Wed Aug 03 18:54:10  2016
[INFO]    Running ODM Load Dataset Cell
[DEBUG]   Loading dataset from: /images/images
Traceback (most recent call last):
  File "/code/run.py", line 38, in <module>
    plasm.execute(niter=1)
  File "/code/scripts/dataset.py", line 57, in process
    self.params.force_ccd)
  File "/code/opendm/types.py", line 32, in **init**
    self.parse_pyexiv2_values(self.path_file, force_focal, force_ccd)
  File "/code/opendm/types.py", line 62, in parse_pyexiv2_values
    val = metadata[key].value
  File "/usr/lib/python2.7/dist-packages/pyexiv2/xmp.py", line 210, in _get_value
    self._compute_value()
  File "/usr/lib/python2.7/dist-packages/pyexiv2/xmp.py", line 200, in _compute_value
    self._value = self._convert_to_python(self._raw_value, self.type[17:])
  File "/usr/lib/python2.7/dist-packages/pyexiv2/xmp.py", line 374, in _convert_to_python
    raise NotImplementedError('XMP conversion for type [%s]' % type)
NotImplementedError: XMP conversion for type [seq Integer]

i searched this forum for something similar   but found nothing. One thread suggested to adjust FocalLength in the file using exiftool, which i did and gives the same error. I also tried using --force-ccd 6.17 --force-focal 3.5 and combinations of this two parameters and nothing. 

I don't have the drone model as this dataset was given to me to experiment with this software, but by the exif info i can know the camera model (FC300X). Can anyone point some clue on how this could be fixed?

cheers

Sergio

dakotabenjamin:
You get the same error output when not using --force-focal?

Can you upload a sample image?

sear:
yes the same result with or without any parameters. Here's a sample image.
![dji_0597](https://cloud.githubusercontent.com/assets/1748336/17411807/99257b96-5a2e-11e6-832a-d2dd1fe1ab2c.jpg)

dakotabenjamin:
[...]

PR #372, opened 2016-09-12
Add exception for implementation error in parsing exif

dakotabenjamin:
This PR fixes an error raised when certain EXIF tags are not able to be parsed by pyexiv2. See: #352


sethfitz:
Perfect. The failures aren't fatal, as those tags should be ignored anyway.

PR #376, opened 2016-09-16
Generate tiled, compressed GeoTIFFs

sethfitz:
This allows more efficient access when reading (particularly over a network), as block-level reads become possible. The tile sizing is relatively arbitrary (though larger than the default value of `128`).

Sparse files, band interleaving, and JPEG compression reduce the file size of the resulting images.

### DIFF
--- a/scripts/odm_orthophoto.py
+++ b/scripts/odm_orthophoto.py
@@ -89,13 +89,13 @@ class ODMOrthoPhotoCell(ecto.Cell):
                 }
 
                 system.run('gdal_translate -a_ullr {ulx} {uly} {lrx} {lry} '
-							'-co TILED=yes'
-							'-co COMPRESS=DEFLATE'
-							'-co PREDICTOR=2'
-							'-co BLOCKXSIZE=256'
-							'-co BLOCKYSIZE=256'
-							'-co INTERLEAVE=band'
-							'-co NUM_THREADS=ALL_CPUS'
+							'-co TILED=yes' 
+							'-co COMPRESS=DEFLATE' 
+							'-co PREDICTOR=2' 
+							'-co BLOCKXSIZE=256' 
+							'-co BLOCKYSIZE=256' 
+							'-co INTERLEAVE=band' 
+							'-co NUM_THREADS=ALL_CPUS' 
                            '-a_srs \"EPSG:{epsg}\" {png} {tiff} > {log}'.format(**kwargs))
                 geotiffcreated = True
             if not geotiffcreated:
--- a/scripts/odm_orthophoto.py
+++ b/scripts/odm_orthophoto.py
@@ -89,6 +89,9 @@ class ODMOrthoPhotoCell(ecto.Cell):
                 }
 
                 system.run('gdal_translate -a_ullr {ulx} {uly} {lrx} {lry} '
+                           '-co TILED=yes -co SPARSE_OK=yes '
+                           '-co BLOCKXSIZE=256 -co BLOCKYSIZE=256 '
+                           '-co INTERLEAVE=BAND -co COMPRESS=JPEG '
                            '-a_srs \"EPSG:{epsg}\" {png} {tiff} > {log}'.format(**kwargs))
                 geotiffcreated = True
             if not geotiffcreated:
--- a/scripts/odm_orthophoto.py
+++ b/scripts/odm_orthophoto.py
@@ -92,6 +92,7 @@ class ODMOrthoPhotoCell(ecto.Cell):
                            '-co TILED=yes -co SPARSE_OK=yes '
                            '-co BLOCKXSIZE=256 -co BLOCKYSIZE=256 '
                            '-co INTERLEAVE=BAND -co COMPRESS=JPEG '
+                           '-co JPEG_QUALITY=90 '
                            '-a_srs \"EPSG:{epsg}\" {png} {tiff} > {log}'.format(**kwargs))
                 geotiffcreated = True
             if not geotiffcreated:
--- a/scripts/odm_orthophoto.py
+++ b/scripts/odm_orthophoto.py
@@ -89,13 +89,13 @@ class ODMOrthoPhotoCell(ecto.Cell):
                 }
 
                 system.run('gdal_translate -a_ullr {ulx} {uly} {lrx} {lry} '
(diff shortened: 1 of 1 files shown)


smathermather (commented):
Good call on the tiling. Sparce_ok is probably fine (nice storage saving feature, and I think I'm OK with it being GDAL only), but I'm not sure about jpeg compress.

sethfitz:
I was originally using `DEFLATE` (as with the DEM data I've been working with recently) until I realized that a) this is imagery, b) generated from JPEGs (usually), and c) lossy, since it's assembled from point clouds.

Happy to make the change, but that there is my thought process ;-)

sethfitz:
Yeah, I'm sold on `DEFLATE`, if for no other reason than it handling transparency better.

scripts/odm_orthophoto.py
```diff
@@ -89,6 +89,13 @@ def process(self, inputs, outputs):
                 }
 
                 system.run('gdal_translate -a_ullr {ulx} {uly} {lrx} {lry} '
+							'-co TILED=yes'
```
sethfitz: These all need spaces at the end of the line so that they're not squashed together.

smathermather:
Hey @dakotabenjamin -- changed this to use deflate (instead of jpg) etc.. Typically could reduce 2GB output to ~800GB. Can you test and merge?

dakotabenjamin:
I hope you mean MB! will do

dakotabenjamin:
```

[INFO]    Creating GeoTIFF
[DEBUG]   running gdal_translate -a_ullr 274798.508987 4603435.32167 275080.535751 4603076.29147 -co TILED=yes -co COMPRESS=DEFLATE -co PREDICTOR=2 -co BLOCKXSIZE=256 -co BLOCKYSIZE=256 -co INTERLEAVE=band -co NUM_THREADS=ALL_CPUS -a_srs "EPSG:32617" /home/dmb2/opendronemap/tests/test_data/odm_orthophoto/odm_orthophoto.png /home/dmb2/opendronemap/tests/test_data/odm_orthophoto/odm_orthophoto.tif > /home/dmb2/opendronemap/tests/test_data/odm_orthophoto/gdal_translate_log.txt
Warning 6: Driver GTiff does not support NUM_THREADS creation option
```

dakotabenjamin:
Looks good, once we address the warning.

sethfitz:
The warning goes away w/ GDAL 2.1+; `NUM_THREADS` becomes a valid creation option: http://www.gdal.org/frmt_gtiff.html

jjmata:
Since "there are no dumb questions" in life :thinking: ... would this sort of GDAL tiling be a workable way to generate `{z}/{x}/{y}.png`-type tiles for different zoom levels to serve via Leaflet?  For starters, how do we export the JPEG tiles from the TIFF wrapper?

I assume a few separate calls would have to be made to generate the different levels of detail, but I could be so off-base as a newcomer to this field to be laughably wrong.

/cc Not sure if this is what you had in mind the other day when you mentioned this was somehow possible, @javisantana 😛

smathermather:
This plus pyramid layers (i.e. `gdaladdo`) gives the pre-cursors for a tile set. @mojodna has spent some time on building a tiler on top of such geotiffs.

jjmata:
The `Driver GTiff does not support NUM_THREADS creation option` warning currently breaks node-OpenDroneMap which @pierotofy's work on WebODM still uses.  Was the "handling" you make mention of on Sep 26 ever implemented, @dakotabenjamin?  Or simply waiting for a new GDAL?

I assume the exit code changed, not sure.

**UPDATE**: sorry, spoke too soon ... I removed [the offending option](https://github.com/OpenDroneMap/OpenDroneMap/blob/dev/scripts/odm_orthophoto.py#L98) and still get a "failure" (most assets seem to be generated) from node-OpenDroneMap, will investigate more.

**FINAL UPDATE**: in case anybody runs into my situation (weird - but you never know) of thinking the `NUM_THREADS` problem is causing an issue when running node-ODM, make sure you have `python-gdal` installed!

sethfitz:
@jjmata it should just be a warning prior to GDAL-2.0 (at which point it triggers behavior that we want), so the exit code should be the same.

sethfitz:
@jjmata I realized that I forgot to respond to your not-dumb question. "Internal tiling" means that the internal structure of the TIFF is laid out in NxN blocks rather than scanlines. This means that adjacent pixels within a region can be read more efficiently by reading blocks/tiles that contain pixels of interest.

I've been using rasterio (with GDAL under the hood) to generate JPEG tiles from GeoTIFFs. ["Warped VRTs"](http://erouault.blogspot.com/2014/10/warping-overviews-and-warped-overviews.html) allow the source data to remain in its native coordinate system while being reprojected to web mercator on the fly. Overviews allow lower zooms to be rendered on the fly w/o needing to load lots and lots of pixels before resampling to fit the target resolution.

The cleanest, single-purpose code-base I'm using for this at the moment is: https://github.com/mojodna/oam-dynamic-tiler

(Similar code w/o the Lambda flasks and w/ offline task processing lives in https://github.com/americanredcross/posm-imagery-api)

If someone (you?) is interested and willing, I'd love to pair with someone to extract out the common functionality into a library that can be used by the various codebases I'm (and possibly others) accumulating that do similar things.

https://github.com/openterrain/deploy/blob/master/functions/hillshade/main.py is also very similar, though it works with single band data instead and generates single-band hillshades or RGBA ramped hillshades.

For one of my earlier experiments (in production on Heroku for Field Papers), using Mapnik directly (which likely does very similar things under the hood, linked to GDAL from C++), check out https://github.com/fieldpapers/fp-tiler (It also caches files locally because I hadn't yet discovered that `libgdal.so` can be replaced with a replacement linked to `curl` to facilitate reading of remote sources.)

jjmata:
Thanks for the wonderful (as always!) explanation.

> If someone (you?) is interested and willing, I'd love to pair with someone to extract out the common functionality into a library that can be used by the various codebases I'm (and possibly others) accumulating that do similar things.

Somebody from our team (cc: @apercas since he's the most likely candidate) will definitely be interested in helping out here.  The problem is our appetites are waaaay bigger than our stomachs at the moment.  :-)

Issue #383, opened 2016-09-25
Orthos are rotated 180º when ignored photos are present in input images directory


sethfitz:
I started with a folder full of images that didn't include geotags and used `exiftool` to add appropriate location tags to them. `exiftool` helpfully created `<img>.JPG_original` backups of each image. The early stages of ODM reads the location tags appropriate (ignoring anything where the filename does not appear to be an image).

However, sometime during `odm_georeferencing` the location tags are re-read (presumably to create a transformation matrix). My guess is that all files in the directory are read and a matrix that effects a 180º rotation is created.

Deleting the `<img>.JPG_original` files from the input direction eliminates the rotation.

dakotabenjamin:
odm_georeferencing calls extract_utm with a -imageslistfile param. What is the contents of that file?

sethfitz:
Good question; it disappeared when the Docker execution completed. I will re-run and find out for you soon.

sethfitz:
It's a list of just the JPGs (no `_original`s), as one would expect, so that theory may be bust.

dakotabenjamin:
OK. It's probably some command that is ignorantly querying the whole directory. I'll start looking for it.

dakotabenjamin:
OK I think I found the problem and/or a potential solution. Lines 82- in scripts/odm_georeferencing:

```
            kwargs = {
                'bin': context.odm_modules_path,
                'bundle': tree.opensfm_bundle,
                'imgs': tree.dataset_raw,
                'imgs_list': tree.opensfm_bundle_list,
                'model': tree.odm_textured_model_obj,
                'pc': tree.pmvs_model,

```

imgs takes the images folder rather than the images_resize folder. I'll create a PR shortly, can you test?

sethfitz:
Definitely. It'll be tomorrow / overnight.

smathermather:
We're not loosing any resolution by doing this, are we?

dakotabenjamin:
Maybe we should look at #364

dakotabenjamin:
I'm considering reverting this change in favor of a more rigorous image verification.

For example, it might be a better idea to ask users to specify their images directory (`-i /path/to/images`) and then copy valid images over to a project directory (which could be the same), rather than force users to adhere to our project structure.

Bavar2142:
From an  end users perspective thatd be appreciated.

dakotabenjamin:
[...]

PR #386, opened 2016-09-28
Change image directory reference to resized rather than raw

dakotabenjamin:
@mojodna please test. Fix for #383


sethfitz:
This produced a correctly-rotated ortho, so it looks good to me.

PR #399, opened 2016-10-14
Dockerfile problem

kikislater:
Usage of sudo broke installation 
sudo: error in /etc/sudo.conf, line 0 while loading plugin `sudoers_policy'
sudo: /usr/lib/sudo/sudoers.so must be only be writable by owner
sudo: fatal error, unable to load plugins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Install dependencies
 RUN apt-get update \
-    && sudo apt-get remove libdc1394-22-dev \
+    && apt-get remove libdc1394-22-dev \
     && apt-get install -y --install-recommends \
         build-essential \
                      cmake \


kikislater:
Problem also present in configure.sh !

> Updating the system
> sudo: error in /etc/sudo.conf, line 0 while loading plugin `sudoers_policy'
> sudo: /usr/lib/sudo/sudoers.so must be only be writable by owner
> sudo: fatal error, unable to load plugins
> ERROR: When Updating the system
> The command '/bin/sh -c bash ./configure.sh &&     mkdir build && cd build && cmake .. && make && cd .. &&     chown -R odm:odm /code' returned a non-zero code: 1

Why there is a "sudo" in configure ?

sethfitz:
Woops, missed that one. Thanks for digging it up.

`sudo` is in `configure.sh` because it serves 2 purposes: 1) configure the Linux environment (which needs to be done as `root`) and 2) build the application (which can and should be done as `$USER`).

I'd welcome splitting these out, as it means that I'd be able to inject `http_proxy` (and potentially other environment variables) into the commands run as `root`.

dakotabenjamin:
This is fixed in dev. see packages.dockerfile

PR #400, opened 2016-10-14
Upgrade OpenSfM for parallel depthmap computation.

paulinus:
Just fetching a newer release.


dakotabenjamin:
Should I wait for https://github.com/mapillary/OpenSfM/pull/106?

sethfitz:
No need. An internal API will change, but nothing external is affected.

sethfitz:
@dakotabenjamin: can you update the MD5 as well? Without that, `dev` is currently broken.

PR #404, opened 2016-10-23
Invalid tags in photos will be deleted

Tomatosoup97:
Delete with exiftool tags that cant be identified and raise NotImplementedError.
Otherwise, in our case creating orthomosaic was impossible.

Treat this PR as an proposition, because i'm not sure if this match with expected behaviour.
However, in our case where we could not create imagery, it worked well.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+## Set up library paths
+RUNPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export PYTHONPATH=$RUNPATH/SuperBuild/install/lib/python2.7/dist-packages:$RUNPATH/SuperBuild/src/opensfm:$PYTHONPATH
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$RUNPATH/SuperBuild/install/lib
+
 ## Before installing
 echo "Updating the system"
 sudo apt-get update
@@ -12,7 +17,8 @@ sudo apt-get install -y -qq build-essential \
                      libgdal-dev \
                      gdal-bin \
                      libgeotiff-dev \
-                     pkg-config
+                     pkg-config \
+                     libimage-exiftool-perl
 
 echo "Getting CMake 3.1 for MVS-Texturing"
 sudo apt-get install -y software-properties-common python-software-properties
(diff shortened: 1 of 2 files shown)


pierotofy (commented):
I think the ability to process images images that have invalid tags would add to the stability of ODM, so I'd welcome this change.

I'm just a little confused as to why it's necessary to rewrite the file without the EXIF tags; shouldn't later code just ignore invalid tags?

Tomatosoup97:
Just by ignoring and passing NotImplementedError, later in execution following error occured:

``` python
[94m[INFO]    Running ODM Georeferencing Cell[0m
[93m[WARNING] Warning: No coordinates file. Generating coordinates file in: /path/test/odm_georeferencing/coords.txt[0m
[92m[DEBUG]   running /path/OpenDroneMap/build/bin/odm_extract_utm -imagesPath /path/test/images/ -imageListFile /path/test/opensfm/list_r000.out -outputCoordFile /path/test/odm_georeferencing/coords.txt -logFile /path/test/odm_georeferencing/odm_georeferencing_utm_log.txt[0m
Error in OdmExtractUtm:
basic_string::substr
```

The error is thrown by `modules/odm_extract_utm/src/UtmExtractor.cpp`
We changed the code as in PR and now it works like a charm.

pierotofy:
Let's see if @dakotabenjamin knows more about this error.

dakotabenjamin:
Sorry I am very busy this week. I won't get to this until Wednesday.

sethfitz:
Please don't modify the source images; if there are tags that aren't understood, just skip them. Keep in mind that images may be used for other purposes, so removing tags that ODM doesn't understand may impact others.

A better change would be to fix `UtmExtractor.cpp` to _also_ ignore the problematic tags.

(Also, if you are going to manipulate tags from within Python, you should use a Python library rather than shelling out.)

pierotofy:
I agree with @mojodna, fixing `UtmExtractor.cpp` would be better.

dakotabenjamin:
I agree with @mojodna and @pierotofy. I'm gonna hold off on this until after the merge, unless their proposed solution is added in the next few days.

smathermather:
New GPS referencing process that Mapillary is working on is likely to eliminate UTMExtractor.cpp anyway. We can make sure that the behavior of the new code handles tags well.

The concept will be to georeference earlier in the toolchain, and use GPS quality flags + SfM in a weighted estimate of positon which should improve accuracy and robustness.

dakotabenjamin:
Improvements have been made to ensure that original images do not get modified and that exif errors here are handled better. I'll close this PR now.

Issue #406, opened 2016-10-24
Input photo coordinates are written as UTM


sethfitz:
In [`odm_georeferencing.py`](https://github.com/OpenDroneMap/OpenDroneMap/blob/dev/scripts/odm_georeferencing.py#L120-L121), input files have their coordinates rewritten as UTM within the EXIF data.

[IMO] input files shouldn't be touched during processing, particularly as the pipeline can only handle input files with Lng/Lat coordinates.

This is the root cause of #397, as I was re-processing imagery that had already been run through.

sethfitz:
(Also volunteering to fix this one, though it's an easy removal, so if someone else gets to it first, that's cool ;-)

dakotabenjamin:
iirc the input photos are the resized photos but i'll have to double-check

dakotabenjamin:
This issue is outdated and will be closed. Please open a new thread in http://community.opendronemap.org and reference this issue if you would like to continue the conversation.

PR #411, opened 2016-10-25
Don't touch input files

sethfitz:
(They were either already geotagged or an external GCP list is available.)

A side-effect of this is that non-geotagged photos will no longer be geotagged during the process.

This allows inputs to be run through ODM repeatedly (after deleting resulting artifacts) without failures on the 2nd+ time.


sethfitz:
Fixes #406

dakotabenjamin:
I was thinking about overhauling the user interface such that the user would specify an input set of images and a project folder. Then the images are copied into the project folder and the originals are left untouched. What do you think?

sethfitz:
Sure, that would work, but I guess I still don't see the need to modify the source images.

dakotabenjamin:
Does this remove the feature to geotag images in the resize dir?

sethfitz:
Possibly, though since I couldn't find any other references to `utm_to_latlon`, I don't think that was working.

scripts/odm_georeferencing.py
```diff
@@ -117,9 +117,6 @@ def process(self, inputs, outputs):
             geo_ref = types.ODM_GeoRef()
             geo_ref.parse_coordinate_system(tree.odm_georeferencing_coords)
 
-            for idx, photo in enumerate(self.inputs.photos):
```
sethfitz, line 120: This is probably the line to change to restore geotagging of resized photos if that's not currently working.

dakotabenjamin:
I had to go back and search the source for `self.inputs.photos` but I found this: https://github.com/mojodna/OpenDroneMap/blob/dev/scripts/odm_app.py#L106
https://github.com/mojodna/OpenDroneMap/blob/dev/scripts/odm_app.py#L136

That means that the original photos in images/ is not touched AFAIK

sethfitz:
Quick way to test: make `images/*` read-only and see what fails.

In my case, the originals were definitely getting touched, preventing them from being processed multiple times.

dakotabenjamin:
Here's the error I get: 
```
[DEBUG]   lat_frac: 548229/1 23/1 56400/1000
  File "/home/dmb2/opendronemap/run.py", line 50, in <module>
    plasm.execute(niter=1)
  File "/home/dmb2/opendronemap/scripts/odm_georeferencing.py", line 121, in process
    geo_ref.utm_to_latlon(tree.odm_georeferencing_latlon, photo, idx)
  File "/home/dmb2/opendronemap/opendm/types.py", line 284, in utm_to_latlon
    metadata.write()
  File "/usr/lib/python2.7/dist-packages/pyexiv2/metadata.py", line 119, in write
    self._image._writeMetadata()
IOError: /home/dmb2/odm_boruszyn_kap/images/img_4858.jpg: Failed to open file (a+b): Permission denied (errno = 13)
```

This must be unexpected behavior due to a bug. odm_georeferencing purposefully uses images_resize as an input source. I checked where _photo.path_file (line 240 in opendm/types.py) comes from and it is definitely resize. Adding `--rerun-from odm_georeferencing` the error disappeared...

I don't think we should outright remove utm_to_latlon because I think it can be useful to have referenced images. It needs to be isolated to the resized images only though.

sethfitz:
> I don't think we should outright remove utm_to_latlon because I think it can be useful to have referenced images. It needs to be isolated to the resized images only though.

Also my thinking. I removed it here because it wasn't, (incorrectly) assuming that it was intentional behavior, not a bug.

This means that something is changing `images_resize`. I may have time to dig deeper this afternoon, but 'til then I'm all speculation ;-)

dakotabenjamin:
Is it possible that when running for the first time (when images_resize doesn't exist), somewhere the code defaults linking the image path to the originals? That would explain why the first time I got the error but on subsequent runs I didn't.

dakotabenjamin:
At your convenience could you re-point this and other open PRs to master? I want to try to delete the dev branch

sethfitz:
> Is it possible that when running for the first time (when images_resize doesn't exist), somewhere the code defaults linking the image path to the originals? That would explain why the first time I got the error but on subsequent runs I didn't.

Possible.

> At your convenience could you re-point this and other open PRs to master? I want to try to delete the dev branch

Merged this with `master`, about to do the same for the others.

dakotabenjamin:
It still says "@mojodna wants to merge 3 commits into OpenDroneMap:dev from mojodna:leave-input-pristine"

I've never tried to change the branch I'm trying to request before, maybe close and open a new one?

dakotabenjamin:
I think I found it: https://github.com/OpenDroneMap/OpenDroneMap/blob/master/scripts/dataset.py#L40
```
        # set images directory
        images_dir = tree.dataset_resize

        if not io.dir_exists(images_dir):
            images_dir = tree.dataset_raw
            if not io.dir_exists(images_dir):
                log.ODM_ERROR("You must put your pictures into an <images> directory")
                return ecto.QUIT
```

When running the first time, images_dir is tree.dataset_raw, otherwise its resize. This gets written into the path_file attribute in ODM_Photo: https://github.com/OpenDroneMap/OpenDroneMap/blob/master/opendm/types.py#L19 and https://github.com/OpenDroneMap/OpenDroneMap/blob/master/scripts/dataset.py#L12


This is loaded into odm_georeferencing here: https://github.com/OpenDroneMap/OpenDroneMap/blob/master/scripts/odm_georeferencing.py#L23 I think.

sethfitz:
Looks like I do need to replace this PR. I'll keep this open for discussion meanwhile though, since this is where all the useful info is.

dakotabenjamin:
Digging deeper: The resize script changes the ODM_Photo object `path_File` (the string containing the image path) here: https://github.com/OpenDroneMap/OpenDroneMap/blob/master/scripts/resize.py#L44

dakotabenjamin:
I think I got it. Testing now. in scripts/odm_app.py (line 142):
```python
        # create odm georeference
        connections += [self.tree[:] >> self.georeferencing['tree'],
                        self.args[:] >> self.georeferencing['args'],
                        self.resize['photos'] >> self.georeferencing['photos'],
                        self.texturing['reconstruction'] >> self.georeferencing['reconstruction']]
```
change self.dataset['photos'] to self.resize['photos']

dakotabenjamin:
seems to have worked. I'll submit a new PR and can you test?

dakotabenjamin:
@mojodna can we close this?

PR #413, opened 2016-10-25
Subprocesses

### LINKED ISSUE #405: ⌃C only takes effect after stages complete
When hitting ⌃C to cancel a running job, it continues running until the active stage completes (rather than quitting immediately). I suspect that this is due to the use of `os.system()` in [`opendm.system.run`](https://github.com/OpenDroneMap/OpenDroneMap/blob/dev/opendm/system.py#L22) rather than the `subprocess` (or similar), which has improved handling of subprocesses.

sethfitz:
This is a first step toward improving interruption / signal handling behavior (#405). It also fixes up the log functions to preserve their `(str, *args, **kwargs)` signature.

When interrupting during the `opensfm` step, orphaned processes are left running, for example, but that may be deeper down the stack.

Fixes #405


dakotabenjamin:
screenshot from 2016-11-23 14-02-36

Is this expected? I'm now stuck in this terminal. EDIT: I ^c during opensfm

sethfitz:
It is. It should eventually quit once each of the multiprocessing tasks completes (still before it would have previously).

This SO issue kind of covers the reason; it's something we can continue to improve based on the notes in that thread, though at the expense of readability.

dakotabenjamin:
I merged this into master by hand: https://github.com/OpenDroneMap/OpenDroneMap/commit/dcb294b49c050ad2bbc31c29a22f4c9f03c3d616

sethfitz:
I'm sorry, I forgot to recreate this. Thank you.

PR #428, opened 2016-11-30
Added dockerfile for march=core2 compilation

pierotofy:
This PR aims to introduce an alternate Dockerfile for building OpenDroneMap and its dependencies using the "-march=core2" flag in all gcc/g++ invocations. This alternative Dockerfile can be used to create portable docker images (images that can be run on a wide range of intel based CPUs). The standard build process optimizes binaries to run on the CPU for which the image is built against, which results in non-portable images. This problem is evident if the docker image is downloaded on a computer older than the one used to build the image (see https://github.com/pierotofy/node-OpenDroneMap/issues/1).

As the number of dependencies is quite large, it would be unfeasible to manually patch each Makefile; instead I replace the gcc/g++ programs with a script that adds/replace the -march directive to use core2 (see https://gcc.gnu.org/onlinedocs/gcc-4.5.3/gcc/i386-and-x86_002d64-Options.html for other options).

People can still build the image and exploit all available CPU instruction sets using the current Dockerfile.

### DIFF
--- a/core2.Dockerfile
+++ b/core2.Dockerfile
@@ -1,0 +1,36 @@
+#Pull in previously built packages image with lots of libraries.
+FROM packages
+
+# Prepare directories
+RUN mkdir /code
+WORKDIR /code
+
+# Copy repository files
+COPY ccd_defs_check.py /code/ccd_defs_check.py
+COPY CMakeLists.txt /code/CMakeLists.txt
+COPY configure.sh /code/configure.sh
+COPY /.git/ /code/.git/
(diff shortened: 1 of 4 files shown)


dakotabenjamin:
Can you offer instructions for this build method?

pierotofy:
Sure, it's just like a normal docker build, just need to use the core2.Dockerfile...

`docker build -t opendronemap/opendronemap:latest -f core2.Dockerfile .`

dakotabenjamin:
Great! Thanks.

PR #443, opened 2016-12-21
Remove CMVS/PMVS from pipeline

dakotabenjamin:
In favor of OpenSfM dense point cloud.


xialang2012:
I think it might be better to keep the PMVS as a choice, and use the OpenSfM dense as the default dense point construction tool.

dakotabenjamin:
Closing in favor of #444

PR #444, opened 2016-12-27
Change OpenSfM to default for point clound densification

dakotabenjamin:
Alternative to #443 

DO NOT MERGE BOTH.


pierotofy:
Is there a tradeoff between CMVS/PMVS and OpenSFM? Speed vs density?

dakotabenjamin:
IIRC OpenSfM is about 3x more dense at a cost of speed (I do not yet know the exact difference in time costs), but its more complicated than that. PMVS/CMVS is finicky and has memory problems. Plus OpenSfM has better multi-threading. 

In my view OpenSfM is a strictly better replacement for C/PMVS. I offered two solutions as a matter of debate; perhaps users need a more gradual transition or OpenSfM needs more robust testing.

pierotofy:
It might be beneficial in the long run to stick to a single method, and if OpenSfM tends to work better, I would be in favor of merging #443.

Less dependencies will lead to more stable software and will make testing easier, as well as make it easier for users to use it (sometimes less options are better).

pierotofy:
With that said, if some people are getting errors using OpenSfM, we might want to have a transition period in which we just use set the default to use OpenSfM. The choice depends on how much testing has been done.

smathermather:
I like the idea of eliminating PMVS/CMVS for all the reasons listed above by Dakota and Piero. Whether to phase it out, or eliminate right away... I suppose we switch the defaults first and identify an end-of-life timeline and identify criteria and/or datasets that OpenSfM needs to perform adequately on for PMVS/CMVS to be fully retired.

dakotabenjamin:
I'm in favor of eliminating PMVS completely as it looks like the errors experienced and documented in #443 are either not critical or unrelated to opensfm.

pierotofy:
Let's go for it then. Hey, we can always revert it back if Armageddon happens. :+1:

pierotofy:
So I might have found a problem which is introduced by using opensfm dense reconstruction...

PMVS output:

image

OpenSfM output:

image

Notice the edges around the trees. I keep seeing those artifacts in a lot of the datasets that I have been processing.

It's obvious that OpenSfM is doing a much better job of creating a point cloud:

image

Vs. PMVS

image

I think that we need to do some type of filtering before the meshing step, in order to get a smoother, non-jagged textured orthophoto. Something I noticed by running the same dataset on mapsmadeeasy is how they "sharpen" the edges of features to get a nice looking orthophoto:

image

Compared to what we currently use:

image

Any thoughts @dakotabenjamin, @smathermather ?

smathermather:
I would anticipate this result from switching to OpenSfM from PMVS -- the point cloud from PMVS is so bad that the texturing for un-reconstructed trees can be better than trying to texture over the bubble-like mesh that results from OpenSfM because the more flat ground is a better surface to texture to minimize distortion.

What mapsmadeeasy is doing is generating what Agisoft refers to as a heightfield mesh -- it's a 2.5 D mesh that's better suited for orthophoto generation, rather than the generic poisson approach we are using in ODM. It would be interesting to see what options PCL has for alternate mesh constructions.

smathermather:
I should also say, it might be interesting to run PDAL's ground classification algorithm to filter out non-ground points prior to meshing and see what differences that results in.

smathermather:
Short and long of it, I think we should keep PMVS in for now, but default to OpenSfM. When people encounter this, then they can throw the flag. In the meantime we look for a better meshing approach.

smathermather:
One last comment: we can't do a height field mesh (easily anyway) until we integrate geo data through the tool chain so that the point cloud is already geo data and we know which way is Z. Once Pau finished this work (its mostly done), then we will have improved geolocation and the precursors for doing height field work.

pierotofy:
Good info, thanks for sharing!

I did more digging, I don't see anything in PCL to compute height fields, but I've seen similar outputs being computed using commercial software. QT Modeler calls it "Gridded Surface" computation, where options such as noise filtering, hole filling and spikes removal are also available.

I might take a stab at implementing this sometimes.

pierotofy:
XYZ --> gdal_grid --> gdal_translate back to XYZ --> PCL poisson could be a tentative approach.

http://www.gdal.org/gdal_grid.html

smathermather:
I like it. Maybe I can give that a whirl. Do you have a dataset you'd like it tested on?

pierotofy:
I've been using this one since it has some trees and it's not too large: https://github.com/pierotofy/drone-datasets/tree/master/sheffield_park

pierotofy:
Processing seems very slow, btw. image

pierotofy:
animation

Steps:

1. Save `merged.ply` from OpenSfM as CSV.
2. Create VRT:

```
<OGRVRTDataSource>
    <OGRVRTLayer name="merged">
        <SrcDataSource>merged.csv</SrcDataSource>
        <SrcLayer>merged</SrcLayer>
        <GeometryType>wkbPoint</GeometryType>
        <GeometryField encoding="PointFromColumns"/>
    </OGRVRTLayer>
</OGRVRTDataSource>
```

3. `GDAL_NUM_THREADS=2 gdal_grid -ot Float32 -outsize 1024 1024 -a linear layer.vrt out.tif`
4. `gdal_translate -of XYZ out.tif out.xyz`
5. Do Poisson Recon on out.xyz (I used depth 12 divider 9)

It's not perfect, as poisson reconstruction tends to create "concave edges" (not sure that's right term), whereas a true heightmap should go straight down in the Z direction.

image

dakotabenjamin:
It also looks like it adversely affects the actual shape of things (like that little building). I am going to merge this PR today.

PR #446, opened 2016-12-29
Multithreaded georef (issues/#435)

pulquero:
Both best camera and best GCP multithreaded.
Seems to be working well.

### DIFF
--- a/modules/odm_georef/src/Georef.cpp
+++ b/modules/odm_georef/src/Georef.cpp
@@ -1141,9 +1141,40 @@ void Georef::createGeoreferencedModelFromExifData()
 
 void Georef::chooseBestGCPTriplet(size_t &gcp0, size_t &gcp1, size_t &gcp2)
 {
-    double minTotError = std::numeric_limits<double>::infinity();
+    size_t numThreads = boost::thread::hardware_concurrency();
+    boost::thread_group threads;
+    std::vector<GeorefBestTriplet*> triplets;
+    for(size_t t = 0; t < numThreads; ++t)
+    {
+        GeorefBestTriplet* triplet = new GeorefBestTriplet();
+        triplets.push_back(triplet);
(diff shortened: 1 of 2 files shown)


modules/odm_georef/src/Georef.cpp
```diff
@@ -1185,9 +1185,39 @@ void Georef::chooseBestGCPTriplet(size_t &gcp0, size_t &gcp1, size_t &gcp2)
 
 void Georef::chooseBestCameraTriplet(size_t &cam0, size_t &cam1, size_t &cam2)
 {
+    size_t numThreads = 2;
```
pierotofy: Very nice! In the future in might be useful to change numThreads from 2 to a function that retrieves the number of available CPUs of the machine.

pierotofy (approved):

pierotofy (approved):

pulquero:
If there is anything else that needs multithreading, let me know. I'll be willing to do it, now I got the hang of it. BTW, I've also noticed that in most places across all the modules, reserve() isn't being used to pre-allocate large vectors. Might be worth considering to save some heap thrashing.

pierotofy:
@pulquero it would be interesting to run a profiling tool and detect the functions / modules that are taking the longest to compute. Those would be the ones to target for additional multithreading. Perhaps you might be interested in publishing the profiling results on the wiki.

dakotabenjamin:
Thanks for this! I do not think `--time` is adequate for what @pierotofy is suggesting because it is too simplistic.

xialang2012:
Ortho module is also slow too. So...

pulquero:
I think the geotiff is to blame for that. It seems to take an extraordinary amount of time to generate the geotiff compared to the png.

xialang2012:
The expensive part of the Ortho module is https://github.com/OpenDroneMap/OpenDroneMap/blob/master/modules/odm_orthophoto/src/OdmOrthoPhoto.cpp#L468.

pulquero:
That looks like it can be multi-threaded, I'll have a go at it.

dakotabenjamin:
That and https://github.com/OpenDroneMap/OpenDroneMap/blob/master/scripts/odm_orthophoto.py#L94 for large datasets. This is already multithreaded if you have an updated gdal. It always throws an unsupported warning for me though.

PR #452, opened 2017-01-12
Make number of Opensfm processes configurable

pulquero:
A significant amount of memory is required to run opensfm on num_cores compared to all the other stages. We can reduce its memory needs by simply running less instances of it in parallel. This PR adds a --opensfm-processes option to make it possible to run opensfm on less than num_cores without affecting everything else. Ideally we would like more granularity, and run prematching on num_cores too, just reconstruction and depth-mapping on less.


dakotabenjamin (approved):
Good idea. Where does it fail if you input -1?

dakotabenjamin:
It fails 'good enough'
```bash
  File "/home/vagrant/opendronemap/SuperBuild/src/opensfm/bin/opensfm", line 34, in <module>
    command.run(args)
  File "/home/vagrant/opendronemap/SuperBuild/src/opensfm/opensfm/commands/detect_features.py", line 32, in run
    p = Pool(processes)
  File "/usr/lib/python2.7/multiprocessing/__init__.py", line 232, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 152, in __init__
    raise ValueError("Number of processes must be at least 1")
ValueError: Number of processes must be at least 1
  File "/home/vagrant/opendronemap/run.py", line 50, in <module>
    plasm.execute(niter=1)
  File "/home/vagrant/opendronemap/scripts/opensfm.py", line 83, in process
    (context.pyopencv_path, context.opensfm_path, tree.opensfm))
  File "/home/vagrant/opendronemap/opendm/system.py", line 28, in run
Exception: Child returned 1
```

PR #464, opened 2017-02-01
Reduced memory fragmentation

pulquero:
Where possible, vectors are sized with the correct size to prevent memory fragmentation due to vector growth/resizing.


dakotabenjamin:
I can test this on Monday if no one else gets to it before then.

modules/odm_orthophoto/src/OdmOrthoPhoto.cpp
```diff
@@ -397,6 +397,15 @@ void OdmOrthoPhoto::createOrthoPhoto()
         size_t vertexIndexCount = 0;
         for(size_t t = 0; t < mesh.tex_polygons.size(); ++t)
         {
+            for(size_t faceIndex = 0; faceIndex < mesh.tex_polygons[t].size(); ++faceIndex)
+            {
+                vertexIndexCount += 3;
```
pierotofy: Couldn't this be written as:

```c
vertexIndexCount += mesh.tex_polygons[t].size() * 3;
```

Instead of a for loop?

pierotofy:
Other changes look good! @pulquero have you tried to benchmark the execution time before/after the additions? Just curious to see how much things improve with the preallocation of memory.

pulquero:
No, it was more to try to reduce the chance of running out of memory half-way through running something. I suspect it won't make a substantial difference to execution times.

PR #471, opened 2017-02-08
update to latest mvs-texturing patch

smathermather:


dakotabenjamin:
This is the branch that introduced regularization? We can merge this but we could also wait for the changes to be implemented in master with command line access.

pierotofy:
But why wait? :)

smathermather:
Ya command line access will be nice, but the output is demonstrably better even without it.

SuperBuild/cmake/External-MvsTexturing.cmake
```diff
@@ -8,8 +8,8 @@ ExternalProject_Add(${_proj_name}
   STAMP_DIR         ${_SB_BINARY_DIR}/stamp
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
-  URL               https://github.com/nmoehrle/mvs-texturing/archive/dab68acaa693275c183c254a958130ee6d29c3e4.zip
-  URL_MD5           0b0466f5d1046699594ce7fc77bdad02
+  URL               https://github.com/nmoehrle/mvs-texturing/archive/e695a29505c74be570b8bbb87cf42f5b1611cf60.zip
+  #URL_MD5           0b0466f5d1046699594ce7fc77bdad02
```
dakotabenjamin: The MD5 for this file is: D11FA3052CEBBF6D34D2072BED6C29FD

dakotabenjamin (approved):

dakotabenjamin:
We need to edit the configure script for upgrading modules. I was having trouble with an in-place build (works fine for new downloads, of course)

PR #485, opened 2017-02-15
GRASS script to create DEM, contour and textured relief maps.

pulquero:
Scripts for downstream workflows involving other applications. This particular PR addresses the creation of DEM, contour and textured relief maps using GRASS.


pierotofy:
That's pretty cool @pulquero!

I know that we've been looking to add DEM generation/filtering using PDAL and that I think @smathermather is working on it, but using GRASS is also an interesting approach.

smathermather:
This is very cool, and I'm not sure how to proceed. On one hand, I don't want to add a dependency. On the other hand, GRASS is a likely candidate to get added as a dependency some day anyway... . Also, we'd have to add to the installer as well as the docker image the GRASS dependencies if we were to add this. Alternatively, we could add a check for the GRASS dependencies, and only use this feature if they are already available.

I'd like to get @dakotabenjamin's thoughts when he's back next Tuesday. I lean toward finding a way to build this in. We have lots of uses for PDAL other than DEM creation... .

pulquero:
One of the reasons I put this in 'contrib' was because of the bloating issue - I wasn't sure if DEM creation was important enough to be in core ODM. I don't know if you would want to integrate the entire script, or just the DEM part of it and leave the rest in contrib. I was also hoping the presence of the script would stimulate more advanced GRASS users to extend and improve it. For instance, I know that GRASS has some lidar functions for extracting out a terrain model, removing any man-made features. I didn't bother adding this because most of my data is pure terrain anyway (I couldn't really test it).

If you think this is cool, just wait for my next PR. It will be in a similar style, but for another application.

pierotofy:
I wonder if we could start a separate repository for collecting scripts related to ODM and then just reference it in the README. While I don't think we should add another dependency, I think some users would find scripts like these useful.

dakotabenjamin (changes requested):
The folder name has an extra 's'.

Make sure to comply with PEP8

Otherwise :+1: see thread for more discussion

dakotabenjamin:
I like @pierotofy 's idea, but I am not sure how to organize such a repo. I think a contrib folder for now is fine though.

smathermather:
I don't know if we need a separate repository, putting it in contribs does seem like the equivalent, lightweight way to do the same.

dakotabenjamin (approved):

PR #490, opened 2017-02-19
Update config.py

pierotofy:
By default ODM should use the`--outlier_removal` feature of mvs texturing. This improves significantly the output when moving features such as cars or people are caught during a survey.

It's also not an option that many people have noticed, since I had none report that it was unusable on node-opendronemap (due to a bug), so I would propose that we turn it on by default.

`gauss_damping` works well also, but I found more aggressive results using `gauss_clamping`.

Without --outlier_removal:

image

With --outlier_removal gauss_clamping:

image

With --outlier_removal gauss_damping:

image

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -232,7 +232,7 @@ def config():
 
     parser.add_argument('--texturing-outlier-removal-type',
                         metavar='<string>',
-                        default='none',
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
Awesome! Great find. I will test and merge ASAP

PR #498, opened 2017-02-27
Add NDVI contrib module

dakotabenjamin:
Adds a contrib module that computes an NDVI from CIR imagery.


smathermather (approved):
Awesome! And great to have in the contribs directory

PR #505, opened 2017-03-03
Update packages.dockerfile

formatkaka:
The docker image build command crashes with the error `add-apt-repository: command not found `. Need to install the module `software-properties-common` before adding ppa. [source](http://askubuntu.com/q/593433/538532)


dakotabenjamin:
This is a duplicate of #504

PR #510, opened 2017-03-10
Global settings file, Versioning, GDAL creation options

dakotabenjamin:
The main changes are below:

- Global settings file: 
  - This makes it possible to set run-time parameters on a software level, enabling consistent runs and easier task management
  - Copy the default.settings.yaml file to a settings.yaml file and edit as needed. 
  - you have to set the project-path setting

- Project Name UI:
  - With a global project-path setting, now you name projects using the positional parameter:
    - `python run.py --images /home/dmb/toledo/ toledo`
  - `--images <path>` is only needed in creation of the task the first time

- Others:
  - `--version` gives the ODM version
  - GDAL creation options target_srs (dummy parameter for now, need advice from @smathermather ), compression, and tiling options


opendm/config.py
```diff
@@ -17,28 +31,34 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+
+    parser.add_argument('name',
+                        metavar='<project name>',
+                        type=alphanumeric_string,
+                        help='Name of Project (i.e subdirectory of projects folder)')
 
     parser.add_argument('--resize-to',  # currently doesn't support 'orig'
                         metavar='<integer>',
-                        default=2400,
+                        default=defaultSettings['resize_to'],
```
pierotofy: Ahhh this is going to break my argument parser in node-odm :)

Is there a particular reason why the default.settings.yaml file is needed? Information about default values I thought was already available in this file. Just trying to understand better the logic here.

opendm/config.py
```diff
@@ -285,32 +304,61 @@ def config():
...
     parser.add_argument('--orthophoto-resolution',
                         metavar='<float > 0.0>',
-                        default=20.0,
+                        default=defaultSettings['orthophoto']['resolution'],
                         type=float,
                         help=('Orthophoto ground resolution in pixels/meter'
                               'Default: %(default)s'))
 
-    parser.add_argument('--zip-results',
+    parser.add_argument('--orthophoto-target-srs',
```
pierotofy: Yay! Been waiting for this, nice work!

opendm/config.py
```diff
@@ -285,32 +304,61 @@ def config():
...
+                             'Default: %(default)s')
+
+    parser.add_argument('--orthophoto-no-tiled',
                         action='store_true',
                         default=False,
+                        help='Set this parameter if you want a stripped geoTIFF.\n'
+                             'Default: %(default)s')
+
+    parser.add_argument('--orthophoto-compression',
+                        metavar='<STRING>',
```
pierotofy: Should be "`<string>`" (lowercase)

opendm/config.py
```diff
@@ -17,28 +31,34 @@ def __call__(self, parser, namespace, values, option_string=None):
...
                         help='Path to input images'),
 
     parser.add_argument('--project-path',
-                        metavar='<string>',
-                        help='Path to the project to process')
+                        metavar='<path>',
+                        help='Path to the project folder',
+                        default=defaultSettings['project_path'])
+
+    parser.add_argument('name',
```
pierotofy: Is this parameter being added to allow a user to have:

/project/a/images
/project/b/images
...

And thus call ODM with: `run.py --project-path /project/ --name a`?

Or what is the intended use?

README.md
```diff
@@ -65,13 +66,24 @@ Note that using `run.sh` sets these temporarily in the shell.
     
 ### Run OpenDroneMap
 
-First you need a set of images, taken from a drone or otherwise. Example data can be cloned from https://github.com/OpenDroneMap/odm_data
+First you need a set of images, taken from a drone or otherwise. Example data can be obtained from https://github.com/OpenDroneMap/odm_data
+
+Next, you need to copy over the settings file `default.settings.yaml` and edit it. The only setting you must edit is the `project-path` key. Set this to an empty directory within projects will be saved. There are many options for tuning your project. See the [wiki](https://github.com/OpenDroneMap/OpenDroneMap/wiki/Run-Time-Parameters) or run `python run.py -h`
```
pierotofy, line 71: I'm a little confused here, I take `default.settings.yaml` and copy it over (where?).

I think I see where this ties in with the new `--name` parameter, but I'm a little confused. I think other people will be as well. I wonder whether this ability to do project management and organization via command line should be coded in a module separate from `run.py`. Borrowing from Unix philosophy, make lot of small utilities, that work well separately, but flexible enough to be combined.

pierotofy (commented):
GDAL creation and versioning looks great!

I'd like to learn more about the global settings file and project naming.

README.md
merkato: I understand this that way, you create settings.yaml in main directory of ODM, and sets `project-path` to `/directory_with_odm_projects/ `
Then starting ODM with `./run.sh odm_project_name `- this `--name` is optional here ;)

@pierotofy IMHO we need good documentation of OpenDroneMap anyway - in terms of API and using in production.

merkato (commented):
command line argument for compression works as expected. How to use it in settings.yaml?  Added new line in Orthophoto block - compression: LZW and nothing happens.

merkato (commented):
There must be a vaild project name in run arguments, to check version. This might be confusing for beginner's

opendm/config.py
dakotabenjamin: Sorry I should have updated the info but I ran out of time last week. Hopefully the updated comment above will help clear things up.

opendm/config.py
dakotabenjamin: I simply hardcoded them in the yaml file rather than here because that was simplest place to write the changes. Now they can be updated by the user in that interface. I think it would be confusing if the defaults were in both, esp. if they changed in one or the other, not knowing which takes precedence.

opendm/config.py
dakotabenjamin: Unfortunately target_srs doesn't work because gdal_translate doesn't have that option. @smathermather if you have any suggestions?

opendm/config.py
dakotabenjamin: 👍

opendm/config.py
pierotofy: Yes you wouldn't want to to have defaults in two places. I guess I'm failing to see the use case for having a defaults file.

If parameters are hard coded in config.py, those are the defaults. If a user wants to change the defaults, it can do so by creating a config file.

I guess my question is, what is the use case for having a defaults config file that cannot be achieved with a normal config file?

opendm/config.py
pierotofy: I think `gdalwarp` has that. But then perhaps we should include other options such as `resampling_method`.

README.md
dakotabenjamin: The reason there's a default.settings.yaml that you need to copy over is because of git. People editing that and then trying to push changes could rewrite what we've said are reasonable defaults.

pierotofy: You could put default.settings.yaml in the .gitignore (and force add it the first time).

opendm/config.py
dakotabenjamin: Adding gdalwarp would be a good idea.

opendm/config.py
dakotabenjamin: The settings file could have all the options commented out, then if you want to override uncomment and edit. How do we prevent unnecessary changes in git though?

pierotofy: I think you could use .gitignore directives.

dakotabenjamin: So I'm trying to implement these changes:
- keep defaults in config.py
- uncomment lines to override

But I am having trouble scripting the override function. In run.py, after the args are parsed, I want to do this: 
```python
if __name__ == '__main__':

    args = config.config()

    # check for overrides

    # Load global settings file
    with open(context.settings_path) as stream:
        settings = safe_load(stream)

    # check that the project path setting has been set properly
    if settings['project_path'] == '':
        log.ODM_ERROR('You need to set the project path in the settings.yaml file before you can run ODM')
        sys.exit(1)

    # for each setting, check for override
    for setting in settings:
        if setting in args:
            args[setting] = settings[setting]
```

The issue is that I can't assign args[setting] and I can't figure out how to do that without converting to a dict using vars(). I'd rather not go through the whole project reassigning the Namespace to dict.

dakotabenjamin: I need to do something like `args.setting = settings[setting]` but that seems impossible. I also can't iterate through a Namespace.

pierotofy: What if you changed config() in config.py:

From:
```python
return parser.parse_args()
```

To:

```python
return vars(parser.parse_args())
```

?

pierotofy: Speaking of which, it looks like some of the logic that you are writing in __main__ could be inserted into config().

dakotabenjamin: I was trying to avoid using a dictionary for args, but it seems like more trouble to keep it as a Namespace. If I do that then you are correct, I should be putting it in config()

dakotabenjamin:
@pierotofy if you've got the time I'd appreciate reviewing the new changes.

opendm/config.py
```diff
@@ -227,12 +247,14 @@ def config():
     parser.add_argument('--texturing-data-term',
                         metavar='<string>',
                         default='gmi',
+                        choices=['gmi','area'],
```
pierotofy: Nice! This is in line with https://github.com/OpenDroneMap/WebODM/issues/129

scripts/resize.py
```diff
@@ -93,11 +93,11 @@ def process(self, inputs, outputs):
         log.ODM_DEBUG('Resizing dataset to: %s' % tree.dataset_resize)
 
         # check if we rerun cell or not
-        rerun_cell = (args.rerun is not None and
-                      args.rerun == 'resize') or \
-                     (args.rerun_all) or \
-                     (args.rerun_from is not None and
-                      'resize' in args.rerun_from)
+        rerun_cell = (args['rerun'] is not None and
```
pierotofy: I'm seeing the lines:

```
(args['rerun'] is not None and
 -                      args.rerun == 'resize') or \		 +                      args['rerun'] == 'resize') or \
 -                     (args.rerun_all) or \		 +                     (args['rerun_all']) or \
 -                     (args.rerun_from is not None and		 +                     (args['rerun_from'] is not None and
``` 

Being used in several places, this would be a good place to add the logic for checking the rerun in a function. Doesn't have to be done as part of this PR, but it's an improvement for the future.

pierotofy (approved):
It looks very good! As soon as this is merged I will do some work on node-OpenDroneMap to make sure that things work properly. Good job!

pierotofy:
p.s. I know you've already refactored all calls to args in dict style (args.param --> args['param']), but you could have saved some code by using the "Bunch" method: http://stackoverflow.com/questions/2597278/python-load-variables-in-a-dict-into-namespace 👍 

```python
class Bunch(object):
  def __init__(self, adict):
    self.__dict__.update(adict)

a = {'a': True, 'params': 10}
x = Bunch(a)
print(x.params) # 10
```

opendm/config.py
dakotabenjamin: Yes that is what made me check

scripts/resize.py
dakotabenjamin: :+1:

dakotabenjamin:
D'oh! I'm tempted to revert and do it right. I do like Namespaces.

pierotofy:
Up to you 😄

dakotabenjamin:
It was a good thing I did because I found a critical bug, I was overriding options wrong. 

What do you think of using a package like https://pypi.python.org/pypi/appsettings ? It may make the task easier

pierotofy:
If it trims away code so that we don't have to reinvent the wheel, I'm all for it!

dakotabenjamin:
@pierotofy I had to do a hard reset on the remote repo so make sure you remove the branch if you checked it out.

PR #517, opened 2017-03-22
Resume of opensfm from matching

pulquero:
Opensfm matching can take a long time. It is annoying to have to redo matching (and everything before) if opensfm fails on a following step, e.g. reconstruction due to lack of memory. This change writes a marker file after matching so it is possible to resume opensfm from after matching.

Resume from after create tracks is also supported.

### DIFF
--- a/opendm/types.py
+++ b/opendm/types.py
@@ -360,6 +360,7 @@ class ODM_Tree(object):
         self.benchmarking = io.join_paths(self.root_path, 'benchmark.txt')
 
         # opensfm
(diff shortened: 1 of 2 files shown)


pierotofy (approved):
Looks good to me!

smathermather:
This is awesome. I won't be able to test in a reasonable time, so I'll defer to @pierotofy and @dakotabenjamin.

dakotabenjamin:
This works for me. What if we put a similar check between bundling and depth-map computation as well?

dakotabenjamin (approved):

PR #525, opened 2017-03-29
Add BigTIFF creation and overviews option to odm_orthophoto

dakotabenjamin:
This PR adds the `--orthophoto-bigtiff` gdal_translate creation option to orthophoto creation.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -354,6 +354,11 @@ def config():
                              'https://www.gdal.org/frmt_gtiff.html for more info. '
                              '\nDefault: %(default)s')
 
+    parser.add_argument('--build-overviews',
+                        action='store_true',
+                        default=False,
+                        help='Build orthophoto overviews using gdaladdo.')
+
     parser.add_argument('--zip-results',
                         action='store_true',
                         default=False,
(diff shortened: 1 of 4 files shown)


pierotofy (approved):
Looks good! I wonder if there would be a way to detect this automatically in the future.

smathermather:
Can default to
* IF_NEEDED will only create a BigTIFF if it is clearly needed (in the uncompressed case, and image larger than 4GB. So no effect when using a compression).

or 
* IF_SAFER will create BigTIFF if the resulting file *might* exceed 4GB. Note: this is only a heuristics that might not always work depending on compression ratios.

See http://www.gdal.org/frmt_gtiff.html

Could just use the IF_SAFER flag as default. Anyone who wants to override (who has an opinion) can then override, but then it will create a bigtiff if there's a chance the file will exceed 4GB. Also, as we have no compression set by default (I think...) the heuristics should be pretty good.

dakotabenjamin:
Default right now is IF_NEEDED, but we could make it so IF_SAFER is the default for sure. The default compression is DEFLATE. In any case, I'll add all the options as opposed to True/False for more user control.

PR #531, opened 2017-04-03
A docker file update

pepperlk:
Moved to phusion/baseimage
Cleaned up how apt is working on packages
Cleaned up test and sample data in SuperBuild
Moved packages.Dockerfile into Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN add-apt-repository -y ppa:george-edison55/cmake-3.x
 RUN apt-get update -y
 
 # All packages (Will install much faster)
-RUN apt-get install -y --no-install-recommends  git cmake python-pip build-essential software-properties-common python-software-properties libgdal-dev gdal-bin libgeotiff-dev \
+RUN apt-get install -y git cmake python-pip build-essential software-properties-common python-software-properties libgdal-dev gdal-bin libgeotiff-dev \
 libgtk2.0-dev libavcodec-dev libavformat-dev libswscale-dev python-dev python-numpy libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libjasper-dev libflann-dev \
 libproj-dev libxext-dev liblapack-dev libeigen3-dev libvtk5-dev python-networkx libgoogle-glog-dev libsuitesparse-dev libboost-filesystem-dev libboost-iostreams-dev \
 libboost-regex-dev libboost-python-dev libboost-date-time-dev libboost-thread-dev python-pyproj python-empy python-nose python-pyside python-pyexiv2 python-scipy \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,34 +1,71 @@
...
+RUN add-apt-repository -y ppa:ubuntugis/ppa
+RUN add-apt-repository -y ppa:george-edison55/cmake-3.x
+RUN apt-get update -y
+
+# All packages (Will install much faster)
+RUN apt-get install --no-install-recommends -y git cmake python-pip build-essential software-properties-common python-software-properties libgdal-dev gdal-bin libgeotiff-dev \
+libgtk2.0-dev libavcodec-dev libavformat-dev libswscale-dev python-dev python-numpy libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libjasper-dev libflann-dev \
+libproj-dev libxext-dev liblapack-dev libeigen3-dev libvtk5-dev python-networkx libgoogle-glog-dev libsuitesparse-dev libboost-filesystem-dev libboost-iostreams-dev \
+libboost-regex-dev libboost-python-dev libboost-date-time-dev libboost-thread-dev python-pyproj python-empy python-nose python-pyside python-pyexiv2 python-scipy \
+jhead liblas-bin python-matplotlib libatlas-base-dev
```
dakotabenjamin, line 19: Why put this here and in packages.Dockerfile? It seems redundant

dakotabenjamin (commented):
There is also a core2.Dockerfile that is necessary for @pierotofy in WebODM.

Dockerfile
pierotofy: I think this way the packages.Dockerfile could be removed altogether.

dakotabenjamin: I'm for that. Makes it easier on users.

pierotofy:
I think these are good changes, but we need good testing before merging. In particular, packages.Dockerfile builds off of Ubuntu 14.04, whereas phusion/baseimage is based on Ubuntu 16.04.

pierotofy:
I'm testing the Dockerfile right now, I'll take care of updating `core2.dockerfile` once this is merged.

pierotofy:
I wasn't able to build the image with `docker build -t my_odm_image .` (retrying now, trying to get a build log). @pepperlk were you able to build this image?

pepperlk:
I was able to build on my Linux box and windows box. Might need to do no cache on the docker build command if you already have an image by that name. Here is the command I used. `docker build --no-cache=true -t my_odm_image .`

pierotofy:
I was able to build it the second time around. Not sure why it failed the first time, but it might have been because of old cached intermediate containers.

@dakotabenjamin I think this can be safely merged.

@pepperlk thanks for the PR!

dakotabenjamin:
Great! Thanks everyone. 

My only suggestion is that in the future, be more descriptive in your commit messages. Git standards suggest that you format your commit messages as a description of what you have done in the imperative mode, that is as if you were commanding someone. 

See: https://github.com/erlang/otp/wiki/writing-good-commit-messages

PR #540, opened 2017-04-06
Add reinstall and uninstall options to configuration script

dakotabenjamin:
New configuration script usage: 
```
./configure.sh install
./configure.sh reinstall
./configure.sh uninstall
./configure.sh usage
```


pierotofy:
Looks good!

PR #542, opened 2017-04-07
2.5D Mesh

pierotofy:
This PR aims to replace the poisson mesh with a 2.5D mesh for the purpose of generating a higher quality orthophoto.

- Includes a new lib, CGAL, but doesn't add much time to compilation.
- Adds a new module, `odm_25dmeshing`, which also supports parallel execution
- Modifies existing ecto cells to handle the new module
- Adds 4 new command line parameter to ODM for key parameters users should be able to tune. I considered piggy-backing the maxVertexCount parameter from the poisson mesh, but found that lower maxVertexCount values tend to yield better results for the purpose of orthophoto generation, so users might wish to generate a good poisson mesh (high vertex count), but smooth orthophoto (lower vertex count).
- Users can skip the 2.5D mesh generation via `--skip-25dmesh` parameter.

It should compile properly if you run `. configure.sh` from the /code directory.

Feedback welcome. @smathermather @dakotabenjamin


pierotofy:
I'm still unsure about the `50000` default value for `--25d-mesh-size`, I found out that lower values tend to generate better results on flatter scenes, so we might want to tweak that out.

dakotabenjamin:
Because of multithreading some of the lines here may be for another job:
```
-- USING GCC_VERSION = '5.4.0 20160609'
-- Using gcc version 4 or later. Adding -frounding-math
== Generate version files (DONE) ==

== Set up flags ==
-- Build type: Release
-- USING CXXFLAGS = ' -frounding-math -O3 -DNDEBUG'
-- USING LDFLAGS = ' '
== Set up flags (DONE) ==

== Detect external libraries ==
-- External libraries supported: GMP;GMPXX;MPFR;ZLIB;OpenGL;LEDA;MPFI;RS;RS3;OpenNL;Eigen3;BLAS;LAPACK;QGLViewer;ESBTL;Coin3D;NTL;IPE
-- Preconfiguring library: GMP ...
-- Could NOT find GMP (missing:  GMP_LIBRARIES GMP_INCLUDE_DIR) 
CMake Error at cmake/modules/CGAL_SetupDependencies.cmake:66 (message):
  CGAL requires GMP to be found
Call Stack (most recent call first):
  CMakeLists.txt:673 (include)


-- Configuring incomplete, errors occurred!
See also "/home/dmb2/opendronemap/SuperBuild/build/cgal/CMakeFiles/CMakeOutput.log".
CMakeFiles/cgal.dir/build.make:108: recipe for target 'cgal/stamp/cgal-configure' failed
make[2]: *** [cgal/stamp/cgal-configure] Error 1
CMakeFiles/Makefile2:215: recipe for target 'CMakeFiles/cgal.dir/all' failed
make[1]: *** [CMakeFiles/cgal.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

dakotabenjamin:
So I installed libgmp-dev and libmpfr-dev and it worked

dakotabenjamin (changes requested):
Once you update configure.sh and the Docker installation, update the licenses/licenses.md file, and the README (if needed), we can merge this in.

pierotofy:
Thanks @dakotabenjamin, I've added the dependencies as requested.

dakotabenjamin (approved):

dakotabenjamin:
For the max vertex count, I think it's ok. When dealing with larger datasets, we should think about (for the future) determining "optimal" and "cheap" values based on some proportion of the point cloud size.

pierotofy:
Just found that processing this dataset causes mvs texturing to seg fault with the 2.5D mesh (only). https://support.pix4d.com/hc/en-us/articles/204845195-Example-Datasets-Available-for-Download-DJI-Demo

Perhaps wait to merge this until I can figure this one out.

pierotofy:
Z-occlusion filtering! W00t!

image

image

smathermather:
Whoa!

dakotabenjamin:
On a fresh 16.04 vagrant machine: 
```
In file included from /home/ubuntu/OpenDroneMap/modules/odm_25dmeshing/src/PlyInterpreter.cpp:1:0:
/home/ubuntu/OpenDroneMap/modules/odm_25dmeshing/src/PlyInterpreter.hpp:11:20: fatal error: CGAL.hpp: No such file or directory
compilation terminated.
modules/odm_25dmeshing/CMakeFiles/odm_25dmeshing.dir/build.make:62: recipe for target 'modules/odm_25dmeshing/CMakeFiles/odm_25dmeshing.dir/src/PlyInterpreter.cpp.o' failed
make[2]: *** [modules/odm_25dmeshing/CMakeFiles/odm_25dmeshing.dir/src/PlyInterp                                                                                                 reter.cpp.o] Error 1
CMakeFiles/Makefile2:393: recipe for target 'modules/odm_25dmeshing/CMakeFiles/odm_25dmeshing.dir/all' failed
make[1]: *** [modules/odm_25dmeshing/CMakeFiles/odm_25dmeshing.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

pierotofy:
I'll take a look at it sometimes next week; I'm also still working on resolving some issues related to the alignment of the mesh triangles which tend to generate some distortions in the textures. This addition is proving to be more difficult than I initially expected.

dakotabenjamin:
@pierotofy perhaps if the current status of is good enough we can push the current changes and add that fix as you get to it, once we fix the installation issue.

pierotofy:
Sure, we could do that, but I should probably make the 2.5D processing optional (disabled by default), instead of being enabled by default.

smathermather:
That sounds like a good approach @pierotofy.

pierotofy:
Ok it should be good to be tested/merged. Processing of the 2.5D mesh is enabled by passing `--use-25dmesh`. @dakotabenjamin

dakotabenjamin:
This looks really good!! 

In my opinion, when we choose to run 2.5d mesh, we do so instead of 3d, so we don't have to run the rest of the process on both.

pierotofy:
My choice for running both was that some people might want to still generate a textured 3d model (and poisson recon does a better job at that). But feel free to change it to run only one method if you think it's better. 👍

pepperlk:
Trying to evaluate this, is the only command option the `--use-25dmesh` and it does both 3d and 2.5d or am I missing something?

pierotofy:
`--use-25dmesh`, `--mesh-remove-outliers`, and `--mesh-wlop-iterations` are the new parameters. `--mesh-size` also affects the 2.5D mesh vertex count.

Turning on `--use-25dmesh` computes both 3D and 2.5D, but the orthophoto is generated using the 2.5D.

pepperlk:
Kinda think this wasn't the expected output.
image
run options: --use-25dmesh --orthophoto-resolution 60 --pmvs-num-cores 12

Images: https://drive.google.com/file/d/0BwkZZS_Af25td3AtU2RHU3hHdm8/view?usp=sharing

dakotabenjamin:
Hm, I got a good ortho. 

ran: `./run.sh --project-path ~/ODMProjects fast_test --use-25dmesh --orthophoto-resolution 60 --opensfm-processes 6` (note that --pmvs-num-cores does nothing if you don't use --use-pmvs)
croppercapture 23

pepperlk:
Thanks ill recompile and rerun.. It was behind node-odm is the reason its sets those for the pmvs. any ways will try again.. But I do think the 'or' on the 3d vs 2.5d is a valid deal here to possibly speed up the guys needing the flat files vs the 3d. So maybe a I propose a '--skip-3dmesh' in a  new pr  PR maybe?

dakotabenjamin:
@pepperlk @pierotofy I'll merge this tomorrow as is, unless there are any other issues.

smathermather:
Excellent.

PR #544, opened 2017-04-10
Eigen's aligned allocator is required with pcl

gadomski:
This patch gets things building on OSX with the default clang compiler (at least for me).

### DIFF
--- a/modules/odm_georef/src/Georef.cpp
+++ b/modules/odm_georef/src/Georef.cpp
@@ -1359,7 +1359,7 @@ bool Georef::loadObjFile(std::string inputFile, pcl::TextureMesh &mesh)
     std::size_t f_idx = 0;
     std::string line;
     std::vector<std::string> st;
-    std::vector<Eigen::Vector2f> coordinates;
+    std::vector<Eigen::Vector2f, Eigen::aligned_allocator<Eigen::Vector2f> > coordinates;
     std::vector<Eigen::Vector2f> allTexCoords;
 
     std::map<int, int> f2vt;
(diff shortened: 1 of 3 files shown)


dakotabenjamin:
Thanks for the contribution. Can you provide a little more details about these changes? What problem exactly are you trying to address?

I'm also interested in the OSX build. Will these changes alone facilitate a build or should we expect more to come? What's your process? None of the maintainers have an OSX machine so it will be hard to test, but we will at least make sure any changes don't break the standard build process.

gadomski:
Working from 76f8b765dca6a4fb27a2737fb1752108868d6b45 I get a few errors like this one when building from the top-level CMakeLists.txt:

```
/Users/rdcrlpjg/Repos/OpenDroneMap/modules/odm_georef/src/Georef.cpp:1467:49: error: no viable conversion from 'vector<[...], (default) std::__1::allocator<Eigen::Matrix<float, 2, 1, 0, 2, 1> >>' to 'const vector<[...], Eigen::aligned_allocator<Eigen::Matrix<float, 2, 1, 0, 2, 1> >>'
                mesh.tex_coordinates.push_back (coordinates);
                                                ^~~~~~~~~~~
/usr/local/Cellar/eigen/3.3.3/include/eigen3/Eigen/src/StlSupport/StdVector.h:78:3: note: candidate constructor not viable: no known conversion from 'std::vector<Eigen::Vector2f>' (aka 'vector<Matrix<float, 2, 1> >') to 'const std::__1::vector<Eigen::Matrix<float, 2, 1, 0, 2, 1>, Eigen::aligned_allocator<Eigen::Matrix<float, 2, 1, 0, 2, 1> > > &' for 1st argument
  EIGEN_STD_VECTOR_SPECIALIZATION_BODY
```

Because PCL [uses `Eigen::aligned_allocator`](https://github.com/PointCloudLibrary/pcl/blob/master/common/include/pcl/TextureMesh.h#L101) in some places, I needed to use `aligned_allocator` in ODM to get things to compile.

> Will these changes alone facilitate a build or should we expect more to come?

I'm guessing there will be more, unless things Just Work :-).

> What's your process?

I'm doing my best to follow your general build workflow, of course subbing out any Ubuntu-specific instructions. I then plan on using your sample data first to ensure functionality, before moving onto my own stuff.

Right now I'm using the OSX-packaged clang, but once that's working I might try out other compilers as well.

> we will at least make sure any changes don't break the standard build process.

👍

smathermather:
And I can test on OS X when you get to that point, @gadomski. Cheers!

gadomski:
I'm closing, as there's going to be more work than just this patch to get things working on OSX. I'll open up a new PR if/when I get it together. Thanks folks.

dakotabenjamin:
I look forward to future contributions!

PR #549, opened 2017-04-20
Add DEM generation using PDAL

dakotabenjamin:
This PR adds another PDAL pipeline function to odm_georeferencing to build a DEM using a progressive morphological filter to remove trees and structures. By default this option is disabled. There are 4 new command-line inputs:
```
--dem
    Use this tag to build a DEM using a progressive morphological filter in PDAL.

--dem-sample-radius
    Minimum distance between samples for DEM generation.
    Default=1.0

--dem-resolution
    Length of raster cell edges in X/Y units.'
    Default: 2.0

--dem-radius
    Radius about cell center bounding points to use to calculate a cell value.
    Default: 0.5
```

It also switches the LAS PDAL pipeline to use JSON instead of XML. There's still some need for tweaking the default parameters.


dakotabenjamin:
croppercapture 21

dakotabenjamin:
With regards to default values, there could be a check for the dataset size and scale them accordingly, but I don't know if that kind of assumptions should be made, or could be made well.

merkato:
Some thoughts of user experience: 
As an reference - GRASS r.in.lidar with default settings, 1m resolution 
r_in_lidar_grass

ODM DEM with default settings
odm_dem_default

dem_sample_radius: 0.5, dem_resolution: 1, dem_radius: 0.8
odm_dem_08r_05s

dem_sample_radius: 1, dem_resolution: 1, dem_radius: 1.2
odm_dem_12r_05s


Model processed from Jablunkovske Sance dataset (old XVII century fortifications on Czech/Slovak/Polish border). I will set new repo for this dataset for testing ODM new features. If these images are of any value, feel free to use for documentation.

PR #560, opened 2017-05-02
Grass update

hblanken:
Clarifies GRASS compile and install and links code to ODM project DEM.


pulquero:
I remember why I didn't have to compile it originally, it came pre-compiled as part of QGIS.

PR #572, opened 2017-05-09
Use custom MVS repo

dakotabenjamin:
There was an update to https://github.com/nmoehrle/mve that caused some issues with installing mvs-texturing. We forked nmoehrle/mvs-texturing to OpenDroneMap/mvs-texturing and pulled a previous version of mve to that installation process. This should be a temporary fix as we would like to keep getting updates to the mvs-texturing software. 

Instructions: 
Run `bash configure.sh reinstall` to make sure the old downloaded files are deleted.


dakotabenjamin:
This has been fixed on mvs-texturing, so I'm going to use that instead. See: https://github.com/nmoehrle/mvs-texturing/issues/76

PR #576, opened 2017-05-15
--skip-resize flag and writing final transform in extra file, so data from reconstruction.json can be transformed into georeferenced space

fredlllll:
adding a --skip-resize flag, so instead of having to figure out the size of a picture and then passing that to --resize-to, the images just get copied, which saves us the work of encoding the images again, even if size didnt change


pierotofy (approved):
Looks good to me!

dakotabenjamin:
I'm running into this error: 
```
[DEBUG]   Resizing dataset to: /home/dmb2/opendronemap/tests/test_data/images_resize
  File "/home/dmb2/opendronemap/run.py", line 46, in <module>
    plasm.execute(niter=1)
  File "/home/dmb2/opendronemap/scripts/resize.py", line 131, in process
    if params.skip_resize:
NameError: global name 'params' is not defined
```

dakotabenjamin:
needs to be `self.params.skip_resize`

dakotabenjamin:
Can you describe the changes being made to georef.cpp?

fredlllll:
darn that wasnt supposed to go into this pullrequest D: it should basically output the final transform in an extra file, not just in the log. i just wanted to test it, but also ran into the error with opencv missing (see gitter). dont know if that is cause of my changes, or cause of something else

fixed the missing self.

dakotabenjamin:
No worries, in the future you can checkout a new branch and work on it so you don't affect `master`

fredlllll:
ok after running configure.sh a second time it found opencv miraculously and the georef.cpp compiled

PR #630, opened 2017-07-18
Make Docker more prominent for installation

dakotabenjamin:
@pierotofy I added Docker instructions to the QUICKSTART section. Please suggest changes that would add to the ease of use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,17 +31,30 @@ OpenDroneMap can run natively on Ubuntu 14.04 or later, see [Build and Run Using
 
 Current version: 0.3.1 (this software is in beta)
 
+### Docker
+The easiest way to run ODM is through Docker. If you don't have it installed,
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Looks good!

I personally would move the following paragraphs:

```
OpenDroneMap can run natively on Ubuntu 14.04 or later, see Build and Run Using Docker for running on Windows / MacOS. A Vagrant VM is also available.

Support for Ubuntu 12.04 is currently BROKEN with the addition of OpenSfM and Ceres-Solver. It is likely to remain broken unless a champion is found to fix it.
```

After the Docker paragraph block, but you decide. 👍

dakotabenjamin:
Hows that?

pierotofy:
Looks great to me!

PR #639, opened 2017-07-26
Repair bad index checking

merkato:
isinstance(object, classinfo) checks only if object has correct python type, not value. Let's try this with more traditional method ;)


pierotofy:
Looks good to me!

PR #644, opened 2017-07-31
Update VTK, PCL, and OpenSfM, remove resize module

dakotabenjamin:
This should make ODM work with 16.10+

Do not merge yet, needs testing. 

Also adds functionality for opensfm to utilize altitude tags if all the photos have altitude data.


pierotofy:
Nice! If the update to opensfm includes the changes introduced from https://github.com/mapillary/OpenSfM/commit/d22b94c6a46f5be88857290363943f084be4a7f9 I would  revert the changes introduced in https://github.com/OpenDroneMap/OpenDroneMap/pull/579 so that the number of processes matches the number of CPUs available.

dakotabenjamin:
It does not, but will be implemented in 0.2.0 I believe.

dakotabenjamin:
Current major issues: 
- odm_georeferencing updates the images_resize with gps info, except it's in UTM and needs to be latlon. opensfm now uses exif info so it won't find any matches if we try to rerun without deleting the resized images
- I need to add some user control for the new params

dakotabenjamin:
I removed the resize module to fix some GCP issues. The resize-to option now only applies to opensfm, and I think it only affects speed and quality, but I will test that more.

opendm/types.py
```diff
@@ -229,48 +254,48 @@ def utm_to_latlon(self, _file, _photo, idx):
...
-        if latlon[2] >= 0:
-            altref = '0'
-        else:
-            altref = '1'
-        key = 'Exif.GPSInfo.GPSAltitudeRef'
-        metadata[key] = pyexiv2.ExifTag(key, altref)
-
-        # write values
-        metadata.write()
+        # #set values
```
pierotofy, line 257: Should these be removed, or are they commented out for a future need?

pierotofy (approved):

opendm/types.py
dakotabenjamin: Yeah I want to be able to write each camera's gps exif when using GCPs but currently it's not properly converting to wgs84.

pierotofy: Ok! Looks good to me then! 😄

dakotabenjamin:
This PR does NOT fix the issues with 16.10/17.04. I may still merge but will update the README to indicate as such.

pierotofy:
For the opensfm update, would it be feasible to include all the changes up to https://github.com/mapillary/OpenSfM/commit/0a5a6f07b66088e9b63bddaf29bb8f49dd39f9e4 instead of using v0.1.0? It would greatly improve memory usage and fix a lot of the memory issues users have been experiencing. Just a thought.

dakotabenjamin:
Yeah I'll test that now

dakotabenjamin:
With this should I set the default number of porcesses for opensfm to nproc again? It's working great so far.

pierotofy:
I would say yes.

dakotabenjamin:
I'm gonna merge around 4PM today because I love to push big updates on a Friday and then turn off notifications for the weekend :rofl:

pierotofy:
image

😆

dakotabenjamin:
image

FWIW memory management seems pretty good on a nearly 1000 image dataset. (16 core 42GB server though :wink:)

PR #645, opened 2017-07-31
Removed unused matcher_threshold and matcher_ratio

pierotofy:
While answering a question for a user, I noticed that matcher_threshold and matcher_ratio are parameters that are currently not being used.

Perhaps somebody can review this statement and confirm this, but I couldn't find any reference to these two parameters anywhere in the project.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -115,21 +115,6 @@ def config():
                               'More features leads to better results but slower '
                               'execution. Default: %(default)s'))
 
-    parser.add_argument('--matcher-threshold',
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
Can confirm, these are from the matcher we used with the old bundler: https://github.com/OpenDroneMap/OpenDroneMap/blob/fa62369f2906860e4b8e4de4824e5a3d320700f3/run.pl#L611
(going back and remembering how unbearable working with perl was)

dakotabenjamin (approved):

PR #668, opened 2017-09-12
added apt-get update to allow repository additions, not building otherwise

PeterSprague:
Revised Dockerfile -> added apt-get update & software-properties-common first thing to allow repository additions, not building otherwise


dakotabenjamin:
Thanks! Sometimes I forget to update the docker install when changing the native install...

Issue #674, opened 2017-09-21
odm_extract_utm does not parse altitude correctly


dakotabenjamin:
Thanks @fredlllll for helping diagnose this bug

In https://github.com/OpenDroneMap/OpenDroneMap/blob/master/modules/odm_extract_utm/src/UtmExtractor.cpp#L310 odm_utm_extract is looking for the jhead output with "GPS Altitude :" but this is not always how image exif is saved. See: 
![image](https://user-images.githubusercontent.com/1847818/30714478-ea549b26-9ee0-11e7-93ae-166423623877.png)
vs:
![image](https://user-images.githubusercontent.com/1847818/30714949-ba1a16b4-9ee2-11e7-923b-dcf234eb7f75.png)

This is both an issue with odm_utm_extract AND jhead. We should find a new library for reading image exif into C++ and update the module. Options include ImageMagick's identify script, exiftool (which is slow and depends on perl), libexif, or [Exiv2](http://www.exiv2.org).

dakotabenjamin:
exiv2 dependencies are all free software:
```
Feature                     Package   Configure options
--------------------------  --------  ----------------------------
PNG image support           zlib      --without-zlib
                                      --with-zlib=DIR
Native language support     gettext   --disable-nls
Characterset conversions    libiconv  --without-libiconv-prefix
                                      --with-libiconv-prefix[=DIR]
XMP support                 expat     --disable-xmp
                                      --with-expat=DIR

zlib         http://zlib.net/
gettext  *)  http://www.gnu.org/software/gettext/
libiconv *)  http://www.gnu.org/software/libiconv/
expat        http://expat.sourceforge.net/
```

fredlllll:
dependencies for imagemagick here https://github.com/ImageMagick/ImageMagick/blob/master/Install-unix.txt under the "Delegate Configuration"

dakotabenjamin:
a native c++ library is probably more ideal anyways. see: http://exiv2.org/examples.html#example1

fredlllll:
looks much cleaner than the current parsing using a temporary file and the sketchy string processing

KommandorKeen:
Just pulled your fix, compiled it, and am running it now. 
In the opensfm directory the data in the individual exif data files matches the RTK data I inserted into the header using EXIFtool :-).  Just processing the images now, but this is a great step forward as the aircraft is a Phantom4 with an Emlid Reach kinematic GPS post-processed.

[...]

PR #675, opened 2017-09-22
Utm exiv2

dakotabenjamin:
This PR replaces jhead command line util with the exiv2 library. That way we can read exif data properly in odm_extract_utm instead of trying to parse strings. See #674 . Tested on the aukerman dataset, you can see that the altitude data is being written to coords.txt now: 

image

### DIFF
--- a/modules/odm_extract_utm/src/UtmExtractor.cpp
+++ b/modules/odm_extract_utm/src/UtmExtractor.cpp
@@ -188,16 +188,21 @@ void UtmExtractor::extractUtm()
       }
 
       // Parse exif data for positional data
-      double lon, lat, alt;
+      double lon, lat, alt = 0.0;
 
-      if (!parsePosition(exifData, lon, lat, alt)) {
-        std::string error("Failed parsing GPS position.");
+      parsePosition(exifData, lon, lat, alt);
+
+      if (lon == 0.0 || lat == 0.0 || alt == 0.0) {
+        std::string error("Failed parsing GPS position for " + imageFilename);
         throw UtmExtractorException(error);
       }
-
       // Convert to UTM
-      double x, y, z;
+      double x, y, z = 0.0;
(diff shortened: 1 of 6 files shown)


modules/odm_extract_utm/src/UtmExtractor.cpp
```diff
@@ -273,81 +280,54 @@ bool UtmExtractor::convert(const double &lon, const double &lat, const double &a
...
-    istr.clear();
-    istr.str(altStr);
-    istr >> alt >> meterUnitChar;
+  // Altitude
+  if (altitudeTag.count() < 1)
+    throw UtmExtractorException("Image is missing GPS Altitude data");
+  else {
+    Exiv2::URational rAlt = altitudeTag.toRational(0);
+    bool below = (altitudeRef.count() >= 1 && altitudeRef.toLong() == 1);
+    alt = (below ? -1 : 1) * (double)rAlt.first / (double)rAlt.second;
```
pierotofy: Could rAlt.second ever be 0?

pierotofy (approved):

modules/odm_extract_utm/src/UtmExtractor.cpp
dakotabenjamin: I hope not, it's being pulled from a rational  type: http://exiv2.org/tags.html


Tag (hex) | Tag (dec) | IFD | Key | Type | Tag description
-- | -- | -- | -- | -- | --
0x0006 | 6 | GPSInfo | Exif.GPSInfo.GPSAltitude | Rational | Indicates the altitude based on the reference in GPSAltitudeRef. Altitude is expressed as one RATIONAL value. The reference unit is meters.

tomasholderness (commented):
Looks good to me - although I haven't got a working build setup to test this on.

PR #717, opened 2017-11-20
Handle bundler files that have invalid/empty camera parameters

pierotofy:
This is a fix for an issue which causes the orthophoto and textured model to be improperly aligned based on EXIF camera positions. The root cause is that OpenSfM, for certain datasets, outputs multiple bundler reconstruction files, and certain cameras that do not appear in certain reconstructions, are initialized to 0 for all parameters.

odm_georef sometimes ends up choosing one of the uninitialized cameras, causing one of the rotation axis to be flipped.

Related discussion on this PR: http://community.opendronemap.org/t/wrong-view-axis/78/13

### DIFF
--- a/modules/odm_georef/src/Georef.cpp
+++ b/modules/odm_georef/src/Georef.cpp
@@ -190,6 +190,11 @@ Vec3 GeorefCamera::getReferencedPos()
     return Vec3(easting_,northing_,altitude_);
 }
 
+bool GeorefCamera::isValid()
+{
+    return focalLength_ != 0 && k1_ != 0 && k2_ != 0;
+}
+
 std::ostream& operator<<(std::ostream &os, const GeorefCamera &cam)
 {
     os << "Focal, k1, k2 : " << cam.focalLength_ << ", " << cam.k1_ << ", " << cam.k2_ << "\n";
@@ -1058,7 +1063,14 @@ void Georef::createGeoreferencedModelFromExifData()
     {
         throw GeorefException("Not enough cameras in \'" + inputCoordFilename_ + "\' coord file.\n");
     }
-    
+
+    // Remove invalid cameras
+    std::vector<GeorefCamera> goodCameras;
+    for (size_t i = 0; i < cameras_.size(); i++){
+        if (cameras_[i].isValid()) goodCameras.push_back(cameras_[i]);
+    }
+    cameras_ = goodCameras;
+
     // The optimal camera triplet.
     size_t cam0, cam1, cam2;
     
--- a/modules/odm_georef/src/Georef.cpp
+++ b/modules/odm_georef/src/Georef.cpp
@@ -1067,8 +1067,9 @@ void Georef::createGeoreferencedModelFromExifData()
     // Remove invalid cameras
     std::vector<GeorefCamera> goodCameras;
     for (size_t i = 0; i < cameras_.size(); i++){
-        if (cameras_[i].isValid()) goodCameras.push_back(cameras_[i]);
+        if (cameras_[i].isValid()) goodCameras.push_back(GeorefCamera(cameras_[i]));
     }
+    cameras_.clear();
     cameras_ = goodCameras;
 
     // The optimal camera triplet.
--- a/modules/odm_georef/src/Georef.hpp
+++ b/modules/odm_georef/src/Georef.hpp
@@ -95,6 +95,11 @@ struct GeorefCamera
      * \brief getReferencedPos     Get the georeferenced position of the camera.
      */
     Vec3 getReferencedPos();
+
+    /*!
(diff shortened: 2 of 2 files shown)


dakotabenjamin:
My retribution for being lazy and not testing when goodCameras == 0 is that it happened.

dakotabenjamin:
@pierotofy if we pass `--use-fixed-camera-params` then I think it doesn't get k1 and k2. see this gdb output:
```
$29 = {focalLength_ = 2888.8888888900001, k1_ = 0, k2_ = 0, easting_ = -192.5555689, northing_ = 117.0138391, altitude_ = 242.339, transform_ = 0x0,
  position_ = 0x0, pose_ = 0x0}
```

pierotofy:
That seems to be the case; k1 and k2 are read from `bundle_rXXX.out`. What's the content of that file? Also note that odm_georef has only support for a single bundler file, so maybe that's the underlying problem, as opensfm outputs multiple bundler files on occasion.

dakotabenjamin:
I left work before I saw this so I can't send over the bundle file until Monday. OpenSfM only reconstructs the first out file anyways so we shouldn't bother with the others either.

Maybe when the user opts to skip OpenSfM self-calibration they should have to provide those k values. The come from the distortion matrix, right?

dakotabenjamin:
```
# Bundle file v0.3
1081 990966
2888.88888889 0.0 0.0
0.875824345969 -0.481436976786 -0.0339138967323
0.481202678762 0.876475847661 -0.0152993601194
0.0370903890685 -0.00291990588452 0.999307648919
233.819984103 -2.67078179539 -219.799586592
2888.88888889 0.0 0.0
0.886740329956 -0.460492053574 0.0404803140288
0.460358179338 0.887636243818 0.0131242286837
-0.0419753969112 0.00699766079195 0.999094139107
-108.909030413 -4.48399991257 -226.592524204
2888.88888889 0.0 0.0
```

pierotofy:
Ok, so only the k1,k2 parameters aren't set for this case. I propose to rewrite GeorefCamera::isValid() to check for "0"s in the transform matrix rather than k1,k2:

```
bool GeorefCamera::isValid()
{
    return focalLength_ != 0 && k1_ != 0 && k2_ != 0;
}
```
Change it to `return focalLength_ != 0 && 
transform(0,0) != 0 && 
transform(0,1) != 0 && 
transform(0,2) != 0 && 
transform(1,0) != 0 &&
transform(1,1) != 0 &&
transform(1,2) != 0 &&
transform(2,0) != 0 &&
transform(2,1) != 0 &&
transform(2,2) != 0`

(And maybe there's a better syntax to check all 3x3 cells of the matrix)?

dakotabenjamin:
I'll check it out

pierotofy:
Ah, also, a better logic would be:

`focalLength_ != 0 && (transform(0,0) != 0 || transform(0,1) != 0 || transform(0,2) != 0 || transform(1,0) != 0 || transform(1,1) != 0 || transform(1,2) != 0 || transform(2,0) != 0 || transform(2,1) != 0 || transform(2,2) != 0)`

Otherwise a single 0 cell value will classify the camera as invalid.

dakotabenjamin:
What if instead we checked if the camera was listed in list_r000.out? Would that lead to the same result?

pierotofy:
I thought `list_r000.out` only contains the list of images?

dakotabenjamin:
I had trouble accessing the transform_ matrix because I don't know how to access an Eigen matrix. would this work? 

bool GeorefCamera::isValid()
{
    return focalLength_ != 0 &&
           (transform_->operator()(0, 0) != 0 ||
           transform_->operator()(0, 1) != 0 ||
           transform_->operator()(0, 2) != 0 ||
           transform_->operator()(1, 0) != 0 ||
           transform_->operator()(1, 1) != 0 ||
           transform_->operator()(1, 2) != 0 ||
           transform_->operator()(2, 0) != 0 ||
           transform_->operator()(2, 1) != 0 ||
           transform_->operator()(2, 2) != 0);
}

pierotofy:
See this line for an example: https://github.com/pierotofy/OpenDroneMap/blob/62dec4bb95374e24a4934f493f158a730399e0aa/modules/odm_georef/src/Georef.cpp#L907

Since transform_ is a pointer, I think you need to dereference it first:

```
(*transform_)(0, 0) != 0
```

https://github.com/pierotofy/OpenDroneMap/blob/62dec4bb95374e24a4934f493f158a730399e0aa/modules/odm_georef/src/Georef.hpp#L112

pierotofy:
Got to love C/C++ 😄

dakotabenjamin:
These changes are going to be reflected in the split-merge update.

PR #720, opened 2017-11-27
Use latest mve for texturing

dakotabenjamin:
This PR builds mvs-texturing from an OpenDroneMap fork which has an updated version of mve. @fredlllll can you test for me?


fredlllll:
compiled fine. ran it and the exception in mve now contains the filename, so the changes are there

PR #736, opened 2018-01-03
Fast Orthophoto via sparse reconstruction

pierotofy:
This PR adds support for a new `--fast-orthophoto` flag, which skips the opensfm dense reconstruction step and uses the sparse reconstruction to generate an orthophoto, while skipping the generation of the full 3D model (the poisson mesh). This is going to benefit users that just want an orthophoto and works especially well on flat areas such as farm fields.

List of changes:

- Added `--fast-orthophoto`, `--mesh-neighbors` and `--mesh-resolution` parameters.
- Removed old 2.5D mesh module and related flags.
- Wrote new 2.5D mesh program, based on VTK7. We cannot use the binaries that ship with Ubuntu because the parallel features, which are critical for good performance, are not available in the binary packages. VTK7 adds time to the compilation, but doesn't increase image size in docker. The 2.5D mesh works by interpolating a DSM using shepard's method and performing a greedy terrain decimation, which allows us to retain the shape of buildings even with few points and fill-in gaps in the terrain. It's multithreaded, so meshing will be relatively fast especially on machines with multiple cores.

image

image

Fast orthophoto:

image

- Cleaned up code that wasn't being used anymore
- Made learner docker images, most of the files in /code/SuperBuild/src are not needed because they are installed in /code/SuperBuild/install (with a few exceptions)
- Users can continue computing a DSM/DTM from the sparse point cloud when `--fast-orthophoto` is used.
- Increased the `min-num-features` parameter default to `8000`. This is both to help users get better orthophotos using the `--fast-orthophoto` flag, but also to help users get better overall results even when using the dense reconstruction. Often times `4000` isn't sufficient for areas of high vegetation and new users seem confused. If this is not a good idea, please let me know.
- Users can keep using the dense reconstruction for generating the orthophoto and they can choose to use the new 2.5D mesh module. The `--use-25dmesh` flag remains. When `--fast-orthophoto` is used, `--use-25dmesh` is implicitly set.

I recommend testing also with the changes introduced here: https://github.com/OpenDroneMap/mvs-texturing/pull/1

The PR is stable enough to be tested, but do not merge yet. I'd love to hear feedback and do some more testing, especially to make sure that nothing in the existing pipeline broke. 😄


pierotofy:
Note I'm also often getting better results by using `--texturing-data-term area` instead of `gmi`.

dakotabenjamin (commented):
First of all, increasing min-num-features is a good idea. I'm excited to start looking at this :D

LucasMM86:
How can I test this PR? I use ubuntu 16.04 and installed ODM like this: https://github.com/OpenDroneMap/OpenDroneMap/wiki/Installation
I would like to install this PR also.

dakotabenjamin:
```
git checkout -b pierotofy-sparserecon master
git pull https://github.com/pierotofy/OpenDroneMap.git sparserecon
./configure.sh reinstall
```

then run it with the new --fast-orthophoto tag

pierotofy:
Also, just added the `--crop` flag that automatically crops all 2D products (orthophotos, dems) using a smooth boundary polygon computed from the point cloud.

image

Can be turned off by setting it to `0`.

pierotofy:
@smathermather

dakotabenjamin:
I ran into an issue with a particularly large dataset: 
```
./run.sh 118 --fast-orthophoto --force-ccd 12.75 --use-hybrid-bundle-adjustment --build-overviews -v --time
.
.
.
[INFO]    Running ODM OpenSfM Cell - Finished
[INFO]    Running ODM Meshing Cell
[DEBUG]   Writing ODM 2.5D Mesh file in: /hdd/sterling/118/odm_meshing/odm_25dmesh.ply
[DEBUG]   running /home/dmb2/opendronemap/build/bin/odm_25dmeshing -inputFile /hdd/sterling/118/opensfm/reconstruction.ply -outputFile /hdd/sterling/118/odm_meshing/odm_25dmesh.ply -logFile /hdd/sterling/118/odm_meshing/odm_25dmeshing_log.txt -maxVertexCount 100000 -neighbors 24 -resolution 10 -verbose
Loading point cloud... loaded 2302016 points
Remove outliers... 3 points removed
Squash point cloud to plane... OK
Plane extentX: 4339.53, extentY: 3569.08
DSM size is 43396x35691 (6196 MB) 
Point interpolation using shepard's kernel... Segmentation fault (core dumped)
  File "/home/dmb2/opendronemap/run.py", line 47, in <module>
    plasm.execute(niter=1)
  File "/home/dmb2/opendronemap/scripts/odm_meshing.py", line 105, in process
    '-resolution {resolution} {verbose}'.format(**kwargs))
  File "/home/dmb2/opendronemap/opendm/system.py", line 34, in run
Exception: Child returned 139
```
I'm double checking it isn't a memory issue but I don't think that's the case.

dakotabenjamin:
Nevermind it's probably an issue with the dataset itself

pierotofy:
Were you able to process the dataset then? Found any problems? I'm wondering if sometimes the size of the DSM is too large (since we end up approximating the mesh with a limited number of vertices anyway) and the program should attempt to use a different formula to calculate the DSM resolution (instead of allowing users to pass it via --mesh-resolution).

dakotabenjamin:
No I didn't get it to finish. I did get through another large set but it's got problems too. The problems are different but may have similar causes. 

In both datasets, the ply output has at least one outlier that is massively far away from most of the points. 
image

I got through texturing but it's wonky (I didn't adjust the resolution so that's probably why):
image

the error:
```
./run.sh Talisay --fast-orthophoto --use-hybrid-bundle-adjustment --build-overviews -v --time
.
.
.
[DEBUG]   Line: WGS84 UTM 51N
[DEBUG]   running /home/dmb2/opendronemap/SuperBuild/build/pdal/bin/pdal pipeline -i /home/dmb2/ODMProjects/Talisay/odm_georeferencing/las.json --readers.ply.filename=/home/dmb2/ODMProjects/Talisay/odm_25dgeoreferencing/odm_georeferenced_model.ply --writers.las.filename=/home/dmb2/ODMProjects/Talisay/odm_georeferencing/odm_georeferenced_model.las
PDAL: writers.las: Unable to convert scaled value (-5.75718e+10) to int32 for dimension 'X' when writing LAS/LAZ file /home/dmb2/ODMProjects/Talisay/odm_georeferencing/odm_georeferenced_model.las.

  File "/home/dmb2/opendronemap/run.py", line 47, in <module>
    plasm.execute(niter=1)
  File "/home/dmb2/opendronemap/scripts/odm_georeferencing.py", line 172, in process
    tree.odm_georeferencing_las_json)
  File "/home/dmb2/opendronemap/opendm/types.py", line 222, in convert_to_las
    '--writers.las.filename={f_out}'.format(**kwargs))
  File "/home/dmb2/opendronemap/opendm/system.py", line 34, in run
Exception: Child returned 1
dmb2@opendronemap-gui:~/opendronemap$ 
```



I wanted to check the ply but for some reason it's saved in binary format now?

pierotofy:
Would you be able to share the input PLY (`reconstruction.ply`) so that I can run some tests?

dakotabenjamin:
check your email, I shared a folder with you.

pierotofy:
Mm, hasn't arrived. Did you share with the @masseranolabs.com address? The other one has some issues receiving mail.

Could you also share the files necessary for texturing? :+1:

dakotabenjamin:
Ah, ok. Sent. It might take a while for the texturing files to upload.

pierotofy:
How was `talisay-reconstruction.ply` generated? I don't think this came out of OpenSfM. Having issues opening it or even converting it to another format. PDAL gives this out: `pdal translate talisay-reconstruction.ply talisay.las
PDAL: writers.las: Unable to convert scaled value (-5.77430208e+10) to int32 for dimension 'X' when writing LAS/LAZ file talisay.las.`

pierotofy:
For `zone118-reconstruction` there's a point so far out of range that messes up the statistical outlier removal filter (that's why only 3 points are removed).

`
3362775.22763 -930810.489233 -16320535.0841 41 86 62
`

I've added a very relaxed radius outlier removal step prior to it that should take care of such occurrences.

pierotofy:
Also I've changed the locator to an octree locator, which is a lot faster for large datasets.

pierotofy:
Made some new changes, if the user sets `--mesh-resolution` to `0` (the default now), the 2.5D mesh program automatically makes a good guess as to what a reasonable resolution value should be, based on the target amount of vertices and the area extent. I realized it makes no sense to create a super high definition DSM to just approximate it with a small number of vertices (which will have an error anyway).

Along with the speed enhancements, this should now work with large datasets. If buildings or cars come out "deformed" while texturing, the suggestion is to further increase `min-num-features`, or increase `mesh-size`.

To test new changes (if you've already built the entire branch once):

```bash
git pull <myrepo> sparserecon
cd /code/build
make odm_25dmeshing
```

smathermather:
Playing with building this into a docker image, it builds fine, but I'm having trouble running on the Toledo dataset:

```docker build -t sparserecon .```

```docker run -it --rm -v $(pwd)/images:/code/images -v $(pwd)/odm_orthophoto:/code/odm_orthophoto -v $(pwd)/odm_texturing:/code/odm_texturing sparserecon --fast-orthophoto```

```
...
Whole texturing procedure took: 191.831s
[INFO]    Running ODM Texturing Cell - Finished
[INFO]    Running ODM Georeferencing Cell
[DEBUG]   None
[WARNING] No coordinates file. Generating coordinates file: /code/odm_georeferencing/coords.txt
[INFO]    Running georeferencing with generated coords file.
[DEBUG]   running /code/build/bin/odm_georef -bundleFile /code/opensfm/bundle_r000.out -inputCoordFile /code/odm_georeferencing/coords.txt -inputFile /code/odm_texturing_25d/odm_textured_model.obj -outputFile /code/odm_texturing_25d/odm_textured_model_geo.obj -inputPointCloudFile /code/opensfm/reconstruction.ply -outputPointCloudFile /code/odm_25dgeoreferencing/odm_georeferenced_model.ply  -logFile /code/odm_25dgeoreferencing/odm_georeferencing_log.txt -outputTransformFile /code/odm_25dgeoreferencing/odm_georeferencing_transform.txt -georefFileOutputPath /code/odm_25dgeoreferencing/odm_georeferencing_model_geo.txt
/code/build/bin/odm_georef: error while loading shared libraries: libvtkIOMPIImage-6.2.so.6.2: cannot open shared object file: No such file or directory
  File "/code/run.py", line 47, in <module>
    plasm.execute(niter=1)
  File "/code/scripts/odm_georeferencing.py", line 159, in process
    '-logFile {log} -outputTransformFile {transform_file} -georefFileOutputPath {geo_sys}'.format(**kwargs))
  File "/code/opendm/system.py", line 34, in run
Exception: Child returned 127
```

**Comment by arn13** [comment]
This looks very promising especially for use in the field.
However running "./configure.sh reinstall" resulted in various errors, which did not appear on a second try. But when I run ./run.sh --fast-orthophoto odm_data_aukerman
usage: run.py [options] <project name>
run.py: error: unrecognized arguments: --fast-orthophoto
This happens.
I am posting all of the output, because I am clueless o what could be the issue and there is a multitude of errors.
The ODM installation is as new as yesterday. My CMakeError.log says:

```
Determining if the pthread_create exist failed with the following output:
Change Dir: /home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_9c968/fast"
/usr/bin/make -f CMakeFiles/cmTC_9c968.dir/build.make CMakeFiles/cmTC_9c968.dir/build
make[1]: Verzeichnis „/home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp“ wird betreten
Building C object CMakeFiles/cmTC_9c968.dir/CheckSymbolExists.c.o
/usr/bin/cc     -o CMakeFiles/cmTC_9c968.dir/CheckSymbolExists.c.o   -c /home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp/CheckSymbolExists.c
Linking C executable cmTC_9c968
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_9c968.dir/link.txt --verbose=1
/usr/bin/cc       CMakeFiles/cmTC_9c968.dir/CheckSymbolExists.c.o  -o cmTC_9c968 -rdynamic 
CMakeFiles/cmTC_9c968.dir/CheckSymbolExists.c.o: In Funktion `main':
CheckSymbolExists.c:(.text+0x16): Nicht definierter Verweis auf `pthread_create'
collect2: error: ld returned 1 exit status
CMakeFiles/cmTC_9c968.dir/build.make:97: die Regel für Ziel „cmTC_9c968“ scheiterte
make[1]: *** [cmTC_9c968] Fehler 1
make[1]: Verzeichnis „/home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp“ wird verlassen
Makefile:126: die Regel für Ziel „cmTC_9c968/fast“ scheiterte
make: *** [cmTC_9c968/fast] Fehler 2

File /home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp/CheckSymbolExists.c:
/* */
#include <pthread.h>

int main(int argc, char** argv)
{
  (void)argv;
#ifndef pthread_create
  return ((int*)(&pthread_create))[argc];
#else
  (void)argc;
  return 0;
#endif
}

Determining if the function pthread_create exists in the pthreads failed with the following output:
Change Dir: /home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_a0096/fast"
/usr/bin/make -f CMakeFiles/cmTC_a0096.dir/build.make CMakeFiles/cmTC_a0096.dir/build
make[1]: Verzeichnis „/home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp“ wird betreten
Building C object CMakeFiles/cmTC_a0096.dir/CheckFunctionExists.c.o
/usr/bin/cc    -DCHECK_FUNCTION_EXISTS=pthread_create   -o CMakeFiles/cmTC_a0096.dir/CheckFunctionExists.c.o   -c /usr/share/cmake-3.5/Modules/CheckFunctionExists.c
Linking C executable cmTC_a0096
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_a0096.dir/link.txt --verbose=1
/usr/bin/cc   -DCHECK_FUNCTION_EXISTS=pthread_create    CMakeFiles/cmTC_a0096.dir/CheckFunctionExists.c.o  -o cmTC_a0096 -rdynamic -lpthreads 
/usr/bin/ld: cannot find -lpthreads
collect2: error: ld returned 1 exit status
CMakeFiles/cmTC_a0096.dir/build.make:97: die Regel für Ziel „cmTC_a0096“ scheiterte
make[1]: *** [cmTC_a0096] Fehler 1
make[1]: Verzeichnis „/home/arn/OpenDroneMap/build/CMakeFiles/CMakeTmp“ wird verlassen
Makefile:126: die Regel für Ziel „cmTC_a0096/fast“ scheiterte
make: *** [cmTC_a0096/fast] Fehler 2`
```

**Comment by pierotofy** [comment]
@smathermather should be fixed now. Turns out I couldn't remove that libvtk6 package... ups. If you don't want to rebuild everything, just issue:

```
docker run -it --entrypoint=/bin/bash sparserecon
apt-get update && apt-get install -y libvtk6-dev
exit
docker commit $(docker ps -aq | head -1) sparserecon2
```

Then run `docker run -it --entrypoint="/code/run.sh" --rm -v $(pwd)/images:/code/images -v $(pwd)/odm_orthophoto:/code/odm_orthophoto -v $(pwd)/odm_texturing:/code/odm_texturing sparserecon2 --fast-orthophoto code`

**Comment by pierotofy** [comment]
@DrArn I'm not entirely sure why you're getting those errors, but I'd recommend to use docker.

**Comment by arn13** [comment]
@pierotofy 
Thanks.
I am running Linux mint natively on my notebook and I had some trouble setting up docker on my win10 machine, guess I will have to try again.
edit: I have set op a working docker image now using:
`docker run -it --rm -v $(pwd)/images:/code/images -v $(pwd)/odm_orthophoto:/code/odm_orthophoto -v $(pwd)/odm_texturing:/code/odm_texturing opendronemap/opendronemap`
How do I use your branch now?
`docker build -t sparserecon .`
This did not work.

**Comment by dakotabenjamin** [comment]
I appear to have produced a y offset when running odm_data_aukerman:
image

Do you see how the two trees don't line up? The south tree is on the mesh of the north tree and the path runs through the NW tree.

Relevant parameters used:
-- texturing-data-term area
--mesh-size 200000
--fast-orthophoto

Note that the orthophoto lines up perfectly, this appears to only affect the textured mesh.

**Comment by pierotofy** [comment]
Thanks @dakotabenjamin, I'll take a further look. Could you share the `reconstruction.ply` file and texturing files?

**Comment by pierotofy** [comment]
I think I got it... I didn't realize that by adding radius outlier filtering I was modifying the point cloud's extent, thus wrongly offsetting a later translation to align the mesh to the original coordinate system.

Please pull the latest changes, recompile the odm_25dmeshing module and try again. Should align properly this time.

**Comment by dakotabenjamin** [comment]
It is currently stuck here: 
```
Point interpolation using shepard's kernel... 
```
just sitting there for many minutes. We'll see if it finished tomorrow.

**Comment by pierotofy** [comment]
I've made another change, which should speed up that step.

**Comment by pierotofy** [comment]
*sigh* never mind... I actually do not know why the coordinates are displaced. I'm also not sure why it's taking long to process the shepard's kernel step.

If the problem persists, please attach:
* `reconstruction.ply`
* `odm_25dmesh.ply`
* mvs_texturing files, cameras and images.

Thanks.

**Comment by dakotabenjamin** [comment]
@pierotofy I ran again this morning with the latest changes and it processes at a normal pace, but still has the offset. Do you still have access to the OneDrive folder I shared with you? Please check that for the files under "aukerman-offset".

**Comment by pierotofy** [comment]
Mm, something's odd. Running the 2.5D module on my arch installation leads to a properly translated mesh. Running the same 2.5D module compiled in the docker environment shifts it by a few meters.

animated

Investigating...

**Comment by pierotofy** [comment]
I'm still puzzled by the behavior, but I think it has to do with VTK's internals (the docs say that `getCenter()` and `getBounds()` are not thread safe, so maybe it has something to do with it). Either way, I was able to get consistent behavior on both my native environment and docker, so give it a try, see if the problem is gone with the changes introduced in `d366ebd`.

It takes about 50 seconds to compute a mesh on my dual core laptop:

```
root@100ce56443a4:/code/build/bin# time ./odm_25dmeshing -inputFile /datasets/test/reconstruction.ply -outputFile /datasets/test/out2.ply -verbose
Loading point cloud... loaded 66334 points
Remove outliers... 4267 points removed
Squash point cloud to plane... OK
Automatically set resolution to 4.226116
Plane extentX: 366.696259, extentY: 271.448975
DSM size is 1550x1148 (8.000000 MB) 
Point interpolation using shepard's kernel... OK
Transfering interpolation results to DSM... OK
Median filter...OK
Triangulate... OK
Transform... OK
Saving mesh to file... OK

real	0m49.967s
user	2m22.045s
sys	0m0.916s
```

dakotabenjamin:
It's working well now. I'm gonna put up a visualization of the processing times once I get another dataset processed.

dakotabenjamin:
Something else seems off here: 
screenshot from 2018-01-17 14-54-41

Is this the dimensions? `10590451P x 3564788L`
The original tif is 39280x23722

pierotofy:
How does `odm_georeferenced_model.bounds.shp` look like in QGIS? Can you share your `reconstruction.ply` file?

dakotabenjamin:
It's much larger than the orthophoto. 

image

pierotofy:
Could you please share:
 * `odm_georeferencing/odm_georeferenced_model.bounds.shp`
 * `reconstruction.ply`
 * `odm_georeferencing/odm_georeferenced_model.las`
 * The original orthophoto

This is most likely due to rogue points coming out of OpenSfM... they need to be filtered.

pierotofy:
Could I also get a copy of `odm_georeferencing/odm_georeferenced_model.las`?

pierotofy:
The LAS file is now first decimated and then filtered (with very relaxed parameters) prior to computing boundaries. I've done a few tests (including the one you attached) and seems to perform well.

Before: image

After: image

Thanks for all of your help testing this PR @dakotabenjamin.

To test simply pull the latest changes (no recompilation needed).

pierotofy:
Didn't test the crop on the orthophoto, but the coverage area seems about right:

image

dakotabenjamin:
image


Parameters: (ran once with and without `--fast-orthophoto`)
Aukerman: `./run.sh aukerman -v --time --mesh-size 200000 --texturing-data-term area `

Zoo: `./run.sh zoo -v --time --texturing-data-term area --rerun-from odm_orthophoto --opensfm-processes 8 --use-hybrid-bundle-adjustment --crop 0`

dakotabenjamin:
How do we feel about merging?

pierotofy:
If you're OK with it, I'm OK with it.

dakotabenjamin:
@pierotofy have you ran into these docker build issues? https://github.com/OpenDroneMap/OpenDroneMap/commit/0fdea3d145a3818c5ae729538155ab9b716e5826#commitcomment-26958472

Issue #738, opened 2018-01-05
Remove sudo statements from configure.sh


tomasholderness:
Currently, configure.sh prefixes individual commands with `sudo` which breaks in environments where `sudo` is not installed and user is already `root`.

I would propose removing `sudo` prefixes from the script and updating documentation to advise users to do `sudo configure.sh install` in environments where it is required.

An example of the configure.sh script without `sudo` statements can be seen [here](https://github.com/OpenDroneMap/OpenDroneMap/commit/53cf8f62fd4d2e6dea1b49d1364fb959e04a8b1f).

dakotabenjamin:
This is a good idea. 

I've always felt weird using `sudo` and on top of that installing dependencies like we do might anger some potential users.

PR #742, opened 2018-01-11
Fix for #738 

tomasholderness:
- remove sudo statements from config.sh


dakotabenjamin (approved):
Once the conflicts are handled I will merge.

PR #743, opened 2018-01-16
Classified point clouds and better DEMs code

pierotofy:
To review and merge after #736.

This branch takes care of the following:
 * Removed the `lidar2dems` module, which was making the code base more complex even though we used only a small subset of its functionality.
 * Added a new `--pc-classify` flag which can take one of `none`, `smrf` or `pmf`. When `smrf` is set, a simple morphological filter is used to classify points in `odm_georeferenced_model.las`. When `pmf` is set, a progressive morphological filter is used instead. When `--dtm` is set, `--pc-classify` is automatically set to `smrf` unless otherwise specified by the user (but cannot be `none`).

This allows users to do basic classification of ground vs. non-ground returns (even when no DTM is being generated, which might be desirable).

image


pierotofy:
Along with the recent changes to node-OpenDroneMap, this is ready to be tested.

When the point cloud is classified, it can be displayed within WebODM. The same point cloud is used for the DSM/DTM generation.

Please let me know if anything should be changed, or if something doesn't work as expected.

Pinging @smathermather for this also.

dakotabenjamin:
I will get to this Monday

pierotofy:
If there are no objections, I'd like to merge this by the end of the week, since among adding the new features it would also solve https://github.com/OpenDroneMap/WebODM/issues/401 and #762 which is currently stopping users from generating a DTM.

dakotabenjamin:
Can you edit the settings.yaml files as well? After that we can merge, and then edit the wiki with the new run-time parameters.

pierotofy:
Modified both .yaml files.

smathermather:
Looking forward to testing!

PR #753, opened 2018-01-26
Split merge / Support opensfm georeferencing

dakotabenjamin:
This PR pulls the latest OpenSfM., which implements a georeferencing feature. I've refactored the odm_georeferencing module/python code a bit while incorporating this new feature. This includes
- improved ODM_Reconstruction and ODM_Georef classes
- better projection support
- new parameter `--proj <projection>` that supports most proj4 strings (see bullet below)
- implicit support for split-merge

Do not merge yet, I still need to work on a few things:
- split-merge interface
- gcps are acting wonky
- Can't handle projections in latlong or that have imperial units (eg. state plane)
- Things I've forgotten since it's Friday


dakotabenjamin:
ugh im gonna fix this on monday

opendm/types.py
```diff
@@ -334,9 +451,8 @@ def __init__(self, root_path, images_path):
         self.odm_georeferencing_latlon = io.join_paths(
             self.odm_georeferencing, 'latlon.txt')
         self.odm_georeferencing_coords = io.join_paths(
-            self.odm_georeferencing, 'coords.txt')
-        self.odm_georeferencing_gcp = io.join_paths(
-            self.odm_georeferencing, 'gcp_list.txt')
+            self.root_path, 'coords.txt') # Todo put this somewhere better
+        self.odm_georeferencing_gcp = io.find('gcp_list.txt', self.root_path)
```
pierotofy: This should probably be set to `args.gcp` (as when passing `--gcp <path>` to `run.py`). The GCP file is not always named `gcp_list.txt`.

pierotofy: Or at least I didn't think so; is the `--gcp` parameter even used across the solution?

scripts/odm_georeferencing.py
```diff
@@ -22,64 +22,27 @@ def declare_params(self, params):
...
         # Benchmarking
         start_time = system.now_raw()
 
         log.ODM_INFO('Running ODM Georeferencing Cell')
 
         # get inputs
-        args = self.inputs.args
-        tree = self.inputs.tree
-        gcpfile = io.join_paths(tree.root_path, self.params.gcp_file) \
-            if self.params.gcp_file else find('gcp_list.txt', tree.root_path)
```
pierotofy, line 45: Removing this will not allow users to specify a different gcp path via `--gcp`.

pierotofy (commented):
I'm excited to try these features out! The only thing I noticed is the GCP path handling. Great work.

opendm/types.py
dakotabenjamin: That looks like something I hard-coded and didn't fix. Nice catch!

pierotofy (approved):
Looks :+1:  :clap: 

I wonder if using a virtual raster would give better performance compared to gdal_merge, but that's something for a future PR.

dakotabenjamin:
I think ideally I'd like to just have tiles- I don't like the idea of producing single massive files. But I'm not sure the best way to go about that.

smathermather:
w00t!

PR #756, opened 2018-02-01
Updated Dockerfile to create a smaller image

arkadyark:
Original image size: 4.9GB
With these changes: 2.55GB

Every additional RUN command created another layer, and they're cumulative, so the final commands that were cleaning up build dependencies weren't actually doing anything. The resulting Dockerfile is a bit harder to read, so perhaps it should be kept alongside it, as compact.Dockerfile or something like that, rather than replacing the existing one completely. Your call.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,33 +2,11 @@ FROM phusion/baseimage
 
 # Env variables
 ENV DEBIAN_FRONTEND noninteractive
-
-#Install dependencies
-RUN apt-get update -y
-RUN apt-get install software-properties-common -y
-#Required Requisites
-RUN add-apt-repository -y ppa:ubuntugis/ppa
-RUN add-apt-repository -y ppa:george-edison55/cmake-3.x
-RUN apt-get update -y
-
-# All packages (Will install much faster)
-RUN apt-get install --no-install-recommends -y git cmake python-pip build-essential software-properties-common python-software-properties libgdal-dev gdal-bin libgeotiff-dev \
-libgtk2.0-dev libavcodec-dev libavformat-dev libswscale-dev python-dev python-numpy libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libjasper-dev libflann-dev \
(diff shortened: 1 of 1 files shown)


pierotofy:
I think we should replace it, this is great.

pierotofy:
I wonder what the final compressed size of the image is when published on docker hub? https://hub.docker.com/r/opendronemap/opendronemap/tags/ the current one stands at ~2GB.

arkadyark:
https://hub.docker.com/r/arkadyark/opendronemap-small/tags/ it's 773MB, big reduction.

pierotofy:
I did some more research in this, apparently there's an experimental feature in docker that now allows to squash layers while retaining multiple RUN commands (for better readability).

`docker build --squash -t opendronemap/opendronemap .`

Experimental features need to be enabled to use the `--squash` flag. https://stackoverflow.com/questions/44346322/how-to-run-docker-with-experimental-functions-on-ubuntu-16-04

If this works we might want to just build images using the squash flag. Thoughts?

pierotofy:
Using the `--squash` flag I was able to generate a `794 MB` image. https://hub.docker.com/r/opendronemap/opendronemap/tags/

arkadyark:
If this can be done in the pipeline that's used to push builds to dockerhub, then this is the better solution for readability's sake. I would also suggest it be added in the README when instructions for building are given, maybe with some explanation or a link explaining what that flag does?

pierotofy:
Agreed. Would you be interested in writing up such instructions and open a PR? Otherwise I can add them when I have a second.

This is the full command I've used:

`docker build --squash -t opendronemap/opendronemap:latest .`

arkadyark:
Sure, can do. Should have some time to do so tomorrow. When I do that I'll close this PR and link to the new one I create.

pierotofy:
We'll wait for the new PR, in the meanwhile I'll close this one. :+1:

arkadyark:
Sorry for the delay, finally got around to it now. Let me know if you have any suggestions!

PR #763, opened 2018-02-16
Add update command for vtk6

dakotabenjamin:
@pierotofy can you check this? Fix for @talltom docker build problems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ COPY VERSION /code/VERSION
 RUN cd SuperBuild && mkdir build && cd build && cmake  .. && make -j$(nproc)     && cd ../.. && mkdir build && cd build && cmake .. && make -j$(nproc)
 
 RUN apt-get -y remove libgl1-mesa-dri git cmake python-pip build-essential
-RUN apt-get install -y libvtk6-dev
+RUN apt-get update && apt-get install -y libvtk6-dev
 
 # Cleanup APT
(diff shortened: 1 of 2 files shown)


pierotofy:
This is actually redundant, apt is already updated at the beginning of the Dockerfile (https://github.com/OpenDroneMap/OpenDroneMap/blob/67e994d11703399725d9fe0efd97b36390403b02/Dockerfile#L7).

The problem you experience is caused by the fact that docker creates intermediate layers and caches those layers for subsequent builds (so you might end up resuming the build process from a layer that has an old apt cache).

You can solve this by adding the `--no-cache` flag to the build process.

`docker build --no-cache -t opendronemap/opendronemap .`

I don't think we should add these changes (but we should add this information in the README).

pierotofy (changes requested):
See comment above.

dakotabenjamin:
I see. It would still be nice to not have to rebuild the entire image because of a caching issue. 

Why are we not putting all of the package installs in the same step?

pierotofy:
Most of them already are. Notable exception is libvtk6-dev https://github.com/OpenDroneMap/OpenDroneMap/blob/67e994d11703399725d9fe0efd97b36390403b02/Dockerfile#L56 which for some reason is removed when we issue the remove in the step above it.

There's certainly room for improvement.

dakotabenjamin:
Got it.

Issue #508, opened 2017-03-08
Add ID number to GCP format


smathermather:

dakotabenjamin:
Combine with #507 and awaited GCP updates. Also we should work on the GCPoint object.

yjmenezes:
i think, it would be good to have or a flag (On/Off or XY,Z,XYZ ) or standard deviation for XY and Z. This way we could turn points ON/OFF by flag or setting point weight.

column 1 # signal to turn a point into a comment, or to insert comments inside the GCP file.

dakotabenjamin:
I like the idea of adding deviations or other error values, but I'm not sure on the implementation.

yjmenezes:
@dakotabenjamin 

I hope  it will be not so hard to implement.
The weight matrix W can be a diagonal matrix derived from point standard deviation.

Please, take a look in your spare time. Chapter4 and 9

www.mygeodesy.id.au/documents/Chapter%204.pdf
www.mygeodesy.id.au/documents/Chapter%209.pdf

I was looking GRASS but no weight is used, just a status to turn GCP On/Off:
https://grass.osgeo.org/grass72/manuals/addons/v.ply.rectify.html

yjmenezes:
@dakotabenjamin , @smathermather 

please, take a look, an alternative approach for affine3D. Just an idea.
i wrote a shell script and an octave script;
octave path must be edited inside shell script.

http://cartognu.org/odm
odm_affine3d_2017-04-16.tar.bz2

thx

yjmenezes:
@dakotabenjamin 

thank you.

Please, if you have a free time, i process boruzsyn with alternative approach for affine3D.
cartognu.org/odm/odm_boruzsyn_affine3d_2017-04-22.tar.bz2
take a look on HV04 screen shots and checksol.txt  pto ( 10,11,12  ).

cheers

dakotabenjamin:
When I get a chance, I will do that. It looks very interesting, but I'll have to read up a bit.

PR #764, opened 2018-02-17
ODM 0.3.1 2018-02-17: fix Georef pixelX_,pixelY_ from integer to doub…

yjmenezes:
please verify first 3 lines Georef.cpp for  formating log_.


modules/odm_georef/src/Georef.cpp
```diff
@@ -25,7 +29,7 @@ GeorefGCP::~GeorefGCP()
 
 void GeorefGCP::extractGCP(std::istringstream &gcpStream)
 {
-    gcpStream >> x_ >> y_ >> z_ >> pixelX_ >> pixelY_ >> image_;
+    gcpStream >> x_ >> y_ >> z_ >> pixelX_ >> pixelY_ >> image_ >> idgcp_;
```
pierotofy: I think this would cause some issues if I use a GCP file which does not include IDs.

modules/odm_georef/src/Georef.hpp
```diff
@@ -46,10 +46,11 @@ struct GeorefGCP
 
     size_t cameraIndex_;    /**< The index to the corresponding camera for the image. **/
 
-    int pixelX_;            /**< The pixels x-position for the GCP in the corresponding image **/
-    int pixelY_;            /**< The pixels y-position for the GCP in the corresponding image **/
+    double pixelX_;            /**< The pixels x-position for the GCP in the corresponding image **/
```
pierotofy, line 52: Good catch, since GCPs support floats we have been effectively truncating the precision here.

pierotofy (commented):
I think this is mostly good, could you please explain a bit about the purpose of adding IDs to each GCP?

We should also support older GCPs that don't use this new format (see comments).

modules/odm_georef/src/Georef.cpp
yjmenezes: @pierotofy 
The idgcp_ can help on large datasets or can be used to label GCPs over Ortomosaic.
I upload a QGIS image.
We can comment and remove it to maintain compatibility with old gcp_list.txt layout but i think idgcp_ really can help.

pierotofy: Indeed! I'm not saying we should remove it, I'm saying that we should support both formats. Some people might not need to label their GCPs.

yjmenezes: @pierotofy ,

i think a try:{...}  inside
void GeorefGCP::extractGCP(std::istringstream &gcpStream)
could solve the idgcp_ issue keeping both layouts.
What do you think ?

pierotofy: Only one way to find out. Could you compile the changes and test the odm_georef module with both varieties?

yjmenezes: @pierotofy 

thx for your comments.
Ok, I will compile and test, this weekend.

yjmenezes:
@pierotofy 

Georef.cpp  code tested, with both layout , seems to be Ok.

I wrote some notes about docker.     odm_docker_readme.txt

dakotabenjamin:
I apologize I just pushed changes that updates OpenSfM, and it looks like that brings a few issues. 
- EPSG codes are not currently supported (I'm looking into it)
- The georeferencing step got moved up so some of it happens in OpenSfM now. Running now leads to this error: 
image


I suggest we open an issue with mapillary/OpenSfM to see if we can get this updated on their side.

yjmenezes:
@dakotabenjamin ,  please, do not apologise.

1- Even moving Georef from botton to top, as an enhancement, please consider the  hash symbol as comments and the idgcp_ field inside gcp_list.txt.

2- remember to change PixelX_ to double;

3-  Wiki:  the way docker maps the gcp_list.txt and some  some fix related Toledo GCP.

docker run -it --rm   -v $(pwd)/gcp_list.txt:/code/gcp_list.txt \

4- the Toledo  .../tests/test_data/gcp_list.txt must be fixed too.

yjmenezes:
@dakotabenjamin 

I read @paulinus  opensfm/io.py code;

Adding 3 lines of code (blue), inside

def read_ground_control_points_list(fileobj, reference_lla, exif):
...

 opensfm must run fine with proposed gcp_list.txt layout;

opensfm_io_py

what do you think ?

pierotofy:
Looks like it would work to me. Can you open a pull request in the opensfm repo? https://github.com/mapillary/OpenSfM

dakotabenjamin:
That's great!

dakotabenjamin:
I'm testing this now, ran into a small issue with opensfm after I pulled the latest changes:
https://github.com/mapillary/OpenSfM/issues/297

Once this is fixed, I think we can merge this.

dakotabenjamin:
@yjmenezes the opensfm changes have been made, and we are nearly ready to merge. Can you apply this patch?


```
Index: SuperBuild/cmake/External-OpenSfM.cmake
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- SuperBuild/cmake/External-OpenSfM.cmake	(date 1522180841000)
+++ SuperBuild/cmake/External-OpenSfM.cmake	(date 1522182494000)
@@ -8,8 +8,8 @@
   STAMP_DIR         ${_SB_BINARY_DIR}/stamp
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
-  URL               https://github.com/mapillary/OpenSfM/archive/e3f56375f834822f1e03d37bd9b48777acc092ed.zip
-  URL_MD5           0560754729245891ee8279f7c788b044
+  URL               https://github.com/mapillary/OpenSfM/archive/93ae099862297c36ae94dd56fca1c062aa2bb60d.zip
+  URL_MD5           f0d8ec8a8dc9e0f6fd55f77d5407e50f
   #--Update/Patch step----------
   UPDATE_COMMAND    ""
   #--Configure step-------------
Index: tests/test_data/gcp_list.txt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/test_data/gcp_list.txt	(date 1522180841000)
+++ tests/test_data/gcp_list.txt	(date 1522182494000)
@@ -1,4 +1,4 @@
-EPSG:32617
++proj=utm +zone=17 +ellps=WGS84 +datum=WGS84 +units=m +no_defs 
 ## comment: the first line must contain EPSG. ( mandatory )
 ## 0/0 of Pixel Coord = Upper left corner pixelX(column_0..) pixelY(row_0..)
 ## x1 y1 z1 pixelx1 pixely1 imagename1 idGCP ( idGCP optional )
```

yjmenezes:
@dakotabenjamin 

Sure,  I will, as soon as possible.
thx.

dakotabenjamin:
Thank you! 

(personal reference to #508 )

yjmenezes:
@dakotabenjamin 
thank you,

gcp_list.txt:   
I add the +north to proj4. Would it be a problem to OpenSfm ?

+proj=utm +zone=17 +north +ellps=WGS84 +datum=WGS84 +units=m +no_defs 

I hope i have done in the right way.

dakotabenjamin:
Thank you so much for the contribution @yjmenezes !!!

pierotofy:
Great job all!

PR #771, opened 2018-02-24
Added explanation of squash flag to build section

arkadyark:
As suggested [here](https://github.com/OpenDroneMap/OpenDroneMap/pull/756), updated README to explain how to use the --squash flag.


pierotofy:
Couldn't have written it better! Thanks!!

PR #780, opened 2018-03-08
Sphinx docs

dakotabenjamin:
My first attempt, it's a start!


danbjoseph:
what about putting the docs in their own repository? keep things a little cleaner and you could have some sort of travis-ci or circl-ci build process for the docs site?

doc/source/conf.py
```diff
@@ -0,0 +1,178 @@
...
+project = u'OpenDroneMap'
+copyright = u'2017, OpenDroneMap'
+author = u'OpenDroneMap'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u'0.4'
```
pierotofy, line 61: Should we read this from https://github.com/OpenDroneMap/OpenDroneMap/blob/master/VERSION ?

dakotabenjamin: Yes!

dakotabenjamin:
@danbjoseph I was originally thinking to possibly use readthedocs.io as a hosting solution. Would this be possible with your suggestion?

danbjoseph:
i was originally going to suggest using travis-ci to build the static pages from the master branch and push it to a gh-pages. but seems like readthedocs would probably be easier. would be happy to help set either up.

danbjoseph:
although they do show ads on your site
https://docs.readthedocs.io/en/latest/ethical-advertising.html

dakotabenjamin:
True, that is a concern. @smathermather maybe you could provide input? We may not be able to use that service in that case.

danbjoseph:
i set up this site https://github.com/IFRCGo/ecv-toolkit to use travis-ci to build and push to gh-pages branch. i could do the same thing with sphinx.

dakotabenjamin:
I created a odm-docs repo and added you, if you wanna have a go at it.

danbjoseph:
awesome. will do so this weekend.

danbjoseph:
good to go: https://github.com/OpenDroneMap/odm-docs

smathermather:
We proceeded with travis-ci and not readthedocs, right?

danbjoseph:
we're using a sphinx theme by readthedocs, but creating the static pages using the sphinx-build code as part of a travis-ci process and pushing to gh-pages.
so no ads or anything and the only reference to readthedocs is a link at the bottom of the page.
https://opendronemap.github.io/odm-docs/
screen shot 2018-03-14 at 4 29 01 pm

smathermather:
Ah, that's fine then. Thanks @danbjoseph.

dakotabenjamin:
Thanks Dan!

PR #791, opened 2018-03-25
Modified odm_georef to preserve double coordinates precision

pierotofy:
This fixes the precision issue first reported on https://github.com/OpenDroneMap/WebODM/issues/411.

I've replaced PCL's PLY IO functions with tinyply, because PCL's support for double coordinates precision is insufficient. On the plus side, the code is now faster.

Note that some software is unable to display double precision coordinates (ex: MeshLab), so if you open the resulting PLY file, your point cloud will still look funky.

image

This is not a problem with the output, it's a problem with the software not being able to handle large coordinates (and PLY, unlike LAS, has no support for `offset_x` and `offset_y`).

The derived LAS file as shown in plas.io: 

image

Derivate products like DSM and DTM also look fine:

image


dakotabenjamin (approved):
Thanks Piero! I'm running some tests but this looks good to merge.

hblanken:
We also tested this and it works fine

PR #803, opened 2018-04-09
Fixed odm_georef normals handling, wrong pdal path for 2.5D meshing step

pierotofy:
This is a fix for two separate problems involving the use of `--fast-orthophoto`:
 * Point clouds without normals weren't currently being handled properly by recent addition to odm_georef. Now we support inputs with optional normals and colors.
 * The variable `odm_georeferencing_model_ply_geo` in `odm_georeferencing.py` was being redefined  but without taking in consideration the path of the 2.5D meshing module (which is used by --fast-orthophoto), making the process fail. The variable is properly defined on line 68 and doesn't need to be set again.

:+1:  :beer:


pierotofy:
Will merge this by tomorrow unless there are objections.

dakotabenjamin (approved):
Looks great! Thanks as always Piero!

pierotofy:
Ah, we'll see! Going to wait for the waterfall of bug reports coming in... :smile: Thanks for reviewing the changes.

PR #806, opened 2018-04-19
Fixes to get non-georeferenced images to process correctly

### LINKED ISSUE #799: Using OpenDroneMap without geo data
Hello,
is there way to reconstruct point cloud / mesh from photos usinf ODM without having the gps coords? I don't need absolute position, just get 3D model for manual editing and aligning or 3d printing.

Right now i get this:

```
[WARNING] Could not generate coordinates file. Ignore if there is a GCP file. Error: Error in OdmExtractUtm:
Image is missing GPS Latitude data
For more detailed information, see log file.

[ERROR]   Could not find file /code/odm_georeferencing/coords.txt
```

I just want ODM to ignore this and give me output mesh with relative coords. I don't have gps in txt, gcp or exif. i just don't want to use them at all. I know this is possible with colmap. But i like ODM more.

Thanks for your time.

pierotofy:
A recent update broke the ability to process images that have no EXIF GPS information or GCPs datasets. These changes allow users to process non-georeferenced images again.

Fixes #799, http://community.opendronemap.org/t/could-not-find-file-odm-georeferencing-coords-txt/487

### DIFF
--- a/opendm/types.py
+++ b/opendm/types.py
@@ -132,18 +132,20 @@ class ODM_Reconstruction(object):
     def __init__(self, photos, projstring = None, coords_file = None):
         self.photos = photos    # list of ODM_Photos
         self.projection = None  # Projection system the whole project will be in
+        self.georef = None
         if projstring:
             self.projection = self.set_projection(projstring)
             self.georef = ODM_GeoRef(self.projection)
         else:
             self.projection = self.parse_coordinate_system(coords_file)
-            self.georef = ODM_GeoRef(self.projection)
+            if self.projection:
+                self.georef = ODM_GeoRef(self.projection)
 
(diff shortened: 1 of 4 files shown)


smathermather:
I can run a dataset through. Anything in particular you want looked at, beyond a "yup, this works?"

pierotofy:
It would be good to run a dataset with EXIF geotags and one without. Beside that, I think just a confirmation that things work as expected will suffice.

smathermather:
Cool. Will run momentarily.

smathermather:
Tested with Toledo and COPR datasets. We are a go.

pierotofy:
Thanks @smathermather!

smathermather:
Just ran with a 600+ image dataset and it ran beautifully!

PR #807, opened 2018-04-20
update readme

danbjoseph:
link isn't rendered as a link on http://opendronemap.org/
screen shot 2018-04-20 at 9 47 44 am


danbjoseph:
fix link markdown
screen shot 2018-04-20 at 9 52 10 am

smathermather (approved):
Looks good. We should probably remove the reference to gitter, and only promote the community forum, but I can do that as a separate pull request if you prefer.

PR #813, opened 2018-04-25
Fixed GCPs

pierotofy:
This PR fixes GCP accuracy issues as reported in http://community.opendronemap.org/t/something-changed-between-versions-of-webodm-and-the-use-of-gcp-files/515/13.

In short:
 * OpenSfM supports GCPs, but from my understanding, they are added in the reconstruction as an observation (https://github.com/mapillary/OpenSfM/blob/d4fe825609a6194cca7fd78a618a64a4fe54ed82/opensfm/reconstruction.py#L131) which increases georeferencing accuracy in the point cloud, but does not produce a perfect fit after we mesh it and generate an orthophoto. It's not an error, it's simply the way things are meant to work.
 * odm_georef uses a different method to generate the transform from GCPs: it selects 3 GCPs that minimize the distance between the "best" triplet of GCPs and the resulting location of the points once they are reprojected. This creates a very tight correlation between coordinates and GCPs.
 * I have reintroduced the block that uses the odm_georef GCPs calculation, while making adjustments to support the new way to handle UTM offsets.
 * There was a bug in odm_orthophoto.py if you pass the --rerun-from or --rerun parameter, causing the georef object not to know the proper UTM offset and should be now fixed.

 - All point clouds outputs include the UTM offset.
 - The georeferenced model does not include the UTM offset (that's added via gdal later)
 - `odm_georeferencing_transform.txt` contains the transformation that is applied to the point cloud outputs. Take away the translationX and translationY portions (row 1 col 4, row 2 col 4) and you have the one for the georeferenced model.
 - `odm_georeferencing_model_geo.txt` contains the UTM offset to be applied to the georeferenced model to obtain the orthophoto.
 - The transformation is now being applied to the mesh, which wasn't being applied before.

image
 
This update also upgrades gippy from 0.3.9 to 1.0 since the old version doesn't compile on Ubuntu anymore due to some package upgrades.


smathermather:
Awesome. I will test. any guess as to how it will affect split and merge?

smathermather:
Also, does this support non UTM coordinate systems natively?

pierotofy:
I don't think this will affect split-merge, but I haven't tested a split-merge dataset and if it does, please let me know.

On the non UTM coordinates system, nothing should have changed if you use the EXIF tags (the program logic is the same, things have changed only when you provide a GCP). If we didn't support non UTM coordinates system with GCPs before #753, then I assume there could be problems.

Do you have examples of datasets using a different CRS (with or without GCPs)?

yjmenezes:
@pierotofy 

Do you mind to share your gcp_list.txt and odm_georeferencing_log.txt ?

thank you.

smathermather:
I've merged this 'cause the general use case is more important than the split-and-merge. That said, @paulinus -- do you have a sense offhand if this will break split-and-merge?

yjmenezes:
@pierotofy 

I apologise for writing over a closed PR.

Could you, please, if you have a spare time, take a look here:
https://github.com/mapillary/OpenSfM/issues/302

I think Least Square Adjustment could be interesting to take all available GCP into account.

Issue #769, opened 2018-02-23
Expose some opensfm depthmaps config parameters


pierotofy:
I can take care of this sometimes in the next few weeks (or if somebody else beats me to it).

I propose to expose some of opensfm's depthmaps parameters since I think the defaults sometimes can lead to less than optimal results. In particular, to increase the quality of our meshes and orthophotos, we could really benefit from getting more points in the dense reconstruction. Here are some of my tests with certain parameters:

`depthmap_min_consistent_views`:

![depthmap_min_consistent_views](https://user-images.githubusercontent.com/1951843/36604214-16aceb6e-188b-11e8-8edf-ff413ca2afd6.gif)

Default: 3 ( ~262k points )
Changed: 2 ( ~515k points )
Notes: more noise

`depthmap_resolution`:

![depthmap_resolution](https://user-images.githubusercontent.com/1951843/36604939-2e6f8714-188d-11e8-86c0-73ef1f984752.gif)

Default: 640 ( ~262k points )
Changed: 1280 ( ~1 million points )
Notes: slower by a factor of ~4

`depthmap_patchmatch_iterations`: virtually no change in output, just slower (increased `3` to `9`).

`depthmap_min_patch_sd`: 

Default: 5 ( ~262k points )
Changed: 2.5 ( ~310k points )
Notes: roads appear! same runtime

![depthmap_min_patch_sd](https://user-images.githubusercontent.com/1951843/36605720-8a605c54-188f-11e8-9195-cfcc4719dbbc.gif)

`depthmap_same_depth_threshold`: virtually no change in output (increased `0.005` to `0.015`)

`depthmap_num_matching_views`: virtually no change in output (increased `2` to `4`)

`depthmap_method`:

![depthmap_method_brute](https://user-images.githubusercontent.com/1951843/36607295-2c815f3e-1894-11e8-964d-94392fc29acc.gif)

Default: PATCH_MAP ( ~262k points )
Changed: BRUTE_FORCE( ~330k points )
Notes: Less noise, roads appear, runtime 3x slower

`depthmap_min_correlation_score`:

![depthmap_min_correlation_score](https://user-images.githubusercontent.com/1951843/36608506-10f31880-1898-11e8-89a1-5514cec5e269.gif)

Default: 0.7 ( ~262k points )
Changed: 0.4 ( ~282k points )
Notes: More noise

pierotofy:
Overall I think we could expose:
 - depthmap_min_consistent_views
[...]

PR #822, opened 2018-05-16
Exposed some opensfm parameters

pierotofy:
Takes care of #769 

I would merge after tomorrow's demo, just in case...

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -160,7 +160,7 @@ def config():
                       type=int,
                       default=3,
                       help=('Minimum number of views that should reconstruct a point for it to be valid. Use lower values '
-                            'if your images have few overlap. Lower values result in denser point clouds '
+                            'if your images have less overlap. Lower values result in denser point clouds '
                             'but with more noise. '
                             'Default: %(default)s'))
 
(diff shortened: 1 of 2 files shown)


opendm/config.py
```diff
@@ -147,6 +147,40 @@ def config():
...
+                        help=('Resolution of the depthmaps. Higher values take longer to compute '
+                              'but produce denser point clouds. '
+                              'Default: %(default)s'))
+
+    parser.add_argument('--opensfm-depthmap-min-consistent-views',
+                      metavar='<integer: 2 <= x <= 9>',
+                      type=int,
+                      default=3,
+                      help=('Minimum number of views that should reconstruct a point for it to be valid. Use lower values '
+                            'if your images have few overlap. Lower values result in denser point clouds '
```
dakotabenjamin: I think the grammatically correct word here is less, this is the only nitpicky comment i can make because the rest of this looks great!

dakotabenjamin (approved):

opendm/config.py
pierotofy: We care about grammar, so I'll fix it.

smathermather (approved):
Looking good. Excited to start twisting these new knobs and optimizing values... .

dakotabenjamin:
We should not forget to update the docs, either.

pierotofy:
Pull request opened in the docs.

PR #827, opened 2018-05-23
More friendly message when GCP projection string is not valid

pierotofy:
This should help people fix common GCP header problems (and reduce the number of times people come to the forum to ask how to fix it).


smathermather (approved):
Looks great. I think this will help a lot.

PR #829, opened 2018-05-25
Binary PLY output in odm_georef, remove normals

pierotofy:
This PR changes odm_georef to output the PLY point cloud in binary format instead of ascii, which saves a lot of space.

It also temporarily removes the normals, because we are not transforming them and they shouldn't be included unless they are correctly transformed.


smathermather (commented):
The normals aren't used in the creation of the mesh?

pierotofy:
We don't use the georeferenced point cloud for meshing. I can see end users wanting the normals perhaps for post processing, but we don't currently use them.

Merged depthmap --> Mesh --> Georeferencing.

smathermather:
Ah, sorry. I looked at this too briefly.  Yes, for georeferenced point cloud, if they aren't correctly calculated we shouldn't be passing them.

PR #832, opened 2018-06-04
Faster Georef, CSV export

pierotofy:
This is a set of 5 updates:
 - CSV export has been fixed, as it broke when I introduced binary PLY files in #829 (sorry about that)
 - I've noticed that we are always exporting a CSV file, but I think this should be optional (not everyone needs a CSV file?), so I added a new `pc-csv` flag. Let me know if this needs to be changed.
 - odm_georef has been modified to remove its tinyply dependency, which is quite slow. The PLY format is quite straightforward for writing, so I wrote my own routine which speeds up processing by ~25%
 - I noticed that we've been building all `odm_*` programs without specifying the `CMAKE_BUILD_TYPE` , which has a big effect on performance. The modifications to both CMakeLists.txt makes sure that by default we build for release configuration. I noticed big speed improvements, especially in `odm_georef` (from 12 seconds for a small dataset before, to 5 seconds with the proper CMAKE_BUILD_TYPE setting).
 - I removed the `mesh` command from opensfm. I don't think anyone is using the results of a sparse reconstruction with delaunay triangulation in JSON format (the rest of the pipeline isn't using it) since it's currently wasting time. We can always reintroduce it if we find that somebody needs it?

Cheers! :clinking_glasses:


smathermather (approved):

smathermather (commented):
👍

Issue #839, opened 2018-06-07
Multi-threaded DSM calculation


pierotofy:
Since we use a multi-step approach to DSM generation (via increasing radiuses), we should parallelize the DSM/DTM generation. This is important to speed up the new orthophoto algo as well.

smathermather:
👍

PR #843, opened 2018-06-11
Concurrent DEM calculations

pierotofy:
These changes drastically (like hundreds fold) decrease the runtime of calculating DSM/DTM products. PDAL invocations have been made concurrent and the gap filling algorithm has been replaced from the slower `griddata` to `distance_transform_edt`.

Should be merged after #842 

Takes care of #839

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -1,8 +1,11 @@
 import os, glob
 import gippy
 import numpy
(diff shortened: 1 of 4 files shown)


sorentorp:
Any update on this? I am very sad to only see one of my 4 cores working :-)

PR #850, opened 2018-06-18
Replayed laz support commit

pierotofy:
Fixes the revert commit that was mistakenly applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ When the process finishes, the results will be organized as follows:
         |-- texture_N.jpg                   # Associated textured images used by the model
     |-- odm_georeferencing/
         |-- odm_georeferenced_model.ply     # A georeferenced dense point cloud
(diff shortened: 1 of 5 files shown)


smathermather:
```
  File "run.py", line 47, in <module>
    plasm.execute(niter=1)
  File "/home/gisuser/OpenDroneMap/scripts/odm_georeferencing.py", line 142, in process
    tree.odm_georeferencing_las_json)
AttributeError: 'ODM_Tree' object has no attribute 'odm_georeferencing_las_json'
```

Needs fixed/made consistent in either scripts/odm_georeferencing.py or opendm/types.py . It was easier to revert... .

pierotofy:
At some point I really would like to remove all global "paths" variables from the codebase (when changes happen like this one, they force us to rename not just the filenames they point to, but we need to rename them as well for them not to lose meaning). So it's just extra work.

smathermather:
👍

PR #863, opened 2018-07-01
Froze gippy version to 1.0.0

pierotofy:
This is a fix affecting all DEM products, caused by updates in the `gippy` repository which seems to be broken as of the latest release picked by pip and creates empty GeoTIFFs.

`gippy==1.0.0` also is no longer available (?) from pypi, even though we've been installing it just fine until now.

Installing directly from GitHub brings things back to working.


smathermather:
👍

Issue #880, opened 2018-08-27
libexiv2-dev needs to be updated


pierotofy:
Currently it will not read certain images' EXIF tags (such as ones we captured with a Mavic Pro). This in turns makes the `odm_extract_utm` program to fail. Using to the latest version from Github (not the Ubuntu package) fixes the problem (plus some minor modifications to odm_extract_utm to support the newer version).

smathermather:
👍

pierotofy:
Related: http://community.opendronemap.org/t/an-orthophoto-could-not-be-generated/1018/4

pierotofy:
![image](https://user-images.githubusercontent.com/1951843/45227446-8027eb00-b28e-11e8-8dbf-1af709b1b2a7.png)

We should switch to GExiv2

PR #892, opened 2018-09-13
Libexiv

pierotofy:
This PR addresses the update of libexiv2, for which the Ubuntu prebuilt package is out of date. We now compile it from sources for use in `odm_extract_utm`. It also replaces `pyexiv2` with `exifread` (which is used by OpenSfM, doesn't introduce additional dependencies, and works with the mavic pro images for which we had issues with, cc @smathermather ).

As an added bonus, reading EXIF tags is now a bit speedier.

I've also cleaned up some old code which was not being used anywhere.


pierotofy:
Fixes #880 and https://github.com/OpenDroneMap/WebODM/issues/520

Issue #872, opened 2018-08-02
Docker volumes on Windows sporadically cause file corruptions via odm_georef


pierotofy:
This is a bug that is likely due to some preemptive runtime optimization, or an corrupted memory fetch:

Symptoms: an orthophoto and georeferenced textured model will have a "kaleidoscope-like" effect (similar to https://github.com/OpenDroneMap/OpenDroneMap/pull/774)

An inspection of `odm_textured_model_geo.obj` will show a series of NULL bytes at one of the texture coordinates (using a Hex Editor, look for "0x00 0x00"):

![image](https://user-images.githubusercontent.com/1951843/43591183-bbfaf31c-9640-11e8-8dc6-b1b3feee8200.png)

Re-running the odm_georef program using another CPU generates a valid `odm_textured_model_geo.obj` which can be used to show that the NULL bytes in fact replace a valid texture coordinate.

Output file generated with modern CPU:

![image](https://user-images.githubusercontent.com/1951843/43591339-1f9f54d0-9641-11e8-8df4-fa5639ca47e9.png)

Output file generated with a "working" CPU:

![image](https://user-images.githubusercontent.com/1951843/43591328-1a413922-9641-11e8-813a-1704b4eb449e.png)

While the hypothesis of a hardware bug is bold, it's currently the one that makes sense.

The following lines are responsible for the write of texture coordinates:

```
  for (int m = 0; m < nr_meshes; ++m)
  {
    if(tex_mesh.tex_coordinates.size() == 0)
      continue;

    //PCL_INFO ("%d vertex textures in submesh %d\n", tex_mesh.tex_coordinates[m].size (), m);
    fs << "# " << tex_mesh.tex_coordinates[m].size() << " vertex textures in submesh " << m <<  std::endl;
    for (size_t i = 0; i < tex_mesh.tex_coordinates[m].size (); ++i)
    {
      fs << "vt ";
      fs <<  tex_mesh.tex_coordinates[m][i][0] << " " << tex_mesh.tex_coordinates[m][i][1] << std::endl;
    }
  }
```
https://github.com/OpenDroneMap/OpenDroneMap/blob/master/modules/odm_georef/src/modifiedPclFunctions.cpp#L173

If `tex_mesh.tex_coordinates` weren't properly initialized, we should expect to see at least a 0x20 (space) and "vt" characters in the rogue line, but we don't.

[...]

PR #896, opened 2018-09-21
Replaced ofstream with ostringstream in saveOBJFile

pierotofy:
Fixes the problem documented in #872.

Testing for this PR should only be done to verify that the behavior of `odm_georef` has not changed. You will not be able to reproduce the original problem it fixes unless you have the environment described in #872. I've verified that the problem is resolved in such environment.

Thank you @jtinguely and everyone at Sentera for supporting the work to fix this problem.


dakotabenjamin (approved):
Tested with both exif and gcp datasets with no no issues! Great work!

pierotofy:
Ace, thanks @dakotabenjamin!

PR #898, opened 2018-09-25
add docs.opendronemap.org nearer to top of readme

smathermather:


pierotofy:
Looks good to me! :+1:

PR #900, opened 2018-09-28
Change default smvs-output-scale parameter

pierotofy:
I'd propose to change the default value of `smvs-output-scale` from `2` to `1`.

According to https://github.com/flanggut/smvs/issues/28 it creates more accurate reconstructions.

Running on a simple dataset seem to lead better results.

screen

(Note how the roof of the building is reconstructed with greater detail).


smathermather:
👍

Issue #911, opened 2018-11-13
Processing report (feature request)


brookjason:
Apologys if in the wrong place this is more of a feature request.

Some sot of report similar to this [Trial_2.pdf](https://github.com/OpenDroneMap/ODM/files/2576856/Trial_2.pdf) This is from agisoft photoscan.

smathermather:
This is an excellent idea, and there is a skeleton of work done in OpenSfM to help build this out. Thanks for the request! I think there might be an issue out there already either in WebODM or ODM. If you can find that and link these, that would be really helpful.

brookjason:
> This is an excellent idea, and there is a skeleton of work done in OpenSfM to help build this out. Thanks for the request! I think there might be an issue out there already either in WebODM or ODM. If you can find that and link these, that would be really helpful.

@smathermather , I was unsure as to which one to put it under (ODM or WebODM) I did have a look and search through the ODM and haven't spotted anything like this. I will have a better look in both and link them if I find anything.

saffa19:
This is something I'd really like to see. @brookjason what did you use to generate the graphics for overlap and camera positions/errors?

brookjason:
> This is something I'd really like to see. @brookjason what did you use to generate the graphics for overlap and camera positions/errors?

This was automatic output from Agisoft Metashape formally known as Photoscan

pierotofy:
This is done! :clinking_glasses:

Issue #883, opened 2018-08-31
Output georeferencing error statistics


pierotofy:
A user mentioned it would be useful to have odm_georef output the georeferencing error value, so that georeferencing accuracy can be measured. I think we already calculate it, we just don't output it.

pierotofy:
- When using GCPs this value is in `odm_georef` (`minTotError` variable).
 - When not using GCPs, this value can be computed from OpenSfM's reconstruction.json file, which contains a "reprojection_error" field for each observation. Returning the mean of all observations would probably be sufficient for report purposes.

gastgit:
> * When using GCPs this value is in `odm_georef` (`minTotError` variable).
> * When not using GCPs, this value can be computed from OpenSfM's reconstruction.json file, which contains a "reprojection_error" field for each observation. Returning the mean of all observations would probably be sufficient for report purposes.

I have looked at the reconstruction.json file after an ODM run at opensfm/reports and at /opensfm and these files do not have the field "reprojection_error". I used ODM 1.0.0.
Was this field removed?

pierotofy:
`odm_georef` is no longer in charge of alignment, so this issue should actually be closed.

Any statistics about georeferencing might be exposed via OpenSfM (if any). Not really sure if they are.

PR #912, opened 2018-11-14
Warn about possible malformed GCPs in the log

pierotofy:
This is a code addition to help diagnose GCP issues in the future.

It attempts to detect GCPs that constantly produce a large (more than 10 meters) georeferencing error.

```
WARNING: Detected possible malformed GCP # 73: 168321 large georeferencing errors.
WARNING: Detected possible malformed GCP # 85: 223415 large georeferencing errors.
```

Hope this can be helpful to others. Not sure it's the best approach, but seems to have served the purpose for the analysis in https://github.com/OpenDroneMap/WebODM/issues/549

### DIFF
--- a/modules/odm_georef/src/Georef.cpp
+++ b/modules/odm_georef/src/Georef.cpp
@@ -936,6 +936,15 @@ void Georef::performGeoreferencingWithGCP()
     log_ << "Optimal gcp triplet chosen: ";
     log_ << gcp0 << ", " << gcp1 << ", " << gcp2 << '\n';
     log_ << '\n';
+
+    // Warn about possible bad GCPs
+    for (size_t gcpIndex = 0; gcpIndex < gcps_.size(); ++gcpIndex)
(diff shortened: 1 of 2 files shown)


smathermather:
Is a warning enough in that long log? I can't think of a better solution... .

pierotofy:
Well, we could also output to standard error perhaps. Ideally at some point I'd like to have odm_georef output a JSON file with a report that includes such warnings as well as the transform, mean georeferencing error and other information that could then be useful for display through WebODM.

pierotofy:
Related https://github.com/OpenDroneMap/ODM/issues/883

smathermather:
Also related: https://github.com/OpenDroneMap/ODM/issues/911, which should probably have a sibling feature request in WebODM.

pierotofy:
Since we use OpenSfM's transform for georeferencing, this PR is no longer applicable.

Issue #914, opened 2018-11-21
[Feature Request] DTM/DSM Resolution


smathermather:
It would be nice to have DSM/DTM resolution based on some multiple of the resolution calculated for the orthophoto, e.g. if ODM calculates that the optimum resolution for orthophoto is 5 cm, default to 20 cm for the DSM and DTM resolutions. This is commonly how closed source packages handle this, and of course, like orthophoto resolution would require a flag for override.

pierotofy:
This would be a nice feature. Currently `--dem-resolution` is the same as `--orthophoto-resolution` by default (and both are capped by the GSD estimate ceiling, unless `--ignore-gsd` is used). So you're suggesting that `--dem-resolution` should be a multiple of `--orthophoto-resolution`? Or perhaps it's the interpolation radius that should be incremented? 

See https://github.com/OpenDroneMap/ODM/blob/master/scripts/odm_dem.py#L96 since we currently don't allow a user to specify the radius.

smathermather:
More the GSD ceiling, and it looks like the interpolation radius gets modified automatically as the GSD changes, though the radius might be useful to expose, I'm not jumping to add more flags for people... . 

I put in a [pull request](https://github.com/OpenDroneMap/ODM/pull/915) for doing 4x this. I'll do some further testing -- I've got someteen thousand images to test this on in the next couple of months, but this patch should help others while we discover what these values should be. I think point clouds need some improvements before we can really understand how to really improve these, but at least we can prevent people from having memory problems.

smathermather:
Closed by https://github.com/OpenDroneMap/ODM/pull/915

PR #915, opened 2018-11-22
Dtm dsm gsd patch

smathermather:
Proposed solution to https://github.com/OpenDroneMap/ODM/issues/914 -- reduce default resolution to 1/16 the resolution of the ortho


pierotofy:
:+1: this is a good suggestion. I wonder if we should change the default from `5` to `20` instead of adding a multiple: https://github.com/OpenDroneMap/ODM/blob/master/opendm/config.py#L417

The effect would be the same, but the requested units would be consistent with the output raster resolution.

smathermather:
> 👍 this is a good suggestion. I wonder if we should change the default from `5` to `20` instead of adding a multiple: https://github.com/OpenDroneMap/ODM/blob/master/opendm/config.py#L417

Good point, default should also be the same just to make it explicit. What are we doing with docs ATM? Are we keeping them in sync with master or waiting for releases?

pierotofy:
We're trying to keep them in sync with master.

pierotofy:
This line can probably be reverted if the default is now `20`:

```
resolution = gsd.cap_resolution(args.dem_resolution, tree.opensfm_reconstruction, ignore_gsd=args.ignore_gsd) * 4.0
```

Then I think it's good to be merged!

smathermather:
If we’re autocalculating ortho resolution, I think we should autocalculate DEM resolution unless overridden.

pierotofy:
Mm, I'm a bit confused on the last comment. :thinking: 

Orthophoto resolution is defined by the user. Then it's capped to GSD by default (it's not scaled, but if a user asks for 1 cm / pixel resolution, and GSD's maximum is 3 cm / pixel, the resolution is capped at 3 cm / pixel, unless `--ignore-gsd` is passed, in which case the user can choose any resolution without caps).

The same currently happens for DSM/DTM. `gsd.cap_resolution` is the function in charge of doing that.

The higher the resolution, the higher the memory usage. So increasing the default `--demo-resolution` from `5` to `20` will effectively reduce memory usage.

Or are you suggesting adding further modifications?

smathermather:
So, if the orthophoto resolution is 20cm, we would default to the same resolution for the elevation models? It seems, if the orthophoto resolution for a given project is e.g. 20cm due to gsd.cap_resolution, we don't want the same resolution for the elevation models, as they are calculated from the SMVS output scale which is a 4x4 patch. That would result in a noisier elevation model than necessary.

pierotofy:
Thanks for the explanation! If you want to cap the resolution to a multiple of GSD, you can use the `gsd_error_estimate` parameter and write the call as:

```
resolution = gsd.cap_resolution(args.dem_resolution, tree.opensfm_reconstruction, gsd_error_estimate=-3, ignore_gsd=args.ignore_gsd)
```

See `opendm/gsd.py:55` for how that would work.

With that, we can revert the default of dem-resolution to `5`.

pierotofy (approved):
Looks good to me!

Issue #917, opened 2018-11-24
[Feature] Cache Images Database


pierotofy:
When a dataset is first processed, we extract EXIF information such as focal length, ccd, camera model, image dimension, etc.

This is time consuming. When a dataset is restarted, this process could be sped up a lot by creating a database of images metadata.

Such database could be also useful for other programs to know more about the dataset that was just processed.

pierotofy:
This is done.

PR #918, opened 2018-11-24
Implemented images database

pierotofy:
This PR speeds up reprocessing of an existing dataset by saving on disk an images database (found at `path-to-dataset/images.json` and loading it when available instead of re-reading each individual image.

It also avoid reruns of `odm_extract_utm`, unless rerun parameters are specified.

cc @smathermather 

Completes #917

### DIFF
--- a/opendm/types.py
+++ b/opendm/types.py
@@ -17,7 +17,6 @@ class ODM_Photo:
 
     def __init__(self, path_file, force_focal, force_ccd):
         #  general purpose
-        self.path_file = path_file
         self.filename = io.extract_file_from_path_file(path_file)
         # useful attibutes
         self.width = None
@@ -33,7 +32,7 @@ class ODM_Photo:
         self.longitude = None
         self.altitude = None
         # parse values from metadata
-        self.parse_exif_values(self.path_file, force_focal, force_ccd)
+        self.parse_exif_values(path_file, force_focal, force_ccd)
         # compute focal length into pixels
         self.update_focal()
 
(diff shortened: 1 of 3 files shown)


pierotofy:
The only issue that comes to mind as a result of this change is if the dataset is moved to another directory.

```
    "height": 2250,
    "latitude": 46.842521250000004,
    "longitude": -91.99420961111112,
    "make_model": "dji fc300s",
    "path_file": "/datasets/brighton/images/DJI_0028.JPG",
```

Perhaps the `path_file` variable should be updated, but I don't think we do this for any of the other part of the pipeline (absolute paths stored in other files are stale anyway).

smathermather:
I like this a lot. It will make a big difference with reruns on split-merge reruns where re-reading thousands of images can be really painful.

smathermather:
What is the behavior of it if the images are moved? Does it trigger a read of the exif if the path is changed?

pierotofy:
It does not (only using `--rerun-from dataset` triggers a re-read). `path_file` is only used in one other place (`run_opensfm.py`) outside of the `ODMPhoto` class, so I think I will refactor that reference to always use a relative path to the project's directory. Then there should be no problems.

pierotofy:
Images can now be moved to another folder and not cause issues as no absolute paths are stored in the database.

smathermather:
Tested. Works great and works fine with rerun all, in that it refreshes the image database.

smathermather (approved):
Tested. Works great and works fine with rerun all, in that it refreshes the image database.

PR #921, opened 2018-12-02
m(o)ve over smvs

smathermather:
There are one big limitation: I have replaced smvs with mve, but perhaps it should be optioned instead.


smathermather:
> This is not a complete pull request -- not meant to merge.
> 
> There are one big limitation: I have replaced smvs with mve, but it should be optioned instead.

(Also, it doesn't even work yet.) :)

smathermather:
Ok @pierotofy and @dakotabenjamin -- this will require some heavy testing. mve seems to have lower memory requirements but increased processing time over smvs. I think for the additional fidelity, the processing time is worth the tradeoff (especially in light of reduced memory usage, but we should test on bigger datasets to be sure.

And if we need to have them all run side-by-side, I am open to that too. We could have low, medium, and high fidelity 3D data with opensfm, smvs, and mve point clouds respectively. But, for the sake of simplicity, this pull request is an outright replacement of smvs.

Also, I pass no parameters for mve -- the results are quite nice with defaults. We may want to add parameters later, but I decided to keep this simple: given the quality of the improvement, there seemed sense to not over-complicate.

pierotofy:
I think less is more, so if quality is improved and memory usage is lowered, even though processing time is increased, I think it's a win and smvs can be placed aside for the time being.

Is it ready to be tested then? I'd love to start running some tests.

scripts/mve.py
```diff
@@ -62,45 +54,38 @@ def process(self, inputs, outputs):
...
+                shutil.rmtree(tree.mve)
 
             # run mve makescene
             if not io.dir_exists(tree.mve_views):
-                system.run('%s %s %s' % (context.makescene_path, tree.mve_path, tree.smvs))
+                system.run('%s %s %s' % (context.makescene_path, tree.mve_path, tree.mve))
 
             # config
             config = [
-                "-t%s" % self.params.threads,
```
pierotofy: We should respect the `--max_concurrency` parameter when invoking dmrecon?

smathermather: Ah yes, this reminds me of the other hitch -- I am not sure dmrecon respects concurrency values. It'd have to be set with OMP: https://github.com/simonfuhrmann/mve/blob/master/apps/dmrecon/dmrecon.cc#L285

pierotofy: Interesting, I don't have much experience with OpenMP. Maybe calling `dmrecon` by setting the  `OMP_NUM_THREADS` environment variable could limit them. https://www.ibm.com/support/knowledgecenter/SSGH2K_13.1.3/com.ibm.xlc1313.aix.doc/compiler_ref/ruomprun4.html

smathermather: Yes, I tried that and manually set it to 5 but it used all cores anyway... . But it is c++, there's a good chance I was doing something wrong.

smathermather: This is one I would need help on -- I'm a bit over my head.

smathermather:
It is ready to be tested. It should just install and run.

scripts/mve.py
smathermather: Might just be able to port the approach that smvs uses, if their approaches to concurrency are similar enough.

pierotofy:
I will make some time for this in the next few days.

smathermather:
👍

smathermather:
> I will make some time for this in the next few days.

I am pretty sure I did something wrong here. It usually works as I have it configured, but sometimes failed (I'm sorry I can't be more specific).:
https://github.com/OpenDroneMap/ODM/blob/434b810089f31aa70f98cd35b8bd538304583f64/scripts/odm_app.py#L121

pierotofy:
Testing in progress, will post updates when I get results.

pierotofy:
My findings so far:

 - Sharp edges are much better preserved compared to SMVS, which creates better DSMs.
 - Noise levels are elevated, in particular there's a consistent pattern of points being "raised" from the contours of objects.

image

image

image

 - The number of points generated is often overkill, there's probably a flag to use a lower resolution depthmap and we should take advantage of it.
 - The output from dmrecon uses special characters to display the progress bar, if I recall correctly there's a flag to use more conventional / silent progress output without special characters. We should use that as well.

I'll look for flags that could help us solve one or more of these problems.

pierotofy:
Here's another example:

image

pierotofy:
The flag we should always pass: `--progress=silent`. It defaults to `fancy`.

Should also pass `--force` for good measure.

`--max-pixels` uses the same logic as `smvs`, (the code is the same, so perhaps you can re-add it?). This should reduce both memory usage, speed and perhaps noise.

```
It is rarely useful to reconstruct at full resolution as it will produce less complete depth maps with more noise at a highly increased processing time.
```

https://github.com/simonfuhrmann/mve/wiki/MVE-Users-Guide#multi-view-stereo

All options for reference:
```
Available options: 
  -n, --neighbors=ARG    amount of neighbor views (global view selection)
  -m, --master-view=ARG  reconstructs given master view ID only
  -l, --list-view=ARG    reconstructs given view IDs (given as string "0-10")
  -s, --scale=ARG        reconstruction on given scale, 0 is original
  --max-pixels=ARG       Limit master image size [1500000]
  -f, --filter-width=ARG  patch size for NCC based comparison [5]
  --nocolorscale         turn off color scale
  -i, --image=ARG        specify source image embedding [undistorted]
  --local-neighbors=ARG  amount of neighbors for local view selection [4]
  --keep-dz              store dz map into view
  --keep-conf            store confidence map into view
  -p, --writeply         use this option to write the ply file
  --plydest=ARG          path suffix appended to scene dir to write ply files
  --bounding-box=ARG     Six comma separated values used as AABB [disabled]
  --progress=ARG         progress output style: 'silent', 'simple' or 'fancy'
  --force                Reconstruct and overwrite existing depthmaps
```

opendm/context.py
```diff
@@ -23,15 +23,15 @@
 # define orb_slam2 path
 orb_slam2_path = os.path.join(superbuild_path, "src/orb_slam2")
 
-# define smvs join_paths
+# define mve join_paths
 makescene_path = os.path.join(superbuild_path, 'src', 'elibs', 'mve', 'apps', 'makescene', 'makescene') #TODO: don't install in source
-smvs_path = os.path.join(superbuild_path, 'src', 'elibs', 'smvs', 'app', 'smvsrecon')
+mve_path = os.path.join(superbuild_path, 'src', 'elibs', 'mve', 'apps', 'dmrecon', 'dmrecon')
+mve_path_pc = os.path.join(superbuild_path, 'src', 'elibs', 'mve', 'apps', 'scene2pset', 'scene2pset -F2')
```
pierotofy: `-F2` should not be hard coded as it references the chosen scale value (depends on `--max-pixels`). If a dataset has really large input images (or really small images), this will fail. I think the value can be looked up by looking at the contents of the `.mve` folders.

pierotofy: Another alternative is to calculate a scale value ourselves instead of letting MVE sort it out and pass it to both scene2pset (via -F) and to dmrecon (via -s).

pierotofy:
Reducing the scale of reconstruction reduces the noise a bit, but perhaps this is just the result of having fewer points overall?

image

opendm/context.py
pierotofy: Maybe this is why you noticed a few failures during your tests.

pierotofy:
As far as `scene2pset`, I think we'll want to leverage the scale and confidence values to do the filtering. In particular, I think discarding points below a certain confidence value is a must. I'm getting some good results by discarding values below 0.25 confidence. Perhaps there's a better value, or maybe it should be user-defined.

animated

pierotofy:
scene2pset parameters:

```
Usage: scene2pset [ OPTS ] SCENE_DIR MESH_OUT
Available options: 
  -d, --depthmap=ARG       Name of depth map to use [depth-L0]
  -i, --image=ARG          Name of color image to use [undistorted]
  -n, --with-normals       Write points with normals (PLY only)
  -s, --with-scale         Write points with scale values (PLY only)
  -c, --with-conf          Write points with confidence (PLY only)
  -m, --mask=ARG           Name of mask/silhouette image to clip 3D points []
  -v, --views=ARG          View IDs to use for reconstruction [all]
  -b, --bounding-box=ARG   Six comma separated values used as AABB.
  -f, --min-fraction=ARG   Minimum fraction of valid depth values [0.0]
  -p, --poisson-normals    Scale normals according to confidence
  -S, --scale-factor=ARG   Factor for computing scale values [2.5]
  -C, --correspondence     Output correspondences (in absence of -m and -b only)
  -F, --fssr=ARG           FSSR output, sets -nsc and -di with scale ARG
```

To export confidence values, pass `-c`.

pierotofy:
Overall I think things that could be added/improved:
 - Pass scale args to both dmrecon and scene2pset.
 - Filter based on confidence level. This can be done via PDAL.
 - Fix progress flag.

Finding how to respect `--max-concurrency` is also a priority, but the PR could be merged without it.

smathermather:
Cool. I’ll dive in. This is an awesome review.

scripts/odm_meshing.py
```diff
@@ -49,7 +49,7 @@ def process(self, inputs, outputs):
                      (args.rerun_from is not None and
                       'odm_meshing' in args.rerun_from)
 
-        infile = tree.smvs_model
+        infile = tree.mve_model
```
dakotabenjamin: For smvs as an option: if arg.usemve: infile = mve_model; else smvs model

scripts/odm_georeferencing.py
```diff
@@ -99,7 +99,7 @@ def process(self, inputs, outputs):
                 elif args.use_opensfm_dense:
                     kwargs['pc'] = tree.opensfm_model
                 else:
-                    kwargs['pc'] = tree.smvs_model
+                    kwargs['pc'] = tree.mve_model
```
dakotabenjamin: if arg.usemve: kwargs = mve_model; else smvs model

scripts/odm_app.py
```diff
@@ -164,14 +158,14 @@ def slam_connections(self, p):
         connections += [self.tree[:] >> self.slam['tree'],
                         self.args[:] >> self.slam['args']]
 
-        connections += [self.tree[:] >> self.smvs['tree'],
-                        self.args[:] >> self.smvs['args'],
-                        self.slam['reconstruction'] >> self.smvs['reconstruction']]
```
dakotabenjamin, line 166: if args.usemve:
connections += <mve connection>;
else:
connections += <smvs connection>;

scripts/odm_app.py
```diff
@@ -49,13 +49,7 @@ def declare_cells(p):
                                            fixed_camera_params=p.args.use_fixed_camera_params,
                                            hybrid_bundle_adjustment=p.args.use_hybrid_bundle_adjustment),
                  'slam': ODMSlamCell(),
-                 'smvs': ODMSmvsCell(alpha=p.args.smvs_alpha,
-                                     max_pixels=p.args.depthmap_resolution*p.args.depthmap_resolution,
-                                     threads=p.args.max_concurrency,
-                                     output_scale=p.args.smvs_output_scale,
-                                     shading=p.args.smvs_enable_shading,
-                                     gamma_srgb=p.args.smvs_gamma_srgb,
-                                     verbose=p.args.verbose),
```
dakotabenjamin, line 56: add smvs back

scripts/mve.py
dakotabenjamin: If the user selects mve as the processing option, and also specifies --max-concurrency, then a warning ("Thread limiting is not available for this option and will run on all threads, are you sure you want to continue? [Y/n]

dakotabenjamin (dismissed):
Bringing back smvs as the default is my recommendation; see comments.

scripts/mve.py
pierotofy: If we can't limit the threads a warning is fine, but I'd caution with the Y/n prompt, could cause nodeodm to get stuck indefinitely.

pierotofy:
You should be able to limit the number of threads by setting the OMP_NUM_THREADS variable before calling `dmrecon`.

I tried this with the multithread version of `makescene` and seemed to work.

`OMP_NUM_THREADS=2 ./makescene <input> <output>`

https://github.com/simonfuhrmann/mve/pull/460

smathermather:
Ha! Awesome. I tried that but did it wrong. Nice to see how it’s done.

smathermather:
Todo for India:
* ~~Fix progress flag~~
  * ~~Add ```--progress=silent``` flag to ```dmrecon```~~
* ~~Fix broken georeferencing~~
* Fix --max-concurrency
	* ```"-t%s" % self.params.threads,```
	* ```OMP_NUM_THREADS=2 ./makescene <input> <output>```
	* ```OMP_NUM_THREADS=2 ./dmrecon```
	* ```OMP_NUM_THREADS=2 ./scene2pset```
	* Pull in https://github.com/simonfuhrmann/mve/pull/460
* ~~passing of parameters~~
  * ~~Pass scale args to both dmrecon and scene2pset.~~
* Filter based on confidence level.
  * ~~Add flag for confidence level to scene2pset~~
  * Filter point cloud in PDAL with confidence level
* bring back smvs
* If feasible, make mve more resumable
* We are going to run into memory issues with PDAL with larger datasets due to the larger point cloud size.  Whether in this pull request or later, we will need to address an approach to work around this, likely splitting pdal steps into subsets, or perhaps as ```--large``` comes online, it becomes moot as an issue.

smathermather:
m(o)ve over smvs

> You should be able to limit the number of threads by setting the OMP_NUM_THREADS variable before calling `dmrecon`.
> 
> I tried this with the multithread version of `makescene` and seemed to work.
> 
> `OMP_NUM_THREADS=2 ./makescene <input> <output>`
> 
> [simonfuhrmann/mve#460](https://github.com/simonfuhrmann/mve/pull/460)

@pierotofy Setting this before calling ```dmrecon``` doesn't work, it just stops being threaded. For example, modified to use 8 cores on a 16 core machine and then it only uses one core.

pierotofy:
I'm reading more about OpenMP, `OMP_NUM_THREADS` apparently sets an upper bound on the number of threads, but doesn't guarantee that those many threads will be used. Perhaps the scheduler decided that it was more efficient to use a single thread instead of 8? 

https://stackoverflow.com/questions/42970700/openmp-dynamic-vs-guided-scheduling/43047074

smathermather:
Ah, that would explain it. Well, short of rewriting OpenMP, then the changes we made today on that should work!

smathermather:
@pierotofy -- we should probably pull in your patch to makescene for supporting threading too, yes?

smathermather:
> I'm reading more about OpenMP, `OMP_NUM_THREADS` apparently sets an upper bound on the number of threads, but doesn't guarantee that those many threads will be used. Perhaps the scheduler decided that it was more efficient to use a single thread instead of 8?
> 
> https://stackoverflow.com/questions/42970700/openmp-dynamic-vs-guided-scheduling/43047074

Hmm, it's still not working as expected. It's running on all the cores instead of half of them as specified (we're just hardcoding for now for testing purposes):
```mve_path = os.path.join('OMP_NUM_THREADS=8', superbuild_path, 'src', 'elibs', 'mve', 'apps', 'dmrecon', 'dmrecon --progress=silent')```

**Comment by pierotofy** [comment]
`makescene` has been already patched upstream, we should just add the `OMP_NUM_THREADS` variable per #930

**Comment by pierotofy** [comment]
Ah :) no I don't think you can pass variables via os.path.join.

You have to pass them via `env`.

```
subprocess.call(cmd, shell=True, env=env)
```

See https://github.com/OpenDroneMap/ODM/blob/master/opendm/system.py#L20 for an example.

**Comment by smathermather** [comment]
> Ah :) no I don't think you can pass variables via os.path.join.
> 
> You have to pass them via `env`.
> 
> ```
> subprocess.call(cmd, shell=True, env=env)
> ```
> See https://github.com/OpenDroneMap/ODM/blob/master/opendm/system.py#L20 for an example.

Lol. I am still a python novice. Will do.

**Comment by smathermather** [comment]
We are going to run into memory issues with PDAL with larger datasets due to the larger point cloud size.  Whether in this pull request or later, we will need to address an approach to work around this, likely splitting pdal steps into subsets, or perhaps as ```--large``` comes online, it becomes moot as an issue.

Just ran on a dataset of 750 images with 64GB RAM, and even running with max-concurrency of 2, I run out of memory.

**Comment by pierotofy** [comment]
Are you sure it's a memory issue? How many points are in the input mve point cloud?

**Comment by smathermather** [comment]
Oh, I’m pretty sure it’s memory. I upped the RAM and ran with —max-concurrency 1 and it completed without issue.

I can report back on point cloud size when I get back onto a network (early next week).

File: scripts/mve.py [hunk 7]
```diff
@@ -64,17 +64,30 @@ def process(self, inputs, outputs):
...
                 system.run('%s %s %s' % (context.makescene_path, tree.mve_path, tree.mve))
 
-		print(args)
+
 
             # config
             config = [
-                "-s%s" % args.mve_output_scale
+                "-s%s" % args.mve_output_scale,
+	     	"-s2",
```
pierotofy: `-s` is duplicated here.

smathermather (commented):
It's looking really quite good. Cut and print:
image

smathermather:
With HS it still looks good which is a robust test:
image

smathermather:
Filtering with SMRF will require some additional parameterization, but DTMs are looking hopeful here too:
image

smathermather:
We do have some issues with orthos of tall buildings that are cropping back up:
image

This is with a pretty small subset of 16, so I'm suspicious whether changing the mesh size defaults will help at all here.

smathermather:
> We do have some issues with orthos of tall buildings that are cropping back up:
> ...
> 
> This is with a pretty small subset of 16, so I'm suspicious whether changing the mesh size defaults will help at all here.

On further inspection, this isn't a new issue. Here's how it looks from current master:
image

pierotofy:
Reminder to self that we still need to enforce args.max_concurrency in the various calls.

pierotofy:
DEMs improved!

Untitled

pierotofy:
Getting some bad distortion on certain datasets however.

Untitled

pierotofy:
Probably due to noise: 
image

pierotofy:
0.6 as a default confidence threshold yields much better results:

image

pierotofy:
Fixes #970

pierotofy:
`max-concurrency` is now respected on `makescene`, `dmrecon`, `scene2pset`.

@smathermather @indiajohnson-cm this can be tested and pending further issues I think it can be merged into master as-is.

pierotofy:
Investigating a seg fault for one test dataset (sheffield_cross) that processed fine with SMVS.

image

pierotofy (approved):
Seg fault did not happen again after more testing. This might be tied to some sort of racing condition, but for the time being I think it's OK to move forward.

@smathermather OK to increase version number and merge in your opinion?

pierotofy:
Implemented #972

smathermather:
> Seg fault did not happen again after more testing. This might be tied to some sort of racing condition, but for the time being I think it's OK to move forward.
> 
> @smathermather OK to increase version number and merge in your opinion?

Yes to both! Tests are looking good on my end.

smathermather:
> 0.6 as a default confidence threshold yields much better results:

This matches what I was seeing too -- confidence thresholds greater than 0.5 needed for best results.

smathermather:
Hmm, I dismissed @dakotabenjamin's review based on a myth I think... . @pierotofy -- smvs is still gone from this pull request, correct?

pierotofy:
Yes SMVS is gone.

smathermather:
A sibling docs update that I meant to do as a pull request but did on the active repo (yipe!):
https://github.com/OpenDroneMap/docs/issues/13

Anyway, docs should reflect these changes and more. Let me know if I missed anything.

PR #925, opened 2018-12-12
Smarter concurrency for DEM generation

pierotofy:
Often times the pipeline fails because the DEM module attempts to use all cores available to run PDAL in parallel, and subsequently runs out of memory. We currently suggest users to decrease the `--max-concurrency` parameter, but this slows down other parts of the pipeline. These changes cap the max concurrency parameter based on a memory consumption estimate of the parallel process.

PDAL currently requires ~2x the input file size of memory per available core to process a DEM layer, so `get_max_concurrency_for_dem` computes a value that is capped based on both `max-concurrency` and file size estimate.

Worst case the DEM process runs on a single core. Best case it uses all of them. Sometimes it will select a value in between, based on memory availability.

### DIFF
--- a/opendm/concurrency.py
+++ b/opendm/concurrency.py
@@ -1,0 +1,26 @@
+from psutil import virtual_memory
+import os
+
+def get_max_memory(minimum = 5, use_at_most = 0.5):
+    """
+    :param minimum minimum value to return (return value will never be lower than this)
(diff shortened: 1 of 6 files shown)


pierotofy:
It also refactors the `get_max_memory` function in a single module to keep all memory/concurrency logic in a single place.

smathermather (approved):
Looks great, and I like this much better than decimation.

PR #927, opened 2018-12-23
Updated exiv2 dependency to v27

garlac:


pierotofy:
Looks good, thanks @garlac! :christmas_tree:

Issue #928, opened 2018-12-23
odm_georef does not handle points clouds > 600 million points


pierotofy:
While it doesn't crash, the point cloud it generates is corrupted. This is likely due to some overflow in the PCL PLY parser.

Artifacts:

![image](https://user-images.githubusercontent.com/1951843/50386254-55cd3280-06b1-11e9-91c3-89d5f66a11fb.png)

How it should look instead:

![image](https://user-images.githubusercontent.com/1951843/50386261-65e51200-06b1-11e9-9268-f75f2bdd036e.png)

I will rewrite the PCL loading code and add the ability to do stream read/writes, so that the point cloud does not need to be loaded in memory all at once.

pierotofy:
Or perhaps using the PDAL C++ API.

By using PDAL we could remove the unnecessary step of generating a PLY file and we could go directly to LAZ by passing the SRS as a param to odm_georef.

smathermather:
Could speed that step up a bit too.

garlac:
I came across same issue...Below are the screenshots. 1201 images and point cloud had 217.77 million points and size 9.2 Gb.

Georeferenced point cloud Vs normal dense point cloud:
![gcp_hosur](https://user-images.githubusercontent.com/40228026/50883340-c6908080-140d-11e9-9d8e-52d4315afbbe.png)

It also created some other point cloud below like this:
![gcp_hosur_1](https://user-images.githubusercontent.com/40228026/50883366-d6a86000-140d-11e9-8ba4-ffb9158b7d72.png)

Hope the pull request in - https://github.com/OpenDroneMap/ODM/pull/929/files
solves this problem. Will test it.

Thank you.

garlac:
I have tested it and found that the geo-referenced point cloud came good. Thank you @pierotofy .

pierotofy:
Thanks for testing @garlac!

Out of curiosity, what environment did you test this on? (Windows, Linux, Mac, and was it with docker or native?)

@dakotabenjamin do you need more time to test #929?

garlac:
It was Linux env and native install. 64 Gb RAM 12 cores.

PR #929, opened 2018-12-26
Modified odm_georef to output .LAZ files directly

pierotofy:
This is related to #928. Adds the ability to output files directly in .LAZ format instead of creating an intermediate PLY file.

Currently ready for testing, but don't merge until we have run a few datasets and confirmed that things work as expected (and that large point clouds can be processed correctly).

### DIFF
--- a/modules/odm_georef/CMakeLists.txt
+++ b/modules/odm_georef/CMakeLists.txt
@@ -28,7 +28,6 @@ include_directories(${EIGEN_ROOT})
 
 # PDAL and jsoncpp
 find_package(PDAL REQUIRED CONFIG)
-message(${PDAL_INCLUDE_DIRS})
 include_directories(${PDAL_INCLUDE_DIRS})
 include_directories("${PROJECT_SOURCE_DIR}/../../SuperBuild/src/pdal/vendor/jsoncpp/dist")
 link_directories(${PDAL_LIBRARY_DIRS})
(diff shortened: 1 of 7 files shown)


pierotofy:
Large point clouds (tested with  > 600 millions points) work as expected.

pierotofy:
@dakotabenjamin do you have time to review this one?

dakotabenjamin:
Yeah, I'll have a chance this weekend once I get back from vacation.

dakotabenjamin:
I don't think I have a 600m point dataset but I'm running over a few datasets tonight and looking at the code

dakotabenjamin:
TBH I don't quite understand the code enough to give meaningful critique, however i've run it without problems. I'm glad you made the xyz process pdal-based. I agree that we should see how much we can do solely in pdal. 

@smathermather will this help with gfdrr project?

dakotabenjamin (approved):
Ran it on 6m points no problem. The code looks good, nice improvements!

pierotofy:
Thanks for reviewing @dakotabenjamin!

smathermather:
> TBH I don't quite understand the code enough to give meaningful critique, however i've run it without problems. I'm glad you made the xyz process pdal-based. I agree that we should see how much we can do solely in pdal.
> 
> @smathermather will this help with gfdrr project?

I think it should help some, yes.

PR #941, opened 2019-01-29
Fixed proj string incompatibility with GDAL

pierotofy:
This fixes a problem that causes GDAL to incorrectly assign every orthophoto a southern UTM projection.

Lesson learned: can't trust proj strings boolean values.


pierotofy:
Fixes https://github.com/OpenDroneMap/WebODM/issues/611

PR #947, opened 2019-02-15
Fixed DEMs orthophoto misalignment bug

### LINKED ISSUE #560: Output ODM orthophoto is shifted compared to point cloud and dsm when using GCP
### How did you install WebODM? (Docker, natively, ...)?
Docker on Ubuntu 18.04 + git.

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)
Firefox 63 on Ubuntu Linux.

### What is the problem?
The ouput orthophoto from WebODM is constantly shifted compared to the point cloud and dsm (both PC and dsm are perfectly aligned) when using a gcp_file.txt. This is not the case when no GCP are used. The shift varies depending on the parameters, between 1 and 2m. 

### What should be the expected behavior?
The orthophoto should be aligned with the other assets (.ply PC and dsm.tif are the ones are consider), if it is projected from the PC (if I understand things correctly)

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)
The dataset is the same as in: https://github.com/OpenDroneMap/WebODM/issues/549#issuecomment-437840704, you can download it from:
https://drive.google.com/folderview?id=1B1c5BsG19Hx7RYwzUgZ26nZWb702yLbz

I reconstructed this dataset with and without a GCP file. The coordinates in the GCP file are ETRS89/UTM31N (see discussion in the other issue thread) but the same offset shows up if the coordinates are WGS84/UTM31N (or so I think). Based on my understanding (might be crap), the orthophoto and all the other assets are derived from the mesh, after it is georeferenced with the EXIF data or the GCP file. OK, screenshots where I use QGIS and both the measuring tool and the terrain profile tool to demonstrate. 

### if I don't use a GCP file, 
- the dsm and orthophoto are perfectly aligned. We'll take the change of slope as a reference: we can see that the change of slope as visible in the photo is where the abrupt change of altitudes occur in the dsm values at x=28.151. 
capture d ecran de 2018-11-23 11-10-31

- I also opened the .ply georeferenced point cloud (PC) in CloudCompare et projected a dsm raster along the Z axis. The resulting values are visible in the profile graph above as a red curve, the blue curve being the ODM dsm. They are basically identical (only the resolution is different). Finally I also projected a RGB raster along the Z axis, which is basically the orthophoto (right?), and it is perfectly aligned with ODM orthophoto (I don't show a screenshot). 

### if I use a GCP file
-  I did the same tricks (CloudCompare, projection along Z of a dsm raster and RGB rasters + comparison in QGIS). 
 * first the offset between the odm ortho and CloudCompare projection, about 2m, I observed this number in different tries. 
capture d ecran de 2018-11-23 10-26-26
 * then in the profile plot the ODM dsm and CloudCompare projected dsm are indistinguishable, 
capture d ecran de 2018-11-23 10-24-41
 * ... and they are aligned with the CloudCompare orthophoto, but **not** aligned with the ODM dsm as you can see here (the cursor is on the change of slope):
capture d ecran de 2018-11-23 10-25-19
 * in case you wonder which one is correct, it is definitely the PC (and dsm). 

I'm still skeptical about the SCR thingie (i.e. WGS84UTM31N) and if it is at the origin of the problem (in case it has an impact on how the orthophoto is projected from the PC). But in that case, why is the dsm not projected with the same error/offset? Tell me if I shall try something else. Thanks.
Denis

pierotofy:
This PR should fix https://github.com/OpenDroneMap/WebODM/issues/560 and https://github.com/OpenDroneMap/WebODM/issues/549. It's a problem that has come up on the forum also at https://community.opendronemap.org/t/webodm-vs-odm-differences-in-the-precission-and-quality-of-orthophotos-and-dems/1735/5

It seems to have been caused by two separate problems. 
 - One was the decision to not include the GCP translation offset as part of the georeferenced model transform (instead to aggregate the translation value to the UTM offset). While the choice of doing so seemed to be equivalent, it turns out that when it comes to render the model to a PNG (with `odm_orthophoto`), it does make a difference for certain models (models that are at a 45 degree angle to north seem to be most affected, while "square" models are virtually unaffected, and that's why this was difficult to track down). I didn't pin-point the exact reason, but I believe it has to do with the calculation of bounds and/or the fact that the model is flipped before rendering.
 - Two was a bit more subtle, when we called `odm_georef` we used the 3D mesh as the model to calculate the transform matrix, but then used the 2.5D mesh to render the orthophoto. The two models are almost identical, but not exact, so after fixing the first problem a very small offset between the orthophoto and DEMs still occurred (since they were being georeferenced using two different, although almost identical transforms).

I'm re-running tests on various datasets to check for problems, but so far so good.

Now (good GCP, DSM and ortho alignment):

image

Before (odm_data_copr was particularly bad because it's at 45 degrees to north):

image

After:

image


pierotofy:
Ran tests across a wide range of datasets with good results. Unless there are objections, I will merge this by end of the day tomorrow.

smathermather:
I won’t be able to test until later in the week, but the benefits outweigh the risks, so go for it.

denistestemale:
Hello Piero.
I'll try ASAP on different datasets which suffered from
- dsm - orthophoto misalignements
- lack of precision of GCP in the orthophoto (and point clouds). It is still not clear to me if this problem is solved by your modification. 

Many thanks. Denis

smathermather:
Lack of precision for the point cloud should have been fixed in a recent pull request changing the process for outputting the point cloud.

denistestemale:
Hello.
I still observe a lack of precision, some examples (GCP were measured at the center of the black and white mark):
capture d ecran de 2019-02-21 22-27-03
capture d ecran de 2019-02-21 22-27-17
capture d ecran de 2019-02-21 22-27-56

This is taken from the dataset made available https://github.com/OpenDroneMap/WebODM/issues/560 that Piero already tried. @pierotofy  do you observe the same shifts in your results?

I cannot check the misalignement since I have no point clouds to check against (I cannot open the laz files due to liblas configuration problems that are beyond my computing skills, and the csv point clouds have no color to find the GCP marks in them). 

Cheers

pierotofy:
Some points will be off, but 3 of them should be pretty spot on. I think the reason you are observing the discrepancy in a few locations is because the georeferencing uses an affine transform to align the orthophoto, using the best 3 points it can find. If the camera has too much lens distortion, as an example, the resulting 3D model, and thus the orthophoto will not have perfect distances so the software just picks 3 points and tries to fit the 3D model with the "best" combination of 3 points that yield the lowest loss of accuracy. During this transformation straight lines remain straight lines, so if you have warping (again, which could be caused by camera distortion), there's no way to reconcile that using the current approach (beside removing camera distortion).

This is an area of future improvement, as other software probably uses higher degree polynomials (ex. see https://grass.osgeo.org/grass70/manuals/i.rectify.html and https://en.wikipedia.org/wiki/Thin_plate_spline) to achieve better georeferencing across all points.

denistestemale:
This is a very interesting point that you raise. Let me know if this is not the best place to discuss it, I can open a thread in community.opendronemap.org. 
Indeed I had noted the "best triplet" message in the georeferencing log, and thus I concluded that the georeferencing transformation is affine. Actually, I was surprised that there are very few (if not none) tools to register point clouds (PC) beyond the affine methods in 3D, compared to methods in 2D (polynomial, thin plate spline, etc.). I was told that https://simpleelastix.github.io/ is one candidate to do that, but I'm not expert enough to explore it further. Anyway, based on that and the problems I had with shifts of the orthophoto, I settled with the following method which doesn't use the GCP option of WebODM anymore:
- WebODM without GCP
- thin plate spline registration of the resulting orthophoto in QGIS  --> I obtain an **orthophoto** which is good. 
- georeferencing of the WebODM PC in CloudCompare using 3 GCP (so 3 GCP are spot on, the other ones are shifted due to distorsion)
- projection of the CloudCompare georeferenced PC along the z axis to obtain the DSM and orthophoto
- thin plate spline registration of the resulting orthophoto in QGIS in order to use the obtained TPS parameters to register the DSM --> I obtain a **DSM** which is good .
- combination of the good orthophoto and DSM in CloudCompare to generate a **colored PC** which is good. 

The only drawback of the method is that I cannot avoid the errors in the DSM (distorsion of the images, dome effect). In the case of the dataset above, I observed a dome shape of ~0.7m amplitude along the 700m of the mapped area. I did many tries with different WebODM options (fixed camera parameters, with or without camera calibration, etc.): without camera calibration I observed an inversed dome effect (bowl?), but all in all never a correct result. In comparison, Pix4D results are spot on (orthophoto, DSM and point cloud are spot on on the GCP, no dome effect, etc. down to 2-3 cm accuracy). Please understand that this last remark is not meant to be nasty, I only used the trial version of Pix4D to try to understand the limits of my results.

So now in my case I reckon it narrows down to only one problem: proper correction of the camera distorsion which is the cause of the lack of accuracy in the orthophoto (positions of the GCP) and DSM (dome effect).

smathermather:
What I would actually like to see with GCPs is even deeper: that the GCPs would affect the model development even to the extent of constraining the camera calibration portion, considering a model which solves for camera calibration, SfM, constrained by GCPs.

denistestemale:
Or a way to register finely the point cloud obtained from SfM, in 3D, based on the GCPs? Like Thin Plate Spline in 3D, if that makes sense.

smathermather:
TPS makes sense for sure, but there's potentially some opportunity to reduce the errors in the areas we are not measuring (areas without GCPs) by integrating the constraints into the SfM model itself, especially as this applies to camera calibration and similar non-linear and non-random errors.

smathermather:
Though TPS and other approaches would be faster and simpler to implement and usually good enough, so I wouldn't complain about a pull request that included them... . :)

denistestemale:
So you mean that TPS for xyz point clouds is do-able? Not do-able in the sense "easy", but do-able in the sense "TPS is not limited to 2D". 
As I said above I was surprised to find out that non affine registration is not available in existing PC software. CloudCompare as an example, which proved that it is quite flexible to implement algorithms from the literature, is limited to affine registration if I'm correct.

smathermather:
Ha! Well... . Absent research if what is trivial and not... .

Related: the Hobu crew (PDAL and other tools) are looking at some dedicated point cloud matching approaches in their toolchains, which might harbor some interesting opportunities for ODM when the time comes.

argroos:
We were discussing the misalignment issue and lack of precision in the ODM forum ([here](https://community.opendronemap.org/t/webodm-vs-odm-differences-in-the-precission-and-quality-of-orthophotos-and-dems/1735)). I recognized some further issues and I would like to discuss them in this thread since they are related to denistestemale's comments above (please let me know if you rather prefer a new thread).
1) Using ODM 0.4.1 (2019-02-18), the misalignment issue is fixed. Three GCPs (all measured with differential GPS in the field) are now spot-on, but the offset at the other GCPs still equals to 1-2 m, as also visible in  denistestemale's examples above. Is it default in ODM that only the "optimal gcp-triplet" is chosen for georeferencing or is it also possible to force ODM to consider all available GCPs? Just for comparison: in Pix4D all GCPs are spot-on.
2) For one survey (29.09.2018), the ODM DSM is extremely inclined (towards west). It seems as two of the three chosen GCPs (nr. 7 and 12) are identical (the log file is attached below):
"Optimal gcp triplet chosen: 7, 12, 33
Position in model for **gcp 7: x=-236.564 y=-66.996 z=2449.920**
Position in model for **gcp 12: x=-236.560 y=-66.025 z=2450.144**
Position in model for gcp 33: x=416.630 y=341.667 z=2497.934"
That produced an erroneous DSM with an inclination towards west (see difference of ODM DSM and Pix4D DSM from the same survey; identical location of GCP 7 and 12 is marked in yellow, the other GCPs in black, GCP 33 is outside the field of view). The error in the west is ca. -5m, in the east ca. +5m. Do you have an explanation for this odd choice of GCPs?
dsm_diff_odm_minus_pix4d
3) For our other survey (03.06.2018), the DSM is not inclined, but the vertical mean error is in the order of ca. 1.5 m (using Pix4D, the vertical DSM error decreases to 10-30 cm). We intend to use consecutive DSMs of the glacier surface at our study site to quantify the ice volume loss over the summer. A high precision is therefore required. Can you recommend an alternative georeferencing/postprocessing workflow or do you have any other idea how we could improve the precision of the ODM DSMs and orthophotos?
Final note: we use Pix4D only for testing and comparing...
[odm_georeferencing_log_2018-09-29.txt](https://github.com/OpenDroneMap/ODM/files/2931349/odm_georeferencing_log_2018-09-29.txt)
Many thanks for your support!

denistestemale:
Hello.
Very old thread but I had a long hiatus in the use of drones. 
- It's probably not very useful to comment so long after, but it is just to mention that I updated WebODM and ran again the same datasets I was discussing here: the results are now very good. The precision for all GCP is very good (below 10cm) and DSM and othophoto are aligned. 
- Regarding the discussion we were having with @smathermather about the 3D registration of PC and/or the use of GCP to tune the SFM model: did you implement something in the last 12 months? Indeed I'm very satisfied by the precision of **all GCPs**, and not only a triplet, as if it was not limited by the affine registration of the PC. Another hint is that the doming effect disappeared in the same datasets I was observing it.

Anyway: many thanks! 
Cheers
denis

smathermather:
Hi Denis,

Well, we are only human over here, and a detailed complement is very welcome even after long hiatus. :)

There are two elements that are improved here in OpenSfM, and I will try to give credit where it is due: @YanNoun and others have brought OpenSfM a lot of great features, including improved GCP handling. In addition to this, @bdholt1 and @paulinus and others have added the Brown-Conrady camera model to OpenSfM. Apologies if I missed anyone in those credits.

Finally, recognizing the potential for improving things for our end users, @pierotofy made sure that brown is the default for camera models tried in auto mode in OpenDroneMap. Also, I want to mention Patrice Conrady pointed out the need for Brown-Conrady pretty earl on here: https://community.opendronemap.org/t/on-the-job-self-calibration-of-lens-distortion-with-a-brown-model/627

YanNoun:
A bit if heads-up regarding latest (4 months) features in OpenSfM :
- We added a new camera projection (dual) which can seamlessly transition between fisheye and perspective explicitely. Might not be needed anymore as most drones (>= Phantom 3) now exhibits fairly rectilinear behaviour. Would have been great for the Phantom 2 and its pseudo-rectilinear crops :-) But it's there so you can give a try. 
 - We are rewriting some parts in C++ (tracks manager, camera models), aiming at speeds-up. We'll still keep high-level code parts in Python. 

What's remaining IMHO :
 - Regarding GCP/GPS alignement, one big feature OpenSfM still miss wrt. Pix/Agi is the computation for the bias between the GPS coord. frame wrt. the GCP one (DJI GNSS are often globally shifted). Not something super complicated, just need to find some time for it. Would really improve RMS of the GCPs. 
 - Write covariances of cameras positions so it can be reported in some quality report. There's already things being written, we just need to output more of them.
 - We have internal stuff regarding handling of large scenes we'd like to merge back.

Thank you everyone for your work !

PR #953, opened 2019-02-22
Added --pc-las flag

### LINKED ISSUE #948: Add .LAS as output format
Some users seem to want a .LAS file instead of .LAZ (probably because the software they use to open files does not support laz). Would be interesting to add such an option, similar to `--pc-csv`.

pierotofy:
Fixes #948 

Fairly simple addition cc @smathermather


smathermather:
Looking good.

PR #956, opened 2019-03-06
DEM simplification, filtering for all

pierotofy:
This is a series of improvements related to DSM/DTMs.

Problem: DTMs are coming out pretty bad.

1) Removed the PMF filter which is slower than SMRF, needs more parameter tuning to yield decent results and I suspect most users don't care about what algorithm we use. DTM classification is a difficult task.  For real good results, automated methods are unlikely to yield perfect results so I would edge toward simplicity. If people have suggestions on flags to expose in regard to SMRF, let's hear suggestions. I haven't been able to find good defaults for various scenes (flat vs high slope) and settled on the defaults which seem to yield OK results in a large variety of datasets. `--pc-classify` is now a boolean flag.

2) Removed `--dem-terrain-type` which was originally added because part of `lidar2dems`. To quote their description of the feature: `It is important to note that the main advantage of specifying a region as flat versus complex is reduction in processing time, and not in the accuracy of the results`. I think it's just confusing people (I should get a better classification by specifying the type?) and SMRF doesn't seem to be affected as much by run-time.

3) Removed `--dem-initial-distance` and  `--dem-approximate` which are used by just PMF

4) Renamed `--dem-maxsd` to `--pc-filter`. Why just let DEMs enjoy the results of a clean point cloud? We now filter the dense point cloud as soon as dense reconstruction is over, so all products (mesh, texturing and DEMs) can have a cleaner point cloud as input.

For 4. I've introduced a new program that uses PDAL to create a "normalized" PLY output which is readable by poisson recon and others (standalone PDAL would not cut it).

This will improve outputs for DTMs, meshes and fix problems such as https://community.opendronemap.org/t/point-cloud-has-many-points-at-extreme-depths/1816/3

Things to test: 
 - That nothing broke
 - DTM outputs
 - Point cloud outputs

Hoping to merge these by end of the week unless there are objections / things to fix. :clinking_glasses:


pierotofy:
Ran tests across a wide variety of datasets using oats with no apparent problems. Will merge for now, if we find issues later on we'll fix them in a separate PR.

PR #957, opened 2019-03-07
Cleanup legacy code, much faster dataset loading

pierotofy:
Removes legacy code that is currently not being used, removes `--force-focal` and `--force-ccd` which currently do nothing.

Replaces the logic for reading the image size from the slower cv2.imread (but is still used as a fallback) to get_image_size (a small, self-contained library that reads only the first few bytes of the image instead of the full content), which results in a hundred fold decrease in dataset loading time.

@dakotabenjamin could you take a look over these?


pierotofy:
Also removed `--orthophoto-target-srs` which does nothing. I think it would be a cool feature to have in the future, but the flag should not be there if it's not implemented.

pierotofy:
And `--zip-results`, which also does nothing.

pierotofy:
Ran tests, things seem OK. Will merge this by tomorrow unless there are objections. Requested changes can also be added as part of a new PR.

smathermather:
This seems helpful and harmless. I don’t have time to test ATM, and agree if your testing is working ok, merge away...

PR #959, opened 2019-03-12
More Cleanup

pierotofy:
Removed unused parameter `start_with` which does nothing.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -118,10 +118,10 @@ def config():
                              'matcher-neighbors and this to 0 to skip '
                              'pre-matching. Default: %(default)s')
 
-    parser.add_argument('--use-fixed-camera-params',
-                        action='store_true',
-                        default=False,
-                        help='Turn off camera parameter optimization during bundler')
+    # parser.add_argument('--use-fixed-camera-params',
+    #                     action='store_true',
+    #                     default=False,
+    #                     help='Turn off camera parameter optimization during bundler')
 
     parser.add_argument('--max-concurrency',
                         metavar='<positive integer>',
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -53,12 +53,6 @@ def config():
                         help='resizes images by the largest side for opensfm. '
                              'Set to -1 to disable. Default:  %(default)s')
 
-    parser.add_argument('--start-with', '-s',
-                        metavar='<string>',
-                        default='resize',
-                        choices=processopts,
-                        help=('Can be one of: ' + ' | '.join(processopts)))
-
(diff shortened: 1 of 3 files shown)


smathermather:
We're keeping the fixed camera parameter flag, yes?

pierotofy:
So I've discussed with Dakota about keeping it and it's our understanding that fixed camera parameters are helpful when the camera parameters are manually set (via overrides), but we currently don't have that functionality available in ODM. So I removed it, but I can re-add it if this flag is helpful for something that we didn't see.

Are you currently using it @smathermather ? Can you describe how it's useful?

smathermather:
If I understand it correctly (there’s some high probability that I don’t :)) it’s useful when the corrections have already been externally applied as well, in which case setting this and applying no camera parameters runs without corrections. Perhaps I’m thinking of another flag... .

pierotofy:
Yes I think you're spot on. OpenSfM allows the camera parameters to be overridden by placing an override file along with the configuration. This is useful if a person has run a good dataset (good overlap, at an angle, lots of features, etc.) and later decides to run a bad dataset with the same camera. In that case, using the camera parameters from the first dataset improves the results of the second. You don't want bundle adjustment to tweak the camera parameters in the second dataset, because they have already been optimized from the first (and would make things worse). Then you'd want to set `optimize_camera_parameters` to `no`. Other software does this and I think we should implement it with a good streamlined process in ODM.

By itself however, it just tells the software not to optimize (which is not.. useful?). At least that's my understanding from searching the OpenSfM repo https://github.com/mapillary/OpenSfM/issues/324

If I missed something we should certainly add the flag back however.

smathermather:
Another use case is as follows:

Camera calibration has otherwise been computed and the images undistorted based on those parameters. Now we have perfect images to feed into OpenSfM and we don't want OpenSfM to calculate parameters for them or modify them in any way. This would require we use ```--use-fixed-camera-params``` but not set any parameters. This is the functionality I don't want to remove from ODM.

smathermather:
In other words, we should reinstate ```--use-fixed-camera-params``` :)

smathermather:
> Another use case is as follows:
> 
> Camera calibration has otherwise been computed and the images undistorted based on those parameters. Now we have perfect images to feed into OpenSfM and we don't want OpenSfM to calculate parameters for them or modify them in any way. This would require we use `--use-fixed-camera-params` but not set any parameters. This is the functionality I don't want to remove from ODM.

Also: this isn't just a theoretical use case, but something a couple of colleagues need in their work with ODM.

pierotofy:
No problem, let’s add it back then.

PR #961, opened 2019-03-24
Use fixed parameters revert

smathermather:
Let me know if I missed anything in this revert. Justification here: https://github.com/OpenDroneMap/ODM/pull/959#event-2208780215


pierotofy:
The ecto cell definition need to also be reverted.

smathermather:
> The ecto cell definition need to also be reverted.

Fixed.

PR #962, opened 2019-03-25
[FIX BUG] build failed with PCL version 1.7.2 

zhudaoruyi:
PCL version 1.8.0 can build successfully


pierotofy:
Thanks @zhudaoruyi ! We tecnically are on 1.8.1 so this change was needed. I wonder why the build failed on your platform (it has been building fine until now on docker and many other machines).

PR #963, opened 2019-03-26
formats the Dockerfile for clarity

davidvasandani:
I've edited the Dockerfile for clarity (line length and alphabetization), merged 4 layers (lines 6-12) into a single layer, and removed a duplicate dependency (`software-properties-common` line 8 & 15).


pierotofy:
Looks great, thanks @davidneudorfer !

PR #966, opened 2019-03-30
Update pdal to 1.8.0

mikethefarmer:
Modify project to use current release PDAL 1.8.0.  

By default PDAL 1.8.0 will utilise stream mode if appropriate, which also requires additional 'bounds' parameters.  Adding the '--nostream' parameter forces it to standard mode so that it will act like PDAL 1.6.  

Further work is needed to implement stream mode (which can be looked at separately).  Stream mode 'might' help with very large files but also may slow processing down as getting the bounds from 'pdal info' does take some time.

This has been tested only with a 600 photo set (20Mpx), with no resize.  The '--nostream' was not required for an ortho-only (sparse) .ply, but was required for a full (dense) .ply.

PDAL 1.9 is soon to be released (currently RC) but requires GDAL 2.2+.  The repo version for Ubuntu 16.04 is only 2.1.3 so additional work would be needed to update further.

### DIFF
--- a/SuperBuild/cmake/External-PDAL.cmake
+++ b/SuperBuild/cmake/External-PDAL.cmake
@@ -8,7 +8,7 @@ ExternalProject_Add(${_proj_name}
   STAMP_DIR         ${_SB_BINARY_DIR}/stamp
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
-  URL               https://github.com/PDAL/PDAL/archive/1.6.zip
+  URL               https://github.com/PDAL/PDAL/archive/1.8.0.zip
   #--Update/Patch step----------
   UPDATE_COMMAND    ""
   #--Configure step-------------
--- a/opendm/dem/pdal.py
+++ b/opendm/dem/pdal.py
@@ -155,6 +155,7 @@ def run_pipeline(json, verbose=False):
     cmd = [
         'pdal',
         'pipeline',
+        '--nostream',
         '-i %s' % jsonfile
     ]
     if verbose:


pierotofy:
Thanks @mikethefarmer ! We'll need to run a bit of tests as PDAL is also a dependency for `odm_georef` and `odm_filterpoints`. Time ago I also noticed that PDAL 1.8 was using more memory for generating DEMs (but I can't recall if that was using stream mode or not) compared to 1.6, but perhaps that has changed.

Would also be prudent to add --nostream to other PDAL calls throughout the pipeline, but perhaps it's better with streaming.

The reason for adding `--nostream` is because of memory requirements or some other issue?

mikethefarmer:
Piero,  as per my comment (mikegf) on forum (https://community.opendronemap.org/t/error-in-odm-meshing-cell-pdal-pipeline/1413) a sparse version ran OK but I got an error running the dense without the '--nostream' (PDAL: writers.gdal: Can’t shift existing grid outside of new grid during expansion.).  I assumed (without any diving deep into PDAL code) that the number of points in the sparse version fell below the threshold at which PDAL 1.8 used streaming, and the error on the dense one was due to the fact that it defaulted to streaming - but it then lacked the required 'bounds' parameters for streaming.

I am on training wheels with ODM (and in fact python) but if there are any other calls to PDAL translate or pipeline then they will need to be looked at for compatibility with streaming.  Happy to take some pointers to the other PDAL-calling code and I can have a look at them as well.  They weren't immediately obvious from my trawling through the code trees.

As far as I can tell, in terms of the writers only gdal and laz use streaming where possible, but not sure of other calls (I didn't think that translate did hence no update to that bit of code in pdal.py).  My reading of the PDAL release notes was that v1.7 changed the default from no-stream (with '--stream' as an option) to stream if possible and added '--nostream' to override.  In order to minimise the potential impact of the change to 1.8 I thought the use of '--nostream' (which effectively maintained the v1.6 status quo) was prudent.

If you think that write-laz streaming mode (which does not appear to require additional parameters) would be preferable then we would need to change the build of 'cmd' in pdal.py to include'--nostream' for write-gdal only.  

I am currently waiting to see whether pdal 1.8 solves the segfault (and observe peak memory usage) with my large data set (sfm takes about 60hrs so it is slow process).  In parallel I am looking at what needs to be done to find the 'bounds' parameters for streaming gdal.write.  I have some code snippets (using subprocess.pcode() to run 'pdal info') that worked on my small 600 photo dense .ply file, but even then 'pdal info' on that file took about 15min.  I didn't time the '--nostream' pipeline for a comparison but it may well take longer to run some files in stream with the 'pdal info' overhead.  Of course stream 'should' have the advantage of lower memory usage.  

If my large set (ortho-only) is successful with the sparse .ply without streaming then I may try again with the dense and see how the processing/mem usage is. If it doesn't work on either I'll increase the priority on working with streaming and then try another change/pull request for that update.

Mike

pierotofy:
Makes sense. Like you say, streaming should lower memory usage, so if it does we should certainly take advantage of it.

I'm running tests on a wide range of datasets via https://github.com/OpenDroneMap/oats and will share my results.

pierotofy:
Ran the test suite with no errors.

Did some benchmarking between PDAL 1.6 and PDAL 1.8 (nostream). It seems like memory usage is  on par for DSM/DTM generation.

image

Other calls to pdal are of `pdal translate` nature, so I'm comfortable leaving them with stream mode.

`odm_filterpoints` and `odm_georef` built properly and seem to work correctly.

I will merge this sometimes tomorrow unless there are objections.

smathermather:
No objections here!

pierotofy:
Merged, thanks @mikethefarmer !

Issue #931, opened 2019-01-07
Automatically Skip Corrupted Files


pierotofy:
It would be an interesting enhancement to automatically detect and skip corrupted files at the beginning of the process (during dataset loading), without having to wait for a later stage of the pipeline to detect it. It seems to be a fairly common issue, especially when lower quality SD cards are used for data collection.

```
Processing view view_0054.mve...
Processing view view_0055.mve...
Error loading: /var/www/data/ca562c41-e579-4255-bd16-a2103a369669/images/DJI_0487.JPG
Traceback (most recent call last):
File "/code/run.py", line 47, in <module>
plasm.execute(niter=1)
File "/code/scripts/smvs.py", line 70, in process
system.run('%s %s %s' % (context.makescene_path, tree.mve_path, tree.smvs))
File "/code/opendm/system.py", line 34, in run
raise Exception("Child returned {}".format(retcode))
Exception: Child returned 1
```

pierotofy:
Fixed

PR #968, opened 2019-04-09
Extract_utm_coords replacement for C++ program

### LINKED ISSUE #946: OdmExtractUtm Cannot read EXIFs for Mavic 2 Zoom
Trying to process images from the Mavic 2 Zoom causes odm_extract_utm to fail with:

```
Error in OdmExtractUtm:
Failed parsing GPS position for DJI_0076.jpg
```

Related thread with images sample: https://community.opendronemap.org/t/i-can-not-create-an-orthophoto/1759

pierotofy:
I've rewrote the UtmExtractor program in Python, which will allow for the removal of a few dependencies, speed up runtime and compile time as well as making support for EXIF tags more consistent throughout. I believe this will also fix #946 since we won't have to rely on Exiv2.

TODO before merging:
 - Test south UTM zones
 - Remove old C++ module
 - Remove dependencies from configure.sh, docker files


pierotofy:
Confirm that this now fixes the Mavic 2 Zoom dataset.

image

We should now have much better support for cameras that fail to record an altitude value as well, thanks to some tweaks to OpenSfM's config.

pierotofy:
This also solves #931 since `makescene` has been modified to skip invalid views (https://github.com/simonfuhrmann/mve/commit/eed390befebd8fbf87a3478543f166f637713f99)

PR #1001, opened 2019-06-25
Camera Calibration

pierotofy:
This PR adds support for computing a camera model(s) with a good dataset and reusing the same camera model(s) parameters for another dataset.

This is good for datasets that have been captured in less than ideal conditions but for which a previous good dataset has yielded good camera calibration parameters.

Camera models are exported in `/project_path/cameras.json` and can be re-used via `--cameras /path/to/cameras.json` or by specifying the JSON via `--cameras '{"contents_of_cameras.json": ...}'`

cc @smathermather


pierotofy:
The camera parameters can also be manually extracted from a calibration target via Adobe Lens Profile Creator (and its corresponding .LCP file), although a converter tool might be needed to make this less painful. It's still better to just fly a good dataset compared to using a calibration target.

lupionpe:
@pierotofy what would be?
> *just fly a good dataset compared*

I want to do a good calibration for my DJI Mavic Pro and I want to know:
- Do I have to use Ground Control Points?
- Is there a minimum number of photos?
- Should I change angles and altitudes?

I have read some suggestions on the community, but I'm not sure of the basics. On your book I did not find much on that.

PR #1003, opened 2019-06-28
split-merge camera models handling, ODM_DEBUG --> ODM_INFO, debug flag

pierotofy:
Adds camera models (`--cameras`) support for split-merge.

Changes ODM_DEBUG to ODM_INFO where appropriate, adds a `--debug` flag to enable debug messages.


smathermather:
I am curious -- now that we have camera model import and export, should we start putting together a library of camera parameters?

pierotofy:
The thing about having a library is that cameras are manufactured differently (even if they are the same model), so there could be differences in camera parameters between identical models. It would be an interesting experiment to create one though, especially for the popular drones and compare results.

smathermather:
Yes, even if made the same, shaking, temperature changes, and a host of other factors can make even the same camera different from day-to-day. But, arguably, most data collection geometries are sub-optimum and a good best guess may be better than the estimates generated on poor scene geometry.

So -- is this it's own repo, or part of ODM? 😸

pierotofy:
I'd say own repo, then we can have some integration in ODM to access its contents. The repo should have instructions on how to capture a good dataset for purposes of achieving a good camera calibration model. We should accept only camera models that have been flown with best practices.

PR #1004, opened 2019-06-30
Removed --images arg

pierotofy:
Images are expected to be in `project-path/images`. Specifying any other folder can break other parts of the pipeline and I'm not even sure this is used by anyone. Simplifies things.


smathermather:
Just by me I imagine, and I can change my practices... .

PR #1017, opened 2019-08-13
Add force-gps flag

nchamo:
Based on [this](https://community.opendronemap.org/t/how-to-use-high-precision-images-together-with-gcps/2618) forum post.

**Context**
We currently have high precision gps that we add to the images' exif tags. We want to make sure that OpenSFM uses this information, even if we also used GCPs. 

Per @pierotofy 's recommendation, we want to add a flag that allows us to use both GCPs and the images' gps for the reconstruction.

Let me know if you have any other questions :)


pierotofy:
Looks great @nchamo ! Thanks for the contribution! :+1:

PR #1025, opened 2019-09-12
Do not rerun entwine point cloud on merge if dir already exists

pierotofy:
Useful in case of a rerun where the .laz model could not be generated.


smathermather (approved):
Looks good. Thanks!

PR #1026, opened 2019-09-13
Export cameras during remote sm processing

pierotofy:
Exports `cameras.json` file for remotely processed submodels.


smathermather (approved):
Awesome.

PR #1027, opened 2019-09-14
Epsilon bounds comparison for float imprecision + BIGTIFF ifsafer

pierotofy:
This PR fixes problems with DEM merging in certain split-merge datasets. (WIP)


pierotofy:
I'm wondering whether we should remove `--orthophoto-bigtiff` and simply leave the BIGTIFF="IF_SAFER" option for all GeoTIFFs. @smathermather is there a reason for choosing an option other than "IF_SAFER" for most use cases? I know some older GIS programs do not support BigTIFF, but I can't think of other reasons.

smathermather:
Everyone uses GDAL, and thus supports it at this point. I haven't run into software in the wild that doesn't support bigtiff for probably a decade.

If something doesn't support bigtiff, the toolchain failing won't help, so I don't think there's any advantage to not including the flag.

smathermather (approved):
👍

katarias:
@pierotofy @smathermather 
I have just upgraded ODM 2 and Processing datasets with that. 
Default IF_SAFER option works well for mostly datasets but i get failure for one large dataset have 2700 Images.

[INFO]    running gdal_translate -co NUM_THREADS=38 -co TILED=YES -co COMPRESS=DEFLATE --config GDAL_CACHEMAX 46.7% /mnt/execution/8214/odm_dem/merged.vrt /mnt/execution/8214/odm_dem/tiles.tif^[[0m
Input file size is 52219, 57134
0...10...20...30...40...50...60...70...80...90...ERROR 1: TIFFAppendToStrip:Maximum TIFF file size exceeded. Use BIGTIFF=YES creation option.
ERROR 1: TIFFAppendToStrip:Maximum TIFF file size exceeded. Use BIGTIFF=YES creation option.
ERROR 1: TIFFAppendToStrip:Maximum TIFF file size exceeded. Use BIGTIFF=YES creation option.

Is there any way to pass BIGTIFF flag.

pierotofy:
That actually looks like a bug, which will be fixed with this (once it gets merged): https://github.com/OpenDroneMap/ODM/pull/1210/commits/40bda5bd7affbce892ad964b8718bbafc84f903d

katarias:
Thanks for fixing bug. It should work now

smathermather:
Aha, and me with a process running on 5000 images for 90+ hours with this bug in place... :D. I'll see if I can patch this live. Thanks for the fix Piero!

PR #1038, opened 2019-10-03
requirement updated

sayanmondal2098:


sayanmondal2098:
Just add the support of tiff Images . iT'll return the tiff image array in return .

PR #1049, opened 2019-11-04
Use DSM's euclidean map for DTM merging

pierotofy:
Improves DTM merging by using the DSM euclidean map, performing NN interpolation in areas where rasters overlap but have no direct elevation values.


smathermather:
dtmmergezoom

smathermather (commented):
Looking good. Using it in production.

PR #1069, opened 2020-02-11
SIFT features

pierotofy:
This PR adds support for a new option `--feature-type` which allows to select the feature extraction algorithm `SIFT` or `HAHOG` (current default).

SIFT is patented until `March 7th 2020`, so code in this PR should be used only for research purposes.

The OpenCV contrib modules has been stripped to include only the relevant SIFT code (other nonfree modules have been removed).

Do not merge prior to March 7th.

### DIFF
--- a/SuperBuild/cmake/External-OpenCV.cmake
+++ b/SuperBuild/cmake/External-OpenCV.cmake
@@ -1,7 +1,29 @@
 set(_proj_name opencv)
 set(_SB_BINARY_DIR "${SB_BINARY_DIR}/${_proj_name}")
 
+ExternalProject_Add(opencv_contrib
+  PREFIX            ${_SB_BINARY_DIR}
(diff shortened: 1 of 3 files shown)


edgarriba:
quick note: OpenCV is planning to include refactored SIFT into core again, probably after summer.
Besides, I also recommend to take a look at vlfeat implementation: https://github.com/menpo/cyvlfeat
/cc @paulinus

pierotofy:
Yep, good point @edgarriba :+1: 

I think once OpenCV is updated we'll bring the changes in ODM accordingly.

kikislater:
Interesting discussion, thank you for sharing !

pierotofy:
4... https://www.youtube.com/watch?v=9jK-NcRmVcw :)

pierotofy:
1... :)

pierotofy:
:boom:

PR #1077, opened 2020-02-29
Updated dockefiles to best practices ready for multi stage builds

skywalkerisnull:
I am wanting to get the docker images to be much smaller to make the update/download process quicker. The first step is to make the docker files inline with the docker best practices: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,3 +12,15 @@ odm_meshing
 odm_georeferencing
 images_resize
 .git
+
+/contrib
+/docs
+/hooks
+/img
+/license
+/tests
+tests.sh
+settings.yml
+code_of_conduct.md
+configure_18_04.sh
+index.html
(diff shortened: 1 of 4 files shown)


pierotofy:
Will review and test within the next few days, but looks pretty good! Thanks @skywalkerisnull

pierotofy:
Note that this will speed up a bit the build process, but will not make the images smaller (we usually build using the --squash flag, which already takes care of compressing stages, see numbers for reference):

Current: 3.27 GB
Proposed: 3.27GB

pierotofy (approved):

skywalkerisnull:
The next step I am doing is the multi-stage build to start to trim down the size, just playing with the different packages that are required for the build, vs what is required for running. Just needed to get it into the best practices first.

PR #1083, opened 2020-03-10
Adding Ground Rectification

nchamo:
# Intro
Hey everyone!

As part of my master thesis (on the process of being published), I'm using ODM. I don't want to bore you with the details, but it basically has to do with change detection in forests. Now, one of the biggest problems we had is the fact that DTMs are usually off. This of course happens because the drone's camera can't see the ground when the canopy is dense. Also, the fact that SMRF was built for LiDAR clouds points doesn't help, since canopy points usually end up being classified as ground.

In order to deal with this problem, we built a "Ground Rectification" algorithm that tries to fix this particular pain point. The algorithm can be divided into two: first, a re-classification process that re-classifies wrongly ground points into non-ground points, and then an extension process that fills gaps with new ground points.

# The algorithm

The re-classification process works something like this:
1. Given a pre-classified point cloud, we take only the points that were classified as ground
2. Then, taking only the X & Y coordinates of the points, the cloud in divided into different partitions
3. By using RANSAC, we calculate the best fitting plane for the points in each partition
4. With the plane calculated, we then calculate the vertical distance between each point in the partition and the plane
5. Finally, all points that have a higher distance that a given threshold, are re-classified as non-ground (classification = 1)

Some minor details:
A. Only points that are above the plane will be re-classified. The rest will keep their classification
B. When a plane is calculated, if the angle between the new plane and the XY plane is more than 45 degrees, then it is ignored and all points in the partition keep their classification. This is because when the vertical distance was calculated, a lot of points ended up being re-classified when they shouldn't have.

The extension process is kind of similar:
1. Given a pre-classified point cloud, we take only the points that where classified as ground
2. Then, new 2D points are generated in areas where there are no points nearby. These points are all inside the convex hull of the whole point cloud, and have a distance between them (and the other points in the cloud) that can be configured
3. Then, taking only the X & Y coordinates of the points, the cloud in divided into different partitions
4. After that, the 2D points are matched to the partition that contains them
5. Then, using the original cloud points in each partition, a best fitting plane is calculated using RANSAC
6. With the plane calculated, we calculate the projection of each new 2D point onto the plane, so we can calculate its altitude
7. Finally, all new points are added to the original cloud, with their color being the average color of the points in their respective partitions

Now, the core of these two processes lies in the partition algorithm. We developed 4 different algorithms:

1. A "All in one" partition
This one basically takes the whole point cloud and considers it one big partition. This means that it will only generate one plane. You can see in the image below, that all points have the same color, which means that they belong to the same partition.

example_one

2. Uniform partitions
This algorithm takes the whole point cloud and generates one big partition, with a bounding box around it. Then, it takes the center of the bounding box, and generates four new partitions that have the same size. If all new partitions match a minimum area criteria and minimum points contained criteria, then they are considered valid and the original partition is discarded.

The process then continues until partitions can no longer be divided.   

example_uniform

3. Median partitions
This algorithm is almost the same as the one before, but instead of taking the center of the bounding box, it takes the median of the points. The idea is to try to have the same amount of points in each of the new four partitions, instead of simply dividing by the center of the box.

example_median

4. Surrounding partitions
This last version is a little bit trickier. I can go into more detail if you want, but the main idea is the following:
1. First, areas with no points are detected (similarly to the extension process)
2. Then, we find the points in the point cloud that surround this area, and ground them into a partition
3. If an area is not well surrounded (this usually happens in borders), then other points are searched to bring into the partition
4. Finally, all points that weren't assigned to a partition, will belong to a new partitions of their own

example_surrounding

# Configuration

All these processes allow the configuration of following parameters. The defaults were calculated after some testing:
1. The reclassification algorithm (one of the four described above). Default: median
2. The reclassification threshold. Default 5m
3. The extension algorithm (one of the four described above). Default: surrounding
4. The distance between points during the extensions. Default 5m
5. The min area stopping criteria. Default 750m*m
6. The min number of points stopping criteria. Default 500 points

In this PR, I didn't add the possibility to configure any of this parameters. Users can only choose to execute the whole process with the defaults, or not. Even though these defaults work pretty well, I think it would be nice to expose some of this parameters. Specially the stopping criterias and the reclassification threshold, but I'll let you decide.

# Example datasets
Below, see how some example datasets performed. Take into account that the rectification would probably work better if parameters could be configured for each individual case.

## Aukerman
Normal:
dtm_normal

Rectified:
dtm_rectified

## Toledo
Normal:
dtm_normal_toledo

Rectified:
dtm_rectified_toledo

# Conclusion
I think that's it!

Let me know if you think it makes sense to add this to ODM, or if I should just keep it on my fork :)

Also, as always, comments/changes are more than welcome!

### DIFF
--- a/docker.settings.yaml
+++ b/docker.settings.yaml
@@ -62,3 +62,4 @@ project_path: '/' #DO NOT CHANGE THIS OR DOCKER WILL NOT WORK. It should be '/'
 #build_overviews: FALSE
 #pc-classify: none
 #force_gps: False
+#rectify: False
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -595,11 +595,11 @@ def config():
                           'This flag is useful if you have high precision GPS measurements. '
                           'If there are no GCPs, this flag does nothing. Default: %(default)s'))
 
-    parser.add_argument('--rectify',
+    parser.add_argument('--pc-rectify',
                     action='store_true',
                     default=False,
-                    help=('Perform ground rectification. This means that wrongly classified ground '
-                          'points will be re-classified and gaps will be filled. '
+                    help=('Perform ground rectification on the point cloud. This means that wrongly classified ground '
+                          'points will be re-classified and gaps will be filled. Useful for generating DTMs. '
                           'Default: %(default)s'))
 
     args = parser.parse_args()
@@ -616,8 +616,8 @@ def config():
       log.ODM_INFO('Fast orthophoto is turned on, automatically setting --skip-3dmodel')
       args.skip_3dmodel = True
 
-    if args.rectify and not args.pc_classify:
-      log.ODM_INFO("Groudn rectify is turned on, automatically turning on point cloud classification")
+    if args.pc_rectify and not args.pc_classify:
+      log.ODM_INFO("Ground rectify is turned on, automatically turning on point cloud classification")
       args.pc_classify = True
 
     if args.dtm and not args.pc_classify:
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -595,6 +595,13 @@ def config():
                           'This flag is useful if you have high precision GPS measurements. '
                           'If there are no GCPs, this flag does nothing. Default: %(default)s'))
 
+    parser.add_argument('--rectify',
+                    action='store_true',
+                    default=False,
+                    help=('Perform ground rectification. This means that wrongly classified ground '
+                          'points will be re-classified and gaps will be filled. '
+                          'Default: %(default)s'))
+
     args = parser.parse_args()
 
     # check that the project path setting has been set properly
@@ -608,6 +615,10 @@ def config():
     if args.fast_orthophoto:
       log.ODM_INFO('Fast orthophoto is turned on, automatically setting --skip-3dmodel')
       args.skip_3dmodel = True
+
+    if args.rectify and not args.pc_classify:
+      log.ODM_INFO("Groudn rectify is turned on, automatically turning on point cloud classification")
+      args.pc_classify = True
 
     if args.dtm and not args.pc_classify:
       log.ODM_INFO("DTM is turned on, automatically turning on point cloud classification")
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -8,6 +8,7 @@ import shutil
 from opendm.system import run
 from opendm import point_cloud
 from opendm import io
+from opendm import system
 from opendm.concurrency import get_max_memory
 from scipy import ndimage
 from datetime import datetime
@@ -18,6 +19,7 @@ except:
     import queue
 import threading
 
+from .ground_rectification.rectify import run_rectification
 from . import pdal
 
 def classify(lasFile, scalar, slope, threshold, window, verbose=False):
@@ -27,6 +29,46 @@ def classify(lasFile, scalar, slope, threshold, window, verbose=False):
         pdal.run_pdaltranslate_smrf(lasFile, lasFile, scalar, slope, threshold, window, verbose)
     except:
         raise Exception("Error creating classified file %s" % fout)
+
+    log.ODM_INFO('Created %s in %s' % (os.path.relpath(lasFile), datetime.now() - start))
+    return lasFile
+
+def rectify(lasFile, debug=False, reclassify_threshold=5, min_area=750, min_points=500):
+    start = datetime.now()
+
+    try:
+        # Currently, no Python 2 lib that supports reading and writing LAZ, so we will do it manually until ODM is migrated to Python 3
+        # When migration is done, we can move to pylas and avoid using PDAL for convertion
+        tempLasFile = os.path.join(os.path.dirname(lasFile), 'tmp.las')
+
+        # Convert LAZ to LAS
+        cmd = [
+            'pdal',
+            'translate',
+            '-i %s' % lasFile,
+            '-o %s' % tempLasFile
+        ]
+        system.run(' '.join(cmd))
+
+        log.ODM_INFO("Rectifying {} using with [reclassify threshold: {}, min area: {}, min points: {}]".format(lasFile, reclassify_threshold, min_area, min_points))
+        run_rectification(
+            input=tempLasFile, output=tempLasFile, debug=debug, \
+            reclassify_plan='median', reclassify_threshold=reclassify_threshold, \
+            extend_plan='surrounding', extend_grid_distance=5, \
+            min_area=min_area, min_points=min_points)
+
+        # Convert LAS to LAZ
+        cmd = [
+            'pdal',
+            'translate',
+            '-i %s' % tempLasFile,
+            '-o %s' % lasFile
+        ]
+        system.run(' '.join(cmd))
+        os.remove(tempLasFile)
+
+    except:
+        raise Exception("Error rectifying ground in file %s" % lasFile)
 
     log.ODM_INFO('Created %s in %s' % (os.path.relpath(lasFile), datetime.now() - start))
     return lasFile
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -67,8 +67,8 @@ def rectify(lasFile, debug=False, reclassify_threshold=5, min_area=750, min_poin
         system.run(' '.join(cmd))
         os.remove(tempLasFile)
 
-    except:
-        raise Exception("Error rectifying ground in file %s" % lasFile)
+    except Exception as e:
+        raise Exception("Error rectifying ground in file %s: %s" % (lasFile, str(e)))
 
     log.ODM_INFO('Created %s in %s' % (os.path.relpath(lasFile), datetime.now() - start))
     return lasFile
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -28,7 +28,7 @@ def classify(lasFile, scalar, slope, threshold, window, verbose=False):
     try:
         pdal.run_pdaltranslate_smrf(lasFile, lasFile, scalar, slope, threshold, window, verbose)
     except:
-        raise Exception("Error creating classified file %s" % fout)
+        log.ODM_WARNING("Error creating classified file %s" % lasFile)
 
     log.ODM_INFO('Created %s in %s' % (os.path.relpath(lasFile), datetime.now() - start))
     return lasFile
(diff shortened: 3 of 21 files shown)


pierotofy:
Hey @nchamo :hand: I think this is a great addition! I wouldn't worry about exposing too many flags/knobs for the time being, we typically will hear back from the community if the defaults are too restrictive. :+1: 

I think this could be merged, but I wasn't able to get it to work, I get different errors with different datasets (but maybe it's late at night and I missed something):

input laz: https://drive.google.com/open?id=1vD6YCGyj5IH9Pe04WVi4j96IwYIM6zY0
(images): https://community.opendronemap.org/t/brighton-beach/56

```
   raise Exception("Error rectifying ground in file %s: %s" % (lasFile, str(e)))
Exception: Error rectifying ground in file /datasets/brighton2/odm_georeferencing/odm_georeferenced_model.laz: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)

While executing:  | qhull d Qt Qbb Qz Qc Q12
Options selected for Qhull 2015.2.r 2016/01/18:
  run-id 1055369975  delaunay  Qtriangulate  Qbbound-last  Qz-infinity-point
  Qcoplanar-keep  Q12-no-wide-dup  _pre-merge  _zero-centrum  Qinterior-keep
  Pgood  _max-width 20  Error-roundoff 5e-09  _one-merge 3.5e-08
  Visible-distance 1e-08  U-coplanar-distance 1e-08  Width-outside 2e-08
  _wide-facet 6e-08

precision problems (corrected unless 'Q0' or an error)
      2 flipped facets
      2 degenerate hyperplanes recomputed with gaussian elimination
      4 nearly singular or axis-parallel hyperplanes
      2 zero divisors during back substitute
      2 zero divisors during gaussian elimination

The input to qhull appears to be less than 3 dimensional, or a
computation has overflowed.

Qhull could not construct a clearly convex simplex from points:
- p2(v3): 5.8e+05 5.2e+06 0.00017
- p4(v2): 5.8e+05 5.2e+06    20
- p1(v1): 5.8e+05 5.2e+06 0.00017
- p0(v0): 5.8e+05 5.2e+06     0

The center point is coplanar with a facet, or a vertex is coplanar
with a neighboring facet.  The maximum round off error for
computing distances is 5e-09.  The center point, facets and distances
to the center point are as follows:

center point 5.767e+05 5.188e+06        5

facet p4 p1 p0 distance=    0
facet p2 p1 p0 distance=    0
facet p2 p4 p0 distance=    0
facet p2 p4 p1 distance=    0

These points either have a maximum or minimum x-coordinate, or
they maximize the determinant for k coordinates.  Trial points
are first selected from points that maximize a coordinate.

The min and max coordinates for each dimension are:
  0:  5.767e+05  5.767e+05  difference=   20
  1:  5.188e+06  5.188e+06  difference=    0
  2:         0        20  difference=   20

If the input should be full dimensional, you have several options that
may determine an initial simplex:
  - use 'QJ'  to joggle the input and make it full dimensional
  - use 'QbB' to scale the points to the unit cube
  - use 'QR0' to randomly rotate the input for different maximum points
  - use 'Qs'  to search all points for the initial simplex
  - use 'En'  to specify a maximum roundoff error less than 5e-09.
  - trace execution with 'T3' to see the determinant for each point.

If the input is lower dimensional:
  - use 'QJ' to joggle the input and make it full dimensional
  - use 'Qbk:0Bk:0' to delete coordinate k from the input.  You should
    pick the coordinate with the least range.  The hull will have the
    correct topology.
  - determine the flat containing the points, rotate the points
    into a coordinate plane, and delete the other coordinates.
  - add one or more points to make the input full dimensional.
```

And another dataset: https://drive.google.com/open?id=1J-OeinyjHL-z5X1OyNQ0SGu2Ik0y_Vun
(images): https://community.opendronemap.org/t/sheffield-cross/570

```
   raise Exception("Error rectifying ground in file %s: %s" % (lasFile, str(e)))
Exception: Error rectifying ground in file /datasets/sheffield_cross/odm_georeferencing/odm_georeferenced_model.laz: QH6013 qhull input error: input is less than 3-dimensional since it has the same x coordinate

While executing:  | qhull d Qt Qbb Qz Qc Q12
Options selected for Qhull 2015.2.r 2016/01/18:
  run-id 1061823863  delaunay  Qtriangulate  Qbbound-last  Qz-infinity-point
  Qcoplanar-keep  Q12-no-wide-dup  _pre-merge  _zero-centrum  Qinterior-keep
  Pgood  _max-width 25  Error-roundoff 3e-09  _one-merge 2.1e-08
  Visible-distance 6e-09  U-coplanar-distance 6e-09  Width-outside 1.2e-08
  _wide-facet 3.6e-08
```

`./run.sh --pc-rectify --rerun-from odm_dem --dtm --project-path /datasets sheffield_cross`

(I changed `--rectify` to `--pc-rectify`, hope that's all right)

pierotofy:
`pip freeze`

```
aenum==2.2.3
affine==2.3.0
appsettings==0.8
attrs==19.1.0
autobahn==0.10.3
certifi==2019.11.28
cffi==1.13.2
chardet==3.0.4
Click==7.0
click-plugins==1.1.1
cligj==0.5.0
cloudpickle==1.2.2
cryptography==2.8
cycler==0.9.0
decorator==4.4.1
enum34==1.1.2
ExifRead==2.1.2
Fiona==1.8.9.post2
futures==3.0.5
GDAL==2.2.2
gpxpy==1.3.5
idna==2.8
ipaddress==1.0.16
joblib==0.13.2
laspy==1.6.0
loky==2.5.1
lz4==0.7.0
matplotlib==1.5.1
mpi4py==1.3.1
msgpack-python==0.4.6
munch==2.5.0
networkx==2.2
numpy==1.15.4
Pillow==6.1.0
psutil==5.6.3
pyasn1==0.1.9
pyasn1-modules==0.0.7
pycparser==2.19
pycurl==7.43.0
pyodm==1.5.2b1
pyOpenSSL==19.1.0
pyparsing==2.4.6
pyproj==2.2.2
Pysolar==0.6
Pyste==0.9.10
python-apt==1.1.0b1+ubuntu0.16.4.8
python-dateutil==2.4.2
python-snappy==0.5
pytz==2014.10
PyYAML==3.13
rasterio==1.0.28
repoze.lru==0.7
requests==2.21.0
requests-toolbelt==0.9.1
scikit-learn==0.20.0
scipy==1.2.1
service-identity==16.0.0
simplejson==3.17.0
six==1.10.0
snuggs==1.4.7
trollius==2.0.1
Twisted==16.0.0
txaio==1.0.0
urllib3==1.24.1
xmltodict==0.12.0
zope.interface==4.1.3
```

nchamo:
I'll check those cases to see if I can get it fixed. Apparently, it can be fixed by changing some Qhull options

https://github.com/mikedh/trimesh/issues/24

linusmartensson:
I'm really interested in the feature, but curious about the pictures. What's the regular grid of equidistantly spaced dips you see in a few of the rectified maps? It looks like it's mostly close to nodata-areas of the map, but looks strange.

nchamo:
Hey @linusmartensson !

Those are the added points from the extension process. I'm not sure why sometimes they are visible on the DTM and sometimes they aren't, but if you look at the picture below, you will see in red the added points. It is necessary to add these points, in order to avoid having gaps in the DTM . These gaps became present when we re-classify the trees

Screenshot from 2020-03-10 22-32-40

nchamo:
Hey @pierotofy !

I made a fix and both clouds seems to be working now.

I even run the whole pipeline on brighton beach and it worked

Normal:
dtm_normal

Rectified:
dtm_rectified

nchamo:
And @linusmartensson , I dug a little bit dipper into why the points appeared.

The problem seems to be that some tree points were too low, and the first step of the process didn't re-classify them as non-ground. Since that happened, the estimated plane wasn't really good and the added points weren't as flat as one would like.

I can think of two ways to fix the problem:
A. Lower the re-classification threshold
B. Increase the distance between added points, so there are less points in that place.

With the default parameters, rectification won't be perfect in all cases, but maybe later on we can expose these parameters so people can adjust them to scenarios.

Screenshot from 2020-03-11 09-21-20

pierotofy:
Looks good, thanks for the contribution @nchamo !

nchamo:
Glad to help!

@pierotofy , I have a question about how to implement the other part of my thesis. 

How can we discuss it? Should I send you an email? Create a post on the forum?

pierotofy:
A post on the forum would be best, this way others can join on the discussion. :+1:

smathermather:
This is fantastic! Thanks for the contribution. I am looking forward to playing with this.

smathermather:
Digging the ground rectification tool. I have some challenging datasets that might help with finding better defaults or inspire knobs for controlling the variables. Is this something of interest @nchamo?

nchamo:
Hey @smathermather!

I'm sorry, but I currently don't have a lot of time to dedicate to this.

However, if someone else wants to play with all the variables described above (in the `Configuration` section), I could set up a branch/fork that exposes all the them.

smathermather:
If you have the bandwidth to create a branch that exposes the variables, I am happy to do the testing.

nchamo:
Hey @smathermather !

[Here](https://github.com/OpenDroneMap/ODM/compare/master...nchamo:expose-rectification-config?expand=1) are the changes that expose all available config. To be honest, I wasn't able to test it because my Ubuntu version is too old, and I haven't had the chance to upgrade yet. But it should be pretty straight forward.

Each config variable is defined in this PR, or on the code I just added. However, it's kind of tricky to get your head around it, so let me know if you need any help with understanding them.

And let me know if I can assist you with something else!

smathermather:
Cool. Doing holiday at the moment but will test on return. Thanks!

usman-rind:
https://github.com/OpenDroneMap/ODM/assets/60088877/ac4db56f-ccdd-4701-9122-464018f41150


We are using DJI Mavic 3 M and stitching multispectral images but the bands of the stitched image are not aligned as you can see in the attached video.

could you please help why this is happening and how can we fix that?
Is any of the above mentioned plugin or technique that can be used or any parameters of ODM that can be used to improve and reduce or eliminate this misalignment of different bands.

usman-rind:
> Bands.Misalignment.mp4 
> We are using DJI Mavic 3 M and stitching multispectral images but the bands of the stitched image are not aligned as you can see in the attached video.
> 
> could you please help why this is happening and how can we fix that? Is any of the above mentioned plugin or technique that can be used or any parameters of ODM that can be used to improve and reduce or eliminate this misalignment of different bands.

Guys, we further tested stitching on the Pix 4d Fields and there is no misalignment within the bands, could you please how this can be fixed in ODM. is there any related open issue

PR #1096, opened 2020-03-30
Optimize disk space

pierotofy:
This PR adds support for a new flag, `--optimize-disk-space`, which eliminates heavy intermediate files as the pipeline runs.

Pros: process larger datasets with less HD space available.
Cons: cannot restart the pipeline from an intermediate stage.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -702,6 +702,15 @@ def config(argv=None):
                           'This flag is useful if you have high precision GPS measurements. '
                           'If there are no GCPs, this flag does nothing. Default: %(default)s'))
 
+    parser.add_argument('--optimize-disk-space',
+                action=StoreTrue,
+                nargs=0,
+                default=False,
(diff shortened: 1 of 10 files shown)


smathermather:
Brilliant.

smathermather:
Can WebODM run on intermediate products? If not, should this be default in WebODM?

pierotofy:
WebODM can run on intermediate products, if the processing node is local, for up to 2 days (before NodeODM purges intermediate results). I'd let users decide whether to use it or not.

smathermather:
Ahh, I wondered why those options are exposed there. I'm never using a local node... . Maybe I should change that.

pierotofy:
I should mention, this can be a remote node too, it just need *not* to use S3 uploads (like webodm.net). If the options are exposed, it should be able to use it (they are hidden when this can't be used).

PR #1097, opened 2020-04-02
Better band name support

pierotofy:
Should add better support for Agrowing cameras as reported on https://community.opendronemap.org/t/multispectral-photo-getting-started-question/3807/3


gastgit:
Thank you very much for this merge. Testing the multi-spectral sample set now..

Issue #1076, opened 2020-02-28
Undistort stage drops multi-track reconstructions


linusmartensson:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Chrome 79 on Linux

### What is the problem?

When running the undistort and following steps, data is missing on multi-reconstruction datasets. 

If OpenSfM fails to create a single reconstruction, it automatically splits a dataset into multiple individual reconstructions which need to be managed individually using the "--reconstruction-index" parameter in the 'undistort' step.

If this parameter is not supplied, OpenSfM defaults to 0, and undistorts only the first reconstruction. This is readily apparent if you pick up any dataset with a partial reconstruction result in the iterative reconstruction:
```
2020-02-28 12:03:21,709 INFO: Reconstruction 0: 25 images, 8340 points
2020-02-28 12:03:21,709 INFO: Reconstruction 1: 25 images, 6582 points
2020-02-28 12:03:21,709 INFO: Reconstruction 2: 21 images, 4059 points
2020-02-28 12:03:21,709 INFO: Reconstruction 3: 14 images, 4888 points
2020-02-28 12:03:21,709 INFO: Reconstruction 4: 10 images, 3338 points
2020-02-28 12:03:21,709 INFO: Reconstruction 5: 8 images, 1121 points
2020-02-28 12:03:21,709 INFO: 6 partial reconstructions in total.
[INFO]    Aligning submodels...
```
Following this, pass through the undistortion stage, and compare reconstruction.json to undistorted/reconstruction.json, images to the opensfm/undistorted/images folder.

You'll see that for a result like above, the undistorted folder will have the same number of images as in the first reconstruction path, rather than the number of images in the submodel or project.

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

I was expecting that the undistortion stage be run for each partial reconstruction, then aligned and merged. Given the nature of these datasets, it may be possible to treat them similar to submodels and merge them in a similar manner?

[...]

PR #1103, opened 2020-04-27
Recreate splits after reconstruction

linusmartensson:
Adapted from discussion in #1076 

In short, after reconstruction finishes, this PR will ensure each split only has a single reconstruction track by regenerating the splits.  Without this PR, additional tracks in a reconstruction are dropped as part of the track exports and point cloud generation, resulting in possibly lost areas of a map if large secondary tracks are generated in a split.

I didn't have everything available to test this outside docker, so I'd appreciate if someone could help me out testing this before we merge. It's also possible that there's more ODM oriented APIs for handling some of the file management, so there may be some style issues.

For future work, the resulting set could probably be optimized by discarding any tracks whose images are fully contained in another, longer track. 
Additionally, you could probably get better results if a full bundle adjustment could be done on the resulting reconstruction data at the alignment stage rather than just the current similarity transform, since the multiple tracks may be an indication that the initial split was not ideal for the current dataset. I've noticed some issues relating to this where a track is misaligned on the vertical axis, but this seems common with small split datasets and low overlaps, so your mileage may vary.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -674,6 +674,12 @@ def config(argv=None):
                         'are added to the cluster. This is done to ensure '
                         'that neighboring submodels overlap.')
 
+    parser.add_argument('--split-multitracks',
+                       action=StoreTrue,
+                       nargs=0,
+                       default=False,
+                       help='Split multi-track reconstructions.')
+
     parser.add_argument('--sm-cluster',
                         metavar='<string>',
                         action=StoreValue,
--- a/stages/splitmerge.py
+++ b/stages/splitmerge.py
@@ -1,5 +1,7 @@
 import os
 import shutil
+import json
+import yaml
 from opendm import log
 from opendm.osfm import OSFMContext, get_submodel_argv, get_submodel_paths, get_all_submodel_paths
 from opendm import types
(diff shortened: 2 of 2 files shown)


pierotofy:
Hey @linusmartensson :hand: thanks for the PR! Would love to help test this, do you have a dataset (and parameters to use) that exhibits multitracks during split merge?

linusmartensson:
@pierotofy I checked if I could share the one I have, but unfortunately not. :( 

I think it should be fairly easily to reproduce on other datasets though, as I had similar multitrack results in all my submodels. For reference, the dataset was about 1000 images covering a forest environment. I'm fairly sure a dense forest plays some part in the matcher failures causing track splits, but you should be able to reproduce the issue in other environments as well. 

Either way, I preprocessed the dataset with the following:
```
docker run --network="host" -ti --rm -v ~/dataset:/datasets/code opendronemap/odm --project-path="/datasets" --resize-to 2000 --min-num-features 4000 --matcher-neighbors 8 --camera-lens brown --mve-confidence 0 --use-opensfm-dense --opensfm-depthmap-method BRUTE_FORCE --depthmap-resolution 320 --end-with split --verbose --split 100 --split-overlap 25
```
Then I did my testing on multitrack handling from there.

In effect, you probably only need to run up to the end of the initial reconstructions to see if you have splits, so try something like this and go with defaults on what you can:
```
docker run --network="host" -ti --rm -v ~/dataset:/datasets/code opendronemap/odm --project-path="/datasets" --resize-to 2000 --min-num-features 4000 --matcher-neighbors 8 --camera-lens brown --verbose --split 100 --split-overlap 25 --split-multitracks
```
YMMV, but reducing the feature count should give you matching failures sooner or later.

With --split-multitracks, you get a new "resplit_done.txt" in the root opensfm folder that skips the split, and your original submodels will be moved from "submodels" to "unaligned_submodels" if you want to retry the mechanism. The submodels are at the moment aligned after resplits, so you'll have to delete alignment_done.txt if that exists in your root as well.

Could this work to get you started?

pierotofy:
That's all right, I think this can be merged. If there are problems I'm sure the community will find and report them. 

Thanks for the contribution!

smathermather:
I can run this on a forest dataset that was missing large areas and probably subject to the underlying issue.

Issue #1092, opened 2020-03-24
Feature Request: Additional EXIF and XMP tags supported in weighting embedded GPS Locations


devonjlibby:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Chrome 80 on Windows 10

### What is the problem?

My current understanding is that ODM can use the EXIF tag “GPSDOP” to add more weight to the embedded GPS location and has a default value of “15” .  EXIF and XMP tags have emerged that quantify location accuracy more specifically, particularly important for RTK systems.  ODM doesn't take advantage of these tags in imagery from RTK and other systems that write these tags, resulting in mosaics with lower spatial accuracy than stitching software that does use the tags. 

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

As an enhancement request to accommodate RTK and other systems which write these tags, could you consider the following?

The ability to use other tags to weight or provide accuracy assessments that can be applied to the embedded GPS location.  The requested tags are the EXIF tag “GPSHPositioningError” (standard EXIF tag since EXIF version 2.3, released April 2010), and the XMP tags “GPSXYAccuracy” and “GPSZAccuracy” defined by Pix4D (https://support.pix4d.com/hc/en-us/articles/360016450032-Specifications-of-xmpcamera-tags). This could be done by a user specifying which tag(s) to use or by a search hierarchy in which it looks for “‘GPSXYAccuracy’ and ‘GPSZAccuracy’”, then “GPSHPositioningError”, and finally “GPSDOP”.

pierotofy:
This should be fixed in #1106

PR #1106, opened 2020-05-01
Better support for RTK

pierotofy:
This PR adds support for RTK tags to constraint bundle adjustment and increase accuracy of georeferencing and reconstruction by overriding the GPS DOP values.

anim

Should work with Phantom 4 RTK (proprietary DJI tags) and several other cameras that use the more common `Camera:GPSXYAccuracy` and `Camera:GPSZAccuracy` tags.

Also allows users to override GPS DOP values (cc @smathermather), so no more exiftool magic needed.


pierotofy:
Fixes #1092

PR #1107, opened 2020-05-10
Propose to change default accuracy to 5m

smathermather:
We can expect that most drones achieve an accuracy of 5m or better, even relatively fast fixed wings usually see better than this. If we have a scene with repeating features (e.g. crops) or complicated matching (e.g. forests) we can get really convincing false matches, that result in some really messy reconstructions.

Setting the default accuracy to something more realistic will help prevent some of these problem cases:

image

-->

pc

(footnote: OpenSfM's default of 15m probably makes lots of sense for urban canyons)


pierotofy:
:+1: for changing the accuracy, however you can't just change the default config value, OpenSfM has a value of `15` hard-coded in the source, so you'll need to change it there too or change the ODM logic to always write a `exif_overrides.json` file.

smathermather:
I was wondering about that. It looks like we no longer maintain our own modified OpenSfM, which is sensible: https://github.com/OpenDroneMap/ODM/blob/master/SuperBuild/cmake/External-OpenSfM.cmake

Which... should we be retiring this repo?:
https://github.com/OpenDroneMap/OpenSfM

pierotofy:
Actually we still use a fork (with some important changes that could not be merged in OpenSfM core https://github.com/OpenDroneMap/ODM/blob/master/SuperBuild/cmake/External-OpenSfM.cmake#L11). The work is on the `099` branch.

The easiest way would be to modify these lines https://github.com/OpenDroneMap/ODM/blob/master/opendm/osfm.py#L106-L133 to always write the override file according to the gps-accuracy value (or if the images provide a gps dop value to use that one).

pierotofy:
Once #1156 is merged, this PR will work fine (it will override the OpenSfM default).

It would be good to run some tests before merging this though; it's possible that a lower value would lead to an over-constrained bundle adjustment setup. In the few datasets I've processed (small ones) the accuracy actually seemed to decrease (when visually overlayed to Google Maps). 

We can still change the default, but it would be good to run some tests/collect results in a variety of datasets before changing it.

Or we can change it and see what happens; we can always revert.

pierotofy:
>> The accuracy commitments do not apply to GPS devices, but rather to the signals transmitted in space. For example, the government commits to broadcasting the GPS signal in space with a global average user range error (URE) of ?7.8 m (25.6 ft.), with 95% probability. Actual performance exceeds the specification. On May 11, 2016, the global average URE was ?0.715 m (2.3 ft.), 95% of the time.

https://www.gps.gov/systems/gps/performance/accuracy/#how-accurate

Perhaps `10` or `8` might be a more conservative change, while still improving results.

PR #1109, opened 2020-05-15
Generate camera positions output

pierotofy:
This PR adds the generation of `shots.geojson` files containing the geographical coordinates of the cameras used for reconstruction. Works also for pseudo-geo (non-georeferenced) datasets.

image

Handles split-merge as well.


pierotofy:
image

PR #1117, opened 2020-06-04
Fixed path ambiguity and order of examples

SBCV:
The original example `docker run -ti --rm -v /datasets:/datasets opendronemap/odm --project-path /datasets project` appears on the first sight a bit confusing, since `/datasets` represents actually two different paths.


pierotofy:
That's a good clarification, thank you!

PR #1119, opened 2020-06-05
Fixed AttributeError: 'ODM_Tree' object has no attribute 'shots_geojson'

SBCV:


pierotofy:
Nice catch! :+1: thanks!

PR #1131, opened 2020-07-02
Fixed copy point_cloud.ply if --pc-filter -0

### LINKED ISSUE #1128: When `--pc-fileter ` specified, ODM can't work.
### What is the problem?

https://github.com/OpenDroneMap/ODM/blob/c4bbee2881549b8fca5bcaf68fb86d9cae4b8bd0/opendm/point_cloud.py#L14-L16
https://github.com/OpenDroneMap/ODM/blob/c4bbee2881549b8fca5bcaf68fb86d9cae4b8bd0/opendm/point_cloud.py#L31-L35

When using options `--pc-filter 0` , it will return. ODM won't copy `point_cloud.ply`. But some of the steps that follow require it.

Is it possible to copy even when`-- pc-filter 0` is specified.

pqiaohaoq:
Fixed #1128

### DIFF
--- a/opendm/point_cloud.py
+++ b/opendm/point_cloud.py
@@ -11,8 +11,14 @@ def filter(input_point_cloud, output_point_cloud, standard_deviation=2.5, meank=
     """
     Filters a point cloud
     """
+    if not os.path.exists(input_point_cloud):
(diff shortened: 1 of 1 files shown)


pierotofy:
Looks great, thanks @tinyRatP ! :pray:

PR #1132, opened 2020-07-07
odm_filterpoints improvements

pierotofy:
* Upgraded PDAL to 2.1
* Removed odm_filterpoints module

Still need to add parallelization


pierotofy:
This passes the OATS test suite. :+1: 

We've been able to speed up by 50% the filtering time for https://community.opendronemap.org/t/sheffield-cross/570 which has a ~17 million point cloud on a 4 core CPU. Smaller datasets are unlikely to benefit much from the additions in this PR, but larger datasets and/or computers with lots of CPU cores should see significant improvements in runtime.

We split the point cloud in chunks of 250,000 points (which we benchmarked as an ideal compromise between number of splits which increases disk I/O and speed). We then perform filtering in parallel using PDAL. Finally, we merge the chunks using a dedicated fast PLY merging routine we wrote from scratch.

Thanks to korn704, Rumen Mitrev, Brett Edmond Carlock, Corey Snipes, Stephen Mather, Brett Edmond Carlock, JUAN MANUEL MARTINEZ, Karl Kelle, José Israel Villarreal Barragán, HAWK-i.us, Yu-Huang Wang and Yan for sponsoring the work to make this possible!

PR #1136, opened 2020-07-13
Fix bug when generating shots when there are several reconstructions.

magnus-eigenvision:
Previously, only shots from the last reconstruction were included.

This PR mostly changes indentation, so that two for loops that were previously run in sequence are now nested.


pierotofy:
Hey @magnus-eigenvision :hand: nice catch! Changes look great.

Thanks for the patch! :pray:

PR #1138, opened 2020-07-27
Update mvs texturing

pierotofy:
Contains a small fix related to tone mapping.


smathermather (commented):
Looks good. Curious: `How many threads to use. Set 1 for determinism.` Does this mean it's possible to get a deterministic result at the cost of time/threads?

pierotofy:
I think so. Haven't tested though.

smathermather:
> I think so. Haven't tested though.

Good to know. That's an occasional feature request for sure, especially for folks doing stuff like counting plants and similar. It hurts their teeth to not have a deterministic solution, right or wrong. 😁

PR #1141, opened 2020-08-08
Script to prepare TIFs captured by Parrot Sequoia

samsalway:
As posted [here](https://community.opendronemap.org/t/parrot-sequoia-gps-exif-tags/5085), this script will copy GPS Exif tags from the RGB JPG to the four TIFs for the same shot.


pierotofy:
Thanks @samsalway ! :+1:

PR #1144, opened 2020-08-19
Remove unnecessary scripts for sequoia image preparation.

samsalway:
I found these scripts are totally unnecessary, TIFs from the Sequoia _do_ contain GPS tags.  I'm not sure how I ended up with some TIFs without GPS tags, that's a mystery for another day, but I have tons of old images which definitely do.


pierotofy:
Thanks @samsalway !

I think it's safe to remove these too. If somebody still needs them we can always revert this.

kikislater:
Exact ! I was surprised you push previous PR ... 
TIF without coordinates is due to hardware problem : overheating (the main problem), not shielding cables, power problem, over writig speed than one permit by manufacturer, lack of GPS data in sunshine sensor ...

PR #1161, opened 2020-09-28
Try to estimate GSD for flipped Z axis datasets

pierotofy:
Sometimes a dataset will be reconstructed upside down. The program in this case should still try to come up with a GSD estimate to avoid slowdowns in later stages of the pipeline.


kikislater:
Interesting ! Thanks

PR #1167, opened 2020-10-15
Support CMake+Ninja in SuperBuild

lucyllewy:
* Use CMake-native call to `ProcessorCount` instead of a shell execute
  call to `nproc` that only works in Makefile-style builds.
* Use CMake's implicit default `cmake --build` commands to build CMake
  projects instead of assuming `make`.


pierotofy (approved):

pierotofy:
Thanks @diddledan  :+1:

PR #1168, opened 2020-10-15
Add patch to External-PDAL awaiting upstream fix

lucyllewy:
ref: PDAL/PDAL#3263

* PDAL upstream is missing several `#include` definitions that cause
  compilation to fail.
* PDAL upstream is missing linking and include definition for LASZIP in
  its `CMakeLists.txt` files.

### DIFF
--- a/SuperBuild/cmake/External-PDAL.cmake
+++ b/SuperBuild/cmake/External-PDAL.cmake
@@ -10,6 +10,7 @@ ExternalProject_Add(${_proj_name}
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
   URL               https://github.com/PDAL/PDAL/archive/2.2.0.zip
   #--Update/Patch step----------
+  PATCH_COMMAND     patch -Np1 -i "${SB_ROOT_DIR}/../patches/PDAL.diff"
   UPDATE_COMMAND    ""
   #--Configure step-------------
   SOURCE_DIR        ${SB_SOURCE_DIR}/${_proj_name}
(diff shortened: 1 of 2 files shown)


pierotofy:
Can you expand more on the need for this one? :pray: 

We've been building PDAL fine without the patch up to this point (on Ubuntu 18.04, via docker).

lucyllewy:
I'm just rebuilding in a clean environment to double check each fix in the patch. I encountered them when I was attempting to build the SuperBuild into a snapcraft-known location rather than the source/build's SuperBuild directory, so they may be a byproduct of that nonstandard use (which was fraught with errors, anyway, so I've pulled back from that for now :-)).

lucyllewy:
OK, first run, the headers do seem to still be required (at least when building on a Snapcraft-started Ubuntu 18.04 instance). Re-running now to double check the linker bits.

Those headers and the undefined entities are:

- Function `CPLFree()` from missing `<cpl_conv.h>` in:
  - `io/Ilvis2MetadataReader.cpp`
  - `pdal/Geometry.cpp`
- Type `CPLStringList` from missing `#include <cpl_string.h>` in:
  - `pdal/Geometry.cpp`
- Function `OGRWktReadToken()` from missing `#include <ogr_p.h>` in:
  - `pdal/private/gdal/GDALUtils.cpp`

lucyllewy:
Right only the headers are broken. I've pushed a re-rolled commit with just those in the patch

pierotofy:
It's still a bit strange.

Those headers are part of the GDAL lib, perhaps this has do to with the different version of GDAL installed?

```
gdalinfo --version
GDAL 3.0.4, released 2020/01/28
```

Note that we install GDAL (and we need version >= 3) from a ppa https://github.com/OpenDroneMap/ODM/blob/master/configure.sh#L43

lucyllewy:
aha. I completely missed that line when I read the file (snapcraft isn't using that script at the moment)

pierotofy:
No worries! This project definitely has a lots of dependencies, so it's easy to miss something.

lucyllewy:
Yup, we can drop this. The UbuntuGIS PPA resolved things.

PR #1173, opened 2020-10-20
Update reference image

smathermather:


smathermather:
I don't know why Atom decided to change all those durn spaces.

pierotofy:
That's a really nice DSM!

smathermather:
Yeah, imagery via Federico Debetto - KMedia SAS. It's Stone Town's Fort and the Sultan's Palace / Beit-al-Ajaib. He's offered it as a research dataset, so expect to see it soon in with the test datasets.

smathermather:
I mean, this is awefully purdy:
image

Issue #1177, opened 2020-10-27
[Python Refactoring Task] Replace all calls to io.join_paths with os.path.join


pierotofy:
A nice, simple first issue for anyone with Python knowledge. :)

Replace the calls to `io.join_paths` with the standard `os.path.join` (making sure `import os` is at the top of each file) across the codebase for improved readibility.

smathermather:
w00t! Thanks @koushikjoshi!

PR #1179, opened 2020-10-28
Replaced the calls to io.join_paths with the standard os.path.join

koushikjoshi:


smathermather:
Aha, for https://github.com/OpenDroneMap/ODM/issues/1177. Nice!

run.py
```diff
@@ -32,15 +32,15 @@
 
         # Don't leak token
         if k == 'sm_cluster' and args_dict[k] is not None:
-            log.ODM_INFO('%s: True' % k)
+            log.ODM_INFO('%s: True' % k
```
pierotofy: I think this parenthesis was removed by mistake?

pierotofy:
Thanks @koushikjoshi ! See my comment for a small change needed to merge this.

run.py
koushikjoshi: @pierotofy yes you are right. I'm sorry, this was my first open-source contribution. Do you want me to add the paranthesis?

pierotofy:
Looks great! Thanks @koushikjoshi :clap:

PR #1181, opened 2020-10-30
Dan's magic GitHub Action

Saijin-Naib:
This GitHub Action will build the core18-based SNAP for x64 platforms, and upload it to the SNAP store, provided credentials are created and stored in the GitHub repository as a Secret.

x86 platform support is currently non-functional.

### DIFF
--- a/.github/workflows/snapcraftbuild.yml
+++ b/.github/workflows/snapcraftbuild.yml
@@ -8,40 +8,18 @@ on:
 jobs:
   build-and-release:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        architecture:
-        - linux/amd64
-        - linux/386
     steps:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Build
       id: build
-      run: |
-        echo '{"experimental": true}' | sudo tee /etc/docker/daemon.json > /dev/null
-        sudo systemctl restart docker
-        docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
-
-        ######### Change snap name to temporary test name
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -1,4 +1,4 @@
-name: opendronemap
+name: opendronemap-core18-test
```
pierotofy: Should this be just `opendronemap`?

.github/workflows/snapcraftbuild.yml
```diff
@@ -0,0 +1,29 @@
...
+        snapcraft-args: --enable-experimental-package-repositories
+    - name: Review
+      uses: diddlesnaps/snapcraft-review-tools-action@v1
+      with:
+        snap: ${{ steps.build.outputs.snap }}
+        isClassic: 'false'
+    - name: Publish playtest to Edge
+      uses: snapcore/action-publish@v1
+      with:
+        store_login: ${{ secrets.STORE_LOGIN }}
```
pierotofy, line 27: What do we add here? Some token from the snapcraft store? I'm not familiar so I thought I'd ask.

pierotofy:
Thanks @Saijin-Naib ! See the few comments before we merge this :+1:

.github/workflows/snapcraftbuild.yml
Saijin-Naib: Yes. I can generate it and send you the contents on the Discuss.

Saijin-Naib (commented):
Addressed concerns. i think?

I've never done a GitHub PR review before.

pierotofy:
Yep, looks great! I'll setup the secret and merge this :+1: 

Thanks!

PR #1182, opened 2020-11-02
WSL build

lucyllewy:
* Add GitHub Actions workflow to build and release Docker images (requires that the primary repo add GitHub Secrets for Docker Hub access)
* Add GitHub Actions workflow to test that PRs successfully build
* Update `configure.sh` to pull APT dependency information from `snapcraft.yaml` to allow maintainers to only need to update dependency list in one place
* Update Dockerfiles to reduce size of final image by utilising multiple-stage builds allowing us to exclude the build-time requirements from the final image
* Use the Docker image built by the GitHub Workflow to generate and save a WSL rootfs tarball to GitHub Releases (based on tags)

### DIFF
--- a/.github/workflows/build-container-and-wsl.yaml
+++ b/.github/workflows/build-container-and-wsl.yaml
@@ -1,0 +1,70 @@
+name: Build and Release Container and WSL Image
+
+on:
+  push:
+    tags:
+    - v*
+
+jobs:
+  build:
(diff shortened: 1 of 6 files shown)


pierotofy (approved):
This is amazing. Thanks @diddledan !

I'll setup the appropriate docker secrets before merging.

PR #1183, opened 2020-11-02
Update snapcraftbuild workflow

lucyllewy:
* Use newly published action: `diddlesnaps/snapcraft-multiarch-action`
* Upload every push to master to the Snap Store's Edge channel
* Upload every tagged prerelease to the Snap Store's Beta channel
  (including release candidates which are also handled below)
* Upload every tagged stable or release candidate release to the Snap
  Store's Candidate Channel

(I haven't managed to test these workflows' `if:` lines on the publish steps are correct, so it might fail to parse correctly once merged in which case I'll try to fix it) EDIT: I think I've got them correct in 149baa4


pierotofy (approved):
Love it. Thanks!

PR #1184, opened 2020-11-03
Fix snapcraftbuild

lucyllewy:
I _think_ this fixes the yaml parser error from GitHub Actions (ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#example-expression-in-an-if-conditional)


pierotofy:
Let's give it a try! :)

Thanks @diddledan

PR #1186, opened 2020-11-05
Bump scipy dependency version

lucyllewy:
Part of porting to Ubuntu 20.04, we need a more recent scipy:

* Update scipy to 1.5.4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ attrs==19.1.0
 pyodm==1.5.4
 Pillow==6.1.0
 networkx==2.2
-scipy==1.2.1
+scipy==1.5.4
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
I've checked the list of breaking changes, but we should be in the clear (no issues) :+1: 

https://docs.scipy.org/doc/scipy/reference/release.1.3.0.html#backwards-incompatible-changes
https://docs.scipy.org/doc/scipy/reference/release.1.4.0.html#backwards-incompatible-changes
https://docs.scipy.org/doc/scipy/reference/release.1.5.0.html#backwards-incompatible-changes

Thanks @diddledan !

PR #1187, opened 2020-11-05
Fix incorrect libusb APT package version

lucyllewy:
* Update dependency on libusb to use `libusb1.0` instead of `libusb0.1-4`


pierotofy (approved):
Is this to make it compatible with `20.04`? It's been working OK with `18.04`. Just curious.

lucyllewy:
I noted that during build there was a warning about not finding libusb despite having the 0.1 version installed. I'm not sure it's crucial, because the build finishes despite the warning so it might be a "nice to have" dependency.

PR #1188, opened 2020-11-05
Fix PYTHONPATH environment var in Dockerfiles

lucyllewy:
My previous evolution of the Dockerfiles grouped the `ENV` entries into
a single definition to reduce the number of intermediate Docker image
layers. Unfortunately I didn't account for the duplicate `PYTHONPATH`
definition so only the second was being applied correctly. This commit
fixes that.

* Combine multiple `ENV PYTHONPATH=` to a single definition.


pierotofy (approved):
Nice catch! :+1: 

Thanks.

PR #1189, opened 2020-11-05
Update builds for Ubuntu 20.04

lucyllewy:
* Adjust `PYTHONPATH` in Dockerfiles to newer `python3.8`
* Switch to 20.04 base images in Dockerfiles
* Switch snapcraft to use `base20`
* Update APT dependency versions in `snapcraft.yaml` to match those
  available in Ubuntu 20.04
* Bump `pyproj` from version `2.2.2` to `3.0.0`
* Bump `numpy` from version `1.19.2` to `1.19.4`
* Add missing exclusions to `snapcraft.yaml` to reduce the size of snap
  packages
* Add missing exclusions to Dockerfiles to reduce the size of container
  images
* Update `configure.sh` to account for Ubuntu 20.04 instead of 18.04
* Nit: make `configure.sh` executable directly with `./configure.sh`
  instead of `bash configure.sh`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM ubuntu:18.04 AS builder
+FROM ubuntu:20.04 AS builder
 
 # Env variables
 ENV DEBIAN_FRONTEND=noninteractive \
-    PYTHONPATH="$PYTHONPATH:/code/SuperBuild/install/lib/python3.6/dist-packages:/code/SuperBuild/src/opensfm" \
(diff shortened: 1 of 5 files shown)


pierotofy:
Nice!!

I will run some tests to make sure this works with various datasets before merging, but it's looking good!

pierotofy:
:heavy_check_mark: passes the OATS test suite
:heavy_check_mark: builds properly with NodeODM (after https://github.com/OpenDroneMap/NodeODM/commit/438dacb641382ece529aeecef83d7077a0f5ab1b)

Good to go!

Thanks @diddledan !

PR #1191, opened 2020-11-05
Fix GitHub Actions

### LINKED ISSUE #1185: Fix GitHub Actions
Currently the GitHub actions are broken.

lucyllewy:
I've tested these actually work on my fork, so these should be good to go.

fixes #1185

### DIFF
--- a/.github/workflows/snapcraftbuild.yml
+++ b/.github/workflows/snapcraftbuild.yml
@@ -30,7 +30,7 @@ jobs:
         snap: ${{ steps.build.outputs.snap }}
         isClassic: 'false'
     - name: Publish unstable builds to Edge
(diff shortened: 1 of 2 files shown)


diddledan:
fixes #1185

pierotofy:
Fantastic! Thanks!

Saijin-Naib:
Will the WSL images be rebuilt when the SNAP packages get built/updated with each PR merged?

lucyllewy:
They should be built in parallel - there's a separate workflow for the docker/wsl build and the snapcraft build

both workflows are set to build on new tags being created for releases, and the snapcraft build has an extra configuration to build every commit added to master and put that in the store's `edge` channel.

PR #1192, opened 2020-11-06
Update README.md for WSL instructions

lucyllewy:
* Add details on how to import the `rootfs.tar.gz` file into WSL
* Add instructions to update the WSL rootfs
* Add uninstallation instructions


pierotofy:
Awesome, very simple to get going with WSL.

smathermather:
`rootfs.tar.gz` doesn't show  up in the releases yet. Should it?

lucyllewy:
The Docker/WSL build is currently failing while installing OpenCV. I think this is a race-condition because building the Docker image locally works fine, but it seems something about the GitHub Actions environment is causing it to fail. The message indicates that OpenCV is trying to install (a) file(s) into `SuperBuild/install/bin` but cannot because the directory doesn't exist.

Once that's fixed, or worked-around, the rootfs.tar.gz and Docker image will both be generated when a new tag is created.

smathermather:
Gotcha. Makes sense.

PR #1194, opened 2020-11-07
Update snapcraft.yaml for core20/snapcraft-preload

lucyllewy:
It seems that snapcraft-preload[1] installs to a different location when
building with core20 vs core18. This meant the snap builds fail with
being unable to find the executable. This commit changes the reference
to the preload script in the `snapcraft.yaml` to the core20-built
location.

[1] https://github.com/sergiusens/snapcraft-preload


pierotofy:
:+1: thanks!!

PR #1195, opened 2020-11-07
Update snapcraft.yaml for missing webp library

lucyllewy:
Snapcraft complains about missing `libwebpdemux2` when building. This
commit adds it as a runtime dependency to fix the warning.


pierotofy:
Thanks! :clinking_glasses:

PR #1197, opened 2020-11-07
Update GitHub Actions - Workflow_Dispatch

Saijin-Naib:
Tweak the GitHub Actions to enable workflow_dispatch, so manual runs of the GitHub Actions can be triggered at any point upon request by the repository owner or other authorized contributors.

(I'd like to test the WSL image ;) )


pierotofy:
Mm, the thing about workflow dispatches is that they are meant for actions that don't cause side effects.

Some of those actions push to docker hub; I think if there's an interest in testing the WSL workflow you should make a fork, change a few paths if needed in the workflow files, and test from your own your fork. :+1:

Saijin-Naib:
Alright, I'll see if I can figure that out.

So you don't see any value to having the option to force-run an action with the dispatch from your end, or am I not understanding what they're meant for (which is entirely possible)?

pierotofy:
I'm just unfamiliar with what I'm enabling and I don't know if it's going to cause our docker repo to be updated without notice. We have a lot of moving parts, a trigger without notice at the wrong time could cause a lot of broken ODM installations for a lot of users! :)

Saijin-Naib:
If you don't know, I surely don't either 🤣

I also apparently enabled dependabot so... 🤷‍♂️

Probably best to not merge. I might ping Dan to see if he'd be so kind as to advise.

Saijin-Naib:
Closing, because apparently my clueless poking about in my fork is kicking off builds here, using up your build time.

Blegh.

PR #1198, opened 2020-11-08
Fix Buildtime Race Condition

lucyllewy:
The `mv` in `SuperBuild` for `lasmerge` will copy `lasmerge` to a file
called `SuperBuild/install/bin` **not**
`SuperBuild/install/bin/lasmerge` when the `bin` folder has not been
created before the install step is reached. This is the case when building with 2 jobs (`-j2`) so builds
fail when moving onto OpenCV which requires that `bin` be a directory.

* Change `mv` to more reliable `install` command for `lasmerge`.


pierotofy:
Nice catch! I hadn't seen this PR, but encountered the same issue today :+1: 

Thanks!

PR #1202, opened 2020-11-13
Update to add some missing python packages in Snap

lucyllewy:
I think this was caused by the release of Snapcraft 4.4 recently which
might have different behaviour. The snap is missing a few python
packages/modules now, which were present when building earlier.

* Add missing Python packages
* Ensure libblas and liblapack can be found with an LD_LIBRARY_PATH
  override.
* Rename reference to `python3.6` to `python3.8`.

Ref: https://community.opendronemap.org/t/issue-with-snap/5924

### DIFF
--- a/snap/local/usr/lib/python3.8/sitecustomize.py
+++ b/snap/local/usr/lib/python3.8/sitecustomize.py
@@ -15,7 +15,7 @@ else:
 
 for d in site_directories:
     if d:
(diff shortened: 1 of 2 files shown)


pierotofy (approved):

pierotofy:
Excellent as always. Thanks!

PR #1203, opened 2020-11-13
Disable attempts to build Snap for i386

lucyllewy:
Ubuntu 20.04 Does not support i386, so attempting to build the Snap for
i386 will either fail or be non-runnable on an i386 Linux distro.

* Disable i386 build in `snapcraftbuild.yml` GitHub Action workflow
* Disable i386 build when testing PRs for consistency


pierotofy:
Yep, we're OK not to support i386. :+1: 

Thanks!

PR #1205, opened 2020-11-16
Update build-container-and-wsl.yaml

Saijin-Naib:
Fix up changes to the workflow I accidentally up-streamed, and commented out lines I added, restoring original opendronemap/odm docker entries.

My branch remains in the comments for easy toggling on/off for testing as needed.


pierotofy:
Whoops! I missed it.

Thanks!

Saijin-Naib:
Yeah, me too 🤣

I guess the whole point of working in branches is to prevent this type of thing from happening.

I'll try to be more careful going forward.

PR #1210, opened 2020-11-24
Bug fixes, speed improvements and license change to AGPL

pierotofy:
This is a draft PR for the upcoming 2.3.0 release which will add bug fixes, improvements and change the license to AGPLv3. :tada: 

Will post a blog article in the upcoming weeks discussing the license change decision and how it will further help preserve the freedom of the project.


Saijin-Naib:
Would it not be faster/better to use VRTs of the tiles instead of merging them into a TIFF?

And if we're going to merge, would it not be safer to use BIGTIFF=YES and just save GDAL the possibility of guessing wrong?

Saijin-Naib:
I'm not sure if this is us or upstream, but have you seen this in your bugfixing hunt?
https://community.opendronemap.org/t/pdal-is-stuck-when-processing-large-data-set/6053/7?u=saijin_naib

kikislater:
What will be become NodeMicMac License ? Is there inherit from AGPL ?

pierotofy:
NodeMicMac will continue to be GPL, unless it gets updated with new code from either ODM/NodeODM, in which case it will also need to be switched to AGPL. The next release of NodeODM will be AGPL also.

pierotofy:
> Would it not be faster/better to use VRTs of the tiles instead of merging them into a TIFF?

Users expect a TIFF as output, so unfortunately we can't use just VRTs.

kikislater:
> unless it gets updated with new code from either ODM/NodeODM

I have it working ! I will push it with new license

Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN rm -rf \
   /code/SuperBuild/build/opencv \
   /code/SuperBuild/download \
   /code/SuperBuild/src/ceres \
-  /code/SuperBuild/src/entwine \
+  /code/SuperBuild/src/untwine \
```
smathermather, line 22: w00ooot!

Saijin-Naib: What does untwine give us vs entwine? It is entwine compatible from what I see, and HoBu, so good stuff, but I don't know enough about either library, apparently.

pierotofy: It's the same app behind QGIS's new point cloud stuff. Faster, more memory efficient for individual machines.

smathermather:
> I was having troubles reconstructing a dataset so I initially thought to just always use FLANN. I realized however that in another, BOW matching actually performed better.
> 
> I'm going to introduce a flag `--matcher-type` that lets users choose the matching algorithm, since sometimes one performs better than the other (and BOW is quite a bit faster).

Awesome. I can stop editing ODM to force BOW when I need to force it. :D

PR #1211, opened 2020-11-24
Double, Original, Half, Quarter, and Eighth for pc-quality & feature-quality

Saijin-Naib:
In fighting with this particularly difficult dataset (https://community.opendronemap.org/t/error-writing-ply-file/5683/22?u=saijin_naib) I got the idea that, in some instances, it might not only be desirable, but necessary to have --pc-quality and --feature-quality values that are unmodified from the input data.

To that end, these sets of PRs basically add "super" and "hyper" quality level flags (0.75, 0.50 respectively), and bump "ultra" up to 1.00.

I also normalized the arguments that --pc-quality and --feature-quality accepted (removed "lowest", moved quality values around to match one another), to ensure that cognitively, they're the same to use and quality levels represent the same relative quality level compared to the source dataset for both flags.

I don't believe these values are unprecedented (Ultra being quality of 1.00), as I believe Pix4D Mapper allows one to use "Original" quality for a few parameters in their pipeline.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -152,7 +152,7 @@ def config(argv=None, parser=None):
                         metavar='<string>',
                         action=StoreValue,
                         default='high',
-                        choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low'],
+                        choices=['double', 'original', 'half', 'quarter', 'eighth'],
                         help=('Set feature extraction quality. Higher quality generates better features, but requires more memory and takes longer. '
                             'Can be one of: %(choices)s. Default: '
                             '%(default)s'))
@@ -361,7 +361,7 @@ def config(argv=None, parser=None):
                     metavar='<string>',
                     action=StoreValue,
                     default='medium',
-                    choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low'],
+                    choices=['double', 'original', 'half', 'quarter', 'eighth'],
                     help=('Set point cloud quality. Higher quality generates better, denser point clouds, but requires more memory and takes longer. Each step up in quality increases processing time roughly by a factor of 4x.'
                         'Can be one of: %(choices)s. Default: '
                         '%(default)s'))
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -152,7 +152,7 @@ def config(argv=None, parser=None):
                         metavar='<string>',
                         action=StoreValue,
                         default='high',
-                        choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low', 'lowest'],
+                        choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low'],
                         help=('Set feature extraction quality. Higher quality generates better features, but requires more memory and takes longer. '
                             'Can be one of: %(choices)s. Default: '
                             '%(default)s'))
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -152,7 +152,7 @@ def config(argv=None, parser=None):
                         metavar='<string>',
                         action=StoreValue,
                         default='high',
-                        choices=['ultra', 'high', 'medium', 'low', 'lowest'],
+                        choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low', 'lowest'],
                         help=('Set feature extraction quality. Higher quality generates better features, but requires more memory and takes longer. '
                             'Can be one of: %(choices)s. Default: '
                             '%(default)s'))
@@ -361,7 +361,7 @@ def config(argv=None, parser=None):
                     metavar='<string>',
                     action=StoreValue,
                     default='medium',
-                    choices=['ultra', 'high', 'medium', 'low'],
+                    choices=['ultra', 'super', 'hyper', 'high', 'medium', 'low'],
                     help=('Set point cloud quality. Higher quality generates better, denser point clouds, but requires more memory and takes longer. Each step up in quality increases processing time roughly by a factor of 4x.'
                         'Can be one of: %(choices)s. Default: '
                         '%(default)s'))
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -154,11 +154,12 @@ class OSFMContext:
                 feature_process_size = int(args.resize_to)
             else:
                 feature_quality_scale = {
-                    'ultra': 1,
-                    'high': 0.5,
-                    'medium': 0.25,
-                    'low': 0.125,
-                    'lowest': 0.0675,
+                    'ultra': 1.00,
+                    'super': 0.75,
+                    'hyper': 0.50,
+                    'high': 0.25,
+                    'medium': 0.125,
+                    'low': 0.0675,
                 }
 
                 max_dim = find_largest_photo_dim(photos)
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -154,12 +154,11 @@ class OSFMContext:
                 feature_process_size = int(args.resize_to)
             else:
                 feature_quality_scale = {
-                    'ultra': 1.00,
-                    'super': 0.75,
-                    'hyper': 0.50,
-                    'high': 0.25,
-                    'medium': 0.125,
-                    'low': 0.0675,
+                    'double': 2.00,
+                    'original': 1.00,
+                    'half': 0.50,
+                    'quarter': 0.25,
+                    'eighth': 0.125,
                 }
 
                 max_dim = find_largest_photo_dim(photos)
--- a/opendm/utils.py
+++ b/opendm/utils.py
@@ -10,12 +10,11 @@ def get_depthmap_resolution(args, photos):
         max_dim = find_largest_photo_dim(photos)
 
         pc_quality_scale = {
-            'ultra': 1.00,
-            'super': 0.75,
-            'hyper': 0.50,
-            'high': 0.25,
-            'medium': 0.125,
-            'low': 0.0675
+            'double': 2.00,
+            'original': 1.00,
+            'half': 0.50,
+            'quarter': 0.25,
+            'eighth': 0.125,
         }
 
         if max_dim > 0:
--- a/opendm/utils.py
+++ b/opendm/utils.py
@@ -10,7 +10,9 @@ def get_depthmap_resolution(args, photos):
         max_dim = find_largest_photo_dim(photos)
 
         pc_quality_scale = {
-            'ultra': 0.5,
+            'ultra': 1.00,
+            'super': 0.75,
+            'hyper': 0.50,
             'high': 0.25,
             'medium': 0.125,
             'low': 0.0675


pierotofy:
We can certainly add more settings to the two options, just not sure about the choice of wording:

```
'ultra', 'super', 'hyper'
```

They sound a bit similar?

Was there a significant improvement in results with `--feature-quality`? Adding more options could confuse users, so we should justify the addition.

On the  full resolution option for depthmap computation, I have the same question; did it improve quality overall, or did you just get more points? Perhaps @smathermather  can chime in on whether we should add the option to process depthmaps at full resolution (is it beneficial?)

`original` would be a good term for processing at full resolution.

smathermather:
> 'ultra', 'super', 'hyper'

Before the labor exploitation model used by rideshare companies, I would have advocated for a more germanic version of `super`, but alas, the implications have changed.

In all seriousness, historically the ceiling for effective depthmaps has been 1/2 linear resolution (1/4 effective resolution). This is due to the noise ceiling that we see with most approaches. That said, OpenMVS is dang good at what it does, and may have some unbenownst-to-me way of avoiding the noise ceiling.

So... maybe a manual test of these proposed settings to make sure we aren't just going up to 11, and I'll leave the naming conventions to someone more accustomed to _Latinx_ superlative conventions.

Saijin-Naib:
Yeah, I did a bit of digging and there really isn't a strong consensus on what superlatives go in what order, but I did see more than a few where ultra succeeded hyper and super.

To be frank, I was thinking something easier and more clear would be nice, but I wasn't sure if changing the entire naming structure would be out of scope or not. Original, Half, Quarter, and something for the intermediaries would I think make sense, and should be easier to translate across languages.

As for the reconstruction, I think it did make a difference in the dataset I was struggling against, though in the end, it didn't solve the fundamental issues with the dataset that I was having. It did strongly mitigate their impact, though. The panels' shape did converge more on being regular rectangles, with what seemed like a decent reduction in variation of surface height on the panels. Severity of the fracture was reduced, and smaller features were better represented.

If you poke at the link, you should see the improvement from the run where I'm using the new Ultra (original?) settings for the dataset.

It was MEAN on RAM though.

smathermather:
Ah yes: my apologies. I have been following that thread. Give me settings and I'll run them as a test on an additional test dataset at those levels. My brain is fried from giving a workshop, so hand holding may be necessary.

Saijin-Naib:
> Ah yes: my apologies. I have been following that thread. Give me settings and I'll run them as a test on an additional test dataset at those levels. My brain is fried from giving a workshop, so hand holding may be necessary.

No worries.

I ran using these, but I imagine not all flags need be applied to other datasets:  
./run.sh --camera-lens brown --crop 0 --debug --dem-resolution 1.0
--dsm --feature-quality ultra --ignore-gsd --mesh-octree-depth 12 --mesh-size 1000000 --min-num-features 32768 --opensfm
-depthmap-method BRUTE_FORCE --orthophoto-resolution 1.0 --pc-classify --pc-quality ultra --pc-rectify --rerun-all --tim
e --use-3dmesh --use-hybrid-bundle-adjustment --verbose

Are you able to git pull https://github.com/Saijin-Naib/ODM Saijin-Naib-super_hyper into your /code/ to run on my fork?

As an aside, I'm running the brighton dataset through my fork right now and will put results here for you both to evaluate shortly.

Saijin-Naib:
`./run.sh --crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality ultra --ignore-gsd --mesh-octree-depth 12 --opensfm-depthmap-method BRUTE_FORCE --orthophoto-resolution 1.0 --pc-classify --pc-quality ultra --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose`

Filterpoints:  
image
Dang. That's dense.

DSM:  
image
I'm seeing what looks like strong detail without excess noise.

Composite Hillshade/Color Ramp/Contours:  
image

Orthophoto:  
image

ODM_Meshing:  
image

smathermather:
That's beautiful. What does the hillshade on the DSM look like?

smathermather:
Ye gawds that is beautiful. Name some variables, cut, print. This is fantastic.

Saijin-Naib:
> 
> 
> Ye gawds that is beautiful. Name some variables, cut, print. This is fantastic.

I don't think this would have been tenable prior to the improved filtering that was merged a little bit ago.

It might be wishful thinking, but I believe I'm actually seeing _real_ detail, not just finer point pitch/noise.

Given that my machine is a bit... meh, can I ask for assistance in running more tests/validation?

Saijin-Naib:
For naming schema:  
Does this work acceptably? This could also allow for Higher, Highest, as well as Medium-High, Medium-Low, if needed.

| Argument | Value | 
| --- | --- |
| Original | 1.00 | 
| High | 0.75 | 
| Medium | 0.50 | 
| Low | 0.25 |
| Lower | 0.125 |
| Lowest | 0.0675 |

With full intermediates:  

| Argument | Value | 
| --- | --- |
| Original | 1.00 | 
| Highest | 0.875 |
| Higher | 0.8175 |
| High | 0.75 | 
| Medium-High | .625 | 
| Medium | 0.50 | 
| Medium-Low | 0.375 |
| Low | 0.25 |
| Lower | 0.125 |
| Lowest | 0.0675 |

To me this looks kinda busy, maybe too granular, and the scaling isn't consistent between bins...

smathermather:
`--opensfm-depthmap-method BRUTE_FORCE` I am confused by this parameter being set. Doesn't this apply only to OpenSfM derived point clouds?

Saijin-Naib:
> `--opensfm-depthmap-method BRUTE_FORCE` I am confused by this parameter being set. Doesn't this apply only to OpenSfM derived point clouds?

You know what... You're probably right. That might just be a brainfog moment... When I was digging through the ODM Book to make a processing parameter set it was like 3am. 

You can likely safely leave that out.

pierotofy:
Just in case this wasn't known: we still have `--resize-to` and `--depthmap-resolution` parameters which allow a user to directly specify the exact size at which to process features and depthmaps. So we might not need much granularity. I think it should be limited to `4-5` options (which are already a lot for users to digest).

Saijin-Naib:
Mmm, good point. I'll readjust!

Saijin-Naib:
Trying out this locally, which aligns with an industry product:  
| Argument | Value | 
| --- | --- |
| Double | 2.00 |
| Original | 1.00 | 
| Half| 0.50 | 
| Quarter | 0.25 |
| Eighth | 0.125 |

I'm also testing their demo dataset with "double" quality, and defaults everywhere else, against my fork of ODM with "double" quality, and defaults everywhere else.

Saijin-Naib:
Double-Quality Testing:  

I don't have a license, so I can't export products, so the below are the Preview-Quality assets that get generated along with the Report. I know it isn't exactly 1:1, but it's the best I can do at present.

Commercial Preview-Quality DSM:  
dsm_preview

Commercial Preview-Quality Ortho:  
orthomosaic_preview

---

`./run.sh --crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality double --ignore-gsd --mesh-octree-depth 12 --mesh-size 10000000 --orthophoto-resolution 1.0 --pc-classify --pc-quality double --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose`

ODM DSM+Hillshade:  
image

ODM Ortho:  
image

Saijin-Naib:
`./run.sh --crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality double --ignore-gsd --mesh-octree-depth 12 --mesh-size 10000000 --orthophoto-resolution 1.0 --pc-classify --pc-quality double --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose`

Brighton ODM (Double):  
image

Brighton Ortho (Double):  
image

Saijin-Naib:
`--crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality ultra --gps-accuracy 5 --mesh-octree-depth 12 --mesh-size 400000 --min-num-features 160000 --opensfm-depthmap-method BRUTE_FORCE --orthophoto-resolution 1.0 --pc-classify --pc-quality ultra --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose `

Took a few days since it was swapping out to a WD Green in a USB enclosure, but it finished! 🤣 Please, gib RAM

mf_OBriens DSM:  
image

mf_OBriens Ortho:  
image

kikislater:
> swapping out to a WD Green in a USB enclosure

^^ You will break it ! Swapping on SSD is way forward but not convenient as well

Saijin-Naib:
> 
> ^^ You will break it ! Swapping on SSD is way forward but not convenient as well

I know! Unfortunately, no choice. It is what I have... Feel like buying me more RAM? 🤣

As for this PR, I think I have it where I'd like it for now. I'm satisfied with the results, and I think the newer naming schema/bins are logical enough, while having the added benefit of matching an industry product, so we get to tap into existing familiarity and expectations.

I do want to do further similar work for something like a --mesh-quality flag, but that will take me a while to figure out since it involves multiple parameters, so I want to do that on a separate PR.

Anyone have any thoughts/comments/reviews/suggestions?

Like, comment, subscribe, and smash that Merge button.

pierotofy:
This is great testing, but I'm not sure we should implement all changes proposed here. :pray: 

`quality` is a term that indicates `character with respect to fineness, or grade of excellence`. So `high`, `medium`, `low` and `ultra` are perhaps OK names for such property. `quarter`, `half`, etc. are numeric quantities. I'm not so sure we should change them (I'd love a third or fourth opinion here..)

Although there's a desire to standarize ratios between `pc-quality` and `feature-quality`, two are not really the same and perhaps should not use the same ratios. Also we can't unfortunately use double resolution for feature quality, because OpenSfM won't let it: https://github.com/mapillary/OpenSfM/blob/767756c2c2f35a2d081a03dffdaa0ba21400a89e/opensfm/features.py#L15)

Now to the good part, which is the part which I think we should absolutely include, is the ability to process depthmaps at full resolution (1), which is currently lacking, perhaps by using these ratios:

```
Ultra = full size
High = Downscaled to 50%
Medium = Downscaled to 25%
Low = Downscaled to 12,5%
Lowest = Downscaled to 6,75%
```

Saijin-Naib:
I'd argue that there is a strong precedent for using numeric ratios to refer to quality versus input, ie, Double, Original, Half, Quarter, Eighth. I think that it is fairly intuitive (and self-explanatory without having to dive into documentation/source to see exactly what we think "high" means), and further, shouldn't have the ambiguity of superlatives, and should also translate well across languages where superlatives might not have exact analogues.

I could swear that I'm getting different results with feature-quality Double versus Original. I'll reprocess Brighton with both and attach results.

Original Feature Quality:  
`./run.sh --crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality original --ignore-gsd --mesh-octree-depth 12 --mesh-size 10000000 --orthophoto-resolution 1.0 --pc-classify --pc-quality original --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose`

Double Feature Quality:  
`./run.sh --crop 0 --debug --dem-resolution 1.0 --dsm --feature-quality double --ignore-gsd --mesh-octree-depth 12 --mesh-size 10000000 --orthophoto-resolution 1.0 --pc-classify --pc-quality original --pc-rectify --rerun-all --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose`

OneDrive Link:  
https://1drv.ms/u/s!AvMZEGXuAwQzv74uqwt3wl-GsqSc9g

Is feature extraction and tie-point generation deterministic, or is it variable between runs? There are differences in the sizes of the datasets between double and original for feature_quality, but if they're not deterministic, then it isn't a real difference.

CarlosGrohmann:
Hello, 

I just found this discussion. If I may add an opinion, I think it is very interesting to have these "numeric" settings. Just for comparison, Metashape (Photoscan) uses the following:

Ultra High = double of the original resolution
High = original resolution 
Medium = half of resolution
Low = quarter of resolution
Lowest = eigth of resolution

I never really saw much need for doubling the resolution of the photos, as it generates lots of noise and takes soooo long to process. 

But a more direct indication of what the options mean is very welcome. And having the option to use the original resolution for both feature-quality and pc-quality is great.

Saijin-Naib:
Thanks for contributing!

Here's where I'm at currently:
I want to test Double for feature-quality, but I need to tweak OpenSFM to enable it. This likely will take me a bit to figure out.

As you've added, other industry products tend to align on how they treat "Quality Levels", and aside from familiartiy, I think it is a really intuitive approach to communicating Quality Levels.

I've found that Double for image-quality aligns with Pix4D, and at least Agisoft in addition, as you've noted. So even if there are potential downsides to using them, keeping workflows as compatible as possible with other industry products I think is of great value, especially if we might one day be able to import project files from these other suites to process within ODM (I want to accomplish this, but I doubt my technical ability to do so in the near-term).

I'm really interested in the experience you had with the Double Resolution (in Agisoft, I assume). Most methods of super-sampling and then sub-sampling should reduce noise, not increase it. Do you have an example?

CarlosGrohmann:
@Saijin-Naib I don't have the results with me, but one of my PhD students was looking into this. I'll check with her if we can send you some examples.

Saijin-Naib:
> @Saijin-Naib I don't have the results with me, but one of my PhD students was looking into this. I'll check with her if we can send you some examples.

@CarlosGrohmann , just a ping to see if you have that data for me to poke at!

Saijin-Naib:
@CarlosGrohmann ,

Hope you're well. Do you think you'd be able to share that data in the coming weeks?

CarlosGrohmann:
> @CarlosGrohmann ,
> 
> Hope you're well. Do you think you'd be able to share that data in the coming weeks?

Hello @Saijin-Naib, sorry for the late reply. 

I asked my student about that and I'm afraid we don't have much to share. Maybe my memory was tricking me, as she told she didn't really looked too deeply into this. According to her, in Metashape's forum, 'they' say to only use the 'highest' option in very sharp images and for research purposes (and the only in the alignment step it will double the resolution with that option). She tried with a few datasets and couldn't see much difference in the results, and in one case the 'highest' resulted in less points than the 'high' option. That's all she told me, sorry again if it's not much.

Saijin-Naib:
Thanks for getting back to me!

pierotofy:
Can this be closed?

Saijin-Naib:
I would like to revisit it, but I guess fundamentally it is likely still too heavy.

What is your recommendation?

pierotofy:
Perhaps opening another PR with a smaller scope (e.g. tackle one parameter at a time)?

PR #1212, opened 2020-11-25
Confusing sentence

kikislater:
Sentence is confusing


Saijin-Naib:
Should the SNAP image be mentioned as well?

pierotofy:
Nice catch @kikislater :+1:

PR #1215, opened 2020-12-08
Re-run bug fix, --pc-quality changes

pierotofy:
@Saijin-Naib do you think this could be a good trade-off for the PR opened in #1211 ? It doesn't change the terminology, but bumps `ultra` to `1` and adds a `lowest`. Also sets a minimum cap (depthmaps under 320 are not very usable).


Saijin-Naib:
@pierotofy,

Given the other important things you've included, sure, I'd say merge this immediately.

I can always rebase my other PR against these changes, since I still think changing the names and adding a Double have value.

PR #1217, opened 2020-12-10
Entwine/untwine, life's better together

pierotofy:
Entwine tends to be a bit faster (for now) on multi-core machines.

But untwine does a much better job at handling memory.

So we try both! Best of both worlds.


smathermather:
Aha. I got on here to comment that for some reason entwine is still in the build process. Now I know why...

PR #1230, opened 2021-02-05
Add --split-image-groups parameter

pierotofy:
Adds support for specifying a image group file path instead of relying on the `image_groups.txt` convention. This can be useful for command line users wishing to try different group files. (Works like `--gcp` and `--geo`).


smathermather:
w000t!

PR #1236, opened 2021-03-01
Update OpenMVS to latest

pierotofy:
- [x] Updates OpenMVS to latest version
- [x] Updates OpenSfM export to work with latest OpenMVS
- [x] Adds code to use OpenMVS ability to split dense map fusion into tiles to keep memory usage in check


Saijin-Naib:
> * [x]  Updates OpenMVS to latest version
> * [x]  Updates OpenSfM export to work with latest OpenMVS
> * [x]  Adds code to use OpenMVS ability to split dense map fusion into tiles to keep memory usage in check

The return of Double/Original as a possibility?!

israelbar:
Will the OpenMVS ability to split dense map fusion into tiles reduce the "exited with code 1" errors?

Saijin-Naib:
That's my understanding. If so, I am pretty hyped.

smathermather:
> Will the OpenMVS ability to split dense map fusion into tiles reduce the "exited with code 1" errors?

Yup.

pierotofy:
Ready for testing. Scenes are split if they exceed a certain area (so this won't kick in if the scene is really small). It's a bit tricky to choose what a good default area is. I've set it to a good initial default and we can tweak it later on as soon as reports of problems start to arrive from users.

I've witnessed some problems of chunk of areas being missed when the sub-scene-area was set to a very small value (for testing purposes), but I think that's because the area needs to be sufficiently large to have enough overlap (similar to having problems with split-merge if split-overlap is set to a too low value). 

Anyway, it would be good to test this on as many datasets as possible and compare with the global reconstruction approach to make sure chunks are not being missed.

Saijin-Naib:
@pierotofy , to be clear, the tiling is enabled by default, no special flags needed to pass it right?

pierotofy:
Correct!

Saijin-Naib:
First blush, seems much snappier. Memory usage gets a bit scary on Brighton Beach with the below settings, but it doesn't Code 1, and overall usage seems to be much more controlled.
Products seem good!

`./run.sh --crop 0 --debug --dem-resolution 0.1 --dsm --feature-quality ultra --mesh-octree-depth 14 --pc-classify --pc-quality ultra --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose
`
[report.pdf](https://github.com/OpenDroneMap/ODM/files/6079235/report.pdf)
image
image

Saijin-Naib:
Living Room needed some concurrency limiting to not Code 1 during SIFT feature extraction (which seems to be the heaviest part by far now):  
`./run.sh --crop 0 --debug --dem-resolution 0.1 --dsm --feature-quality ultra --mesh-octree-depth 14 --pc-classify --pc-quality ultra --time --use-3dmesh --use-hybrid-bundle-adjustment --verbose --rerun-all --max-concurrency 4`

Code 1 during odm_meshing:  
```
Vertices / Polygons: 42286498 / 82819133
Corners / Vertices / Edges / Surface / Set Table / Copy Finer: 19.0 / 19.2 / 30.6 / 23.7 / 25.7 / 7.2 (s)
#        Got triangles:     138.3 (s),    9787.1 (MB) /   10927.3 (MB) /       0.0 (MB)
#          Total Solve:     631.0 (s),   10927.3 (MB)
[INFO]    running /code/build/bin/odm_cleanmesh -inputFile /code/odm_meshing/odm_mesh.dirty.ply -outputFile /code/odm_meshing/odm_mesh.ply -removeIslands -decimateMesh 200000 -verbose
Running with parameters:
        --inputFile /code/odm_meshing/odm_mesh.dirty.ply
        --outputFile /code/odm_meshing/odm_mesh.ply
        --decimateMesh 200000
        --removeIslands
        --verbose
Removing islands
Decimating mesh
Current vertex count: 42286498
Wanted vertex count: 200000
Target reduction set to 0.995270

[process exited with code 1]

```

pierotofy:
Mm, I'm getting some strange results. :)

Perhaps I messed up something. This will require further investigation before merging.

Related: https://github.com/cdcseacave/openMVS/issues/655

stages/openmvs.py
```diff
@@ -78,7 +79,7 @@ def process(self, args, outputs):
 
             log.ODM_INFO("Computing sub-scenes")
             config = [
-                "--sub-scene-area 10000",
+                "--sub-scene-area 660000",
```
smathermather: 16 gigs feels, while not unreasonable, not as accessible as 8GB memory ceiling. We get so many folks on the community forum showing up with 8GB on their hardware. Is there large enough a disadvantage to trigger a split at a smaller sub-scene size?

pierotofy: Not really, we can probably lower it. 16GB is an approximation, I'm still trying to figure out exactly how that parameter affects the splits, it's a bit non-intuitive (so not sure by what unit I have to scale it to half memory requirements). Accepting better value suggestions.

Saijin-Naib: To further the 8GB recommendation, WSL uses half the physical memory by default as the max it will allocate to the WSL2 instance... With many machines shipping with 16GB as premium, that means that 8GB is a more comfortable lower limit. It would also make processing on a single Raspberry Pi 8GB tenable, which is cool, if impractical if not clustered.

pierotofy:
I'm also a bit worried about lowering the area too much; I halved `660000` and got 7 sub-scenes instead of 2, but there's definitely a deviation from the global reconstruction where tiling becomes apparent:

image

image

Saijin-Naib:
What's happening during the tiled reconstruction that so many points are being consumed?

pierotofy:
It's a bit like split-merge; a scene is divided into multiple sub-scenes, with some overlap. If the overlap / scene size is too small --> less cameras --> less points.

pierotofy:
I'm tempted to add a flag to disable/enable tiling. I just can't predict if this is going to break the reconstruction for some datasets.

Saijin-Naib:
> I'm tempted to add a flag to disable/enable tiling. I just can't predict if this is going to break the reconstruction for some datasets.

Flag with tiling off for default to not change behavior for the majority until it gets more stable/predictable seems like a good recourse.

pierotofy:
Yep, that's what I'm leaning toward. Mark it as experimental and ship.

smathermather:
Alternatively, if it works consistently at the 16GB threshold, then set it there and ship. Then it's not another flag, fixes things for lots of users, but doesn't introduce artifacts. It's not great news for folks with smaller RAM footprints, but even my ultra portable 2015 laptop has 8GB of RAM.

pierotofy:
I just managed to crash it with the banana dataset.

```
18:32:08 [App     ] Command line: /datasets/banana/opensfm/undistorted/openmvs/scene_0001.mvs --resolution-level 1 --min-resolution 1504 --max-resolution 3008 --dense-config-file /datasets/banana/opensfm/undistorted/openmvs/config.ini --number-views-fuse 2 --max-threads 36 -w /datasets/banana/opensfm/undistorted/openmvs/depthmaps -v 0
18:32:08 [App     ] Preparing images for dense reconstruction completed: 4 images (216ms)
18:32:08 [App     ] Selecting images for dense reconstruction completed: 1 images (2ms)
Fused depth-maps 1 (100%, 110ms)             
18:32:08 [App     ] Densifying point-cloud completed: 0 points (360ms)
18:32:08 [App     ] MEMORYINFO: {
18:32:08 [App     ] 	VmPeak:	 3349376 kB
18:32:08 [App     ] 	VmSize:	 3335612 kB
18:32:08 [App     ] } ENDINFO
[INFO]    running /code/SuperBuild/install/bin/OpenMVS/DensifyPointCloud "/datasets/banana/opensfm/undistorted/openmvs/scene_0001_dense.mvs" --filter-point-cloud -1 -v 0
18:32:08 [App     ] Build date: Mar  4 2021, 15:34:50
18:32:08 [App     ] CPU: Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz (36 cores)
18:32:08 [App     ] RAM: 68.59GB Physical Memory 32.00GB Virtual Memory
18:32:08 [App     ] OS: Linux 5.4.0-1038-aws (x86_64)
18:32:08 [App     ] SSE & AVX compatible CPU & OS detected
18:32:08 [App     ] Command line: /datasets/banana/opensfm/undistorted/openmvs/scene_0001_dense.mvs --filter-point-cloud -1 -v 0
18:32:08 [App     ] error: empty initial point-cloud
  File "/code/run.py", line 68, in <module>
    app.execute()
  File "/code/stages/odm_app.py", line 81, in execute
    self.first_stage.run()
  File "/code/opendm/types.py", line 338, in run
    self.next_stage.run(outputs)
  File "/code/opendm/types.py", line 338, in run
    self.next_stage.run(outputs)
  File "/code/opendm/types.py", line 338, in run
    self.next_stage.run(outputs)
  [Previous line repeated 1 more time]
  File "/code/opendm/types.py", line 319, in run
    self.process(self.args, outputs)
  File "/code/stages/openmvs.py", line 124, in process
    system.run('%s "%s" --filter-point-cloud -1 -v 0' % (context.omvs_densify_path, scene_dense_mvs))
  File "/code/opendm/system.py", line 79, in run
Exception: Child returned 1
```

smathermather:
image

pierotofy:
image

pierotofy:
I'm adding a `--pc-tile` flag. I think the tiling approach works mostly well for most datasets, but there's always some odd one that can cause trouble. More testing is needed and I think the best way to get this done at scale is to allow the community to start testing it via a flag. Once we can identify / fix those issues, we'll remove the flag and tile by default.

The good news is that memory usage should already be improved on its own simply by upgrading OpenMVS.

smathermather:
Sounds good. Yeah: if it's not good for bananas, it's not ready for production.

I'll use the flag by default, so you know I'll break things (hopefully in useful and replicable ways).

Saijin-Naib:
> The good news is that memory usage should already be improved on its own simply by upgrading OpenMVS.

Yeah, I didn't OOM with the Living Room dataset, and that was without tiling... Wow.

YanNoun:
@pierotofy you might want to play with these two magic number ;
https://github.com/cdcseacave/openMVS/commit/7a88fefb0476fa4f6229e228a69641ff3febeb10#diff-66d736c1d38c1b27ace4b48ffbf5a91639659add7b752ea4fb7070b347b1a01cR866
and
https://github.com/cdcseacave/openMVS/commit/7a88fefb0476fa4f6229e228a69641ff3febeb10#diff-66d736c1d38c1b27ace4b48ffbf5a91639659add7b752ea4fb7070b347b1a01cR752.
Maybe putting the first down to 0.05 and the latter up to 0.05 might help

pierotofy:
Thanks for the tip @YanNoun !

PR #1239, opened 2021-03-05
GPU Support

pierotofy:
- [x] Adds support for GPU SIFT (OpenCL based)

There's also the option to enable GPU SIFT matching, however in my benchmarks it seems to underperform the CPU based FLANN, so I'm not going to enable it. It can be enabled by adding `matcher_type: SIFT_GPU` to the `config.yaml` (via `osfm.py`) if anyone wants to try.


Saijin-Naib:
First, let me just say:  
It'sHappening!GIF
This is beyond awesome.

Next, with my limited understanding, I came across POCL:
Would POCL maybe help with loading/detecting/depending the various GPU-driver ICDs?
https://community.opendronemap.org/t/gpu-programming-and-overview-assistance/5526/22?u=saijin_naib

It also would allow things to run with OpenCL acceleration even just on CPU.
In some cases, it can be better than the native OpenCL implementations:  
http://portablecl.org/pocl-1.6.html

pierotofy:
I'm not familiar with POCL, but we already rely on PyOpenCL for dealing with OpenCL.

pierotofy:
I was hoping to add mesh refinement (via `RefineMesh` from OpenMVS) in the mix, but I couldn't get it to work with CUDA and the CPU implementation runs out of memory pretty quickly on small datasets, so we're going to leave that for a future PR.

Saijin-Naib:
> I'm not familiar with POCL, but we already rely on PyOpenCL for dealing with OpenCL.

POCL, amazingly, is not PyOpenCL but yet another same-abbreviated module (Portable OpenCL). Very clear 🤣

I think POCL replaces all the vendor-specific OpenCL drivers.

pierotofy:
@smathermather if you want to play around with mesh refinement, one can simply issue:

```
/code/SuperBuild/install/bin/OpenMVS/RefineMesh --mesh-file /datasets/brighton/odm_meshing/odm_mesh.ply  -i /datasets/brighton/opensfm/undistorted/openmvs/scene_dense.mvs -o /datasets/brighton/odm_meshing/odm_mesh_refined.ply -v 2
```

If you stop the pipeline at meshing, run the refinement, then replace odm_mesh.ply with the refined mesh, you can resume the pipeline with the new refined mesh. It has yielded quite a bit of improvement in the details of the mesh, and if one has a pretty heavy machine, after a bit of time it yields some pretty good results.

smathermather:
Awesome! I did this a few weeks ago with a river dataset but haven't tried yet with an aerial dataset. I had lots of memory, so I didn't notice any issues 😀

pierotofy:
Some early benchmark numbers:

Dataset | https://github.com/pierotofy/drone_dataset_sheffield_cross/
-- | --
Images | 173
  |  
Extract Features (CPU 4 cores) | 2m50.908s
Extract Features (Intel HD Graphics 620 GPU) | 1m35.576s
Extract Features (CPU 32 cores) | 0m24.014s
Extract Features (GPU NVIDIA Tesla T4) | 0m31.527s
Colmap Extract Features (GPU NVIDIA Tesla T4) | 0m27.859s

The performance gain is noticeable on computers with few cores (most consumer devices), but the gap is closed as the number of cores scales up. Also ran COLMAP (which has an efficient implementation) for comparison.

pierotofy:
It would be interesting to use an hybrid approach (fill up CPU usage AND use the GPU when available), but I think the gains are small and the efforts are large. Other areas might be better candidates for speed optimization.

Saijin-Naib:
That hybrid approach does sound really promising, though. What would the stumbling block be? How to manage what work units get done by GPU/CPU?

pierotofy:
> What would the stumbling block be?  

Writing a consistent implementation of SIFT between CPU and GPU.

PR #1258, opened 2021-04-05
Show more friendly error logs

### LINKED ISSUE #1242: Better error messages reporting
We should try to output more friendly error messages throughout the pipeline. Most of the confusion in error messages arises from the failure of one program in the processing pipeline (executed via system.run), which prints a stack trace. 

Each system.run call should:

1. Check for obvious status codes (e.g. 137 out of memory) and provide a clear error message (oops! Download more ram! https://downloadmoreram.com/) 
2. Provide an additional description in case the error is a generic code 1. Most errors are due to bad input or bad parameters. It should be possible to provide a good description and possible things to try.

pierotofy:
Closes #1242


Saijin-Naib:
For the error message about the WSL2, it isn't entirely accurate, I don't think.

Using the .wslconfig you can tweak WSL2 to use as much RAM/SWAP as you want, beyond the default 50%.
https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configure-global-options-with-wslconfig

pierotofy:
Oh, I didn't know that. Feel free to open a PR to edit it?

PR #1260, opened 2021-04-05
Update odmapp.py for WSL2 wslconfig note

Saijin-Naib:
Add note citing https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configure-global-options-with-wslconfig for .wslconfig tweaking to help with OOM errors.


pierotofy:
Thanks @Saijin-Naib !

PR #1273, opened 2021-04-22
Avoid bug when `resize-to=-1` in WebODM lead to `feature_process_size=1` in OpenSFM

charnesp:
When `resize-to`is set to `-1` in WebODM (default case in WebODM), `resize_to_is_set`is equal to `-1`, thus  `feature_process_size`is set to `-1` and opensfm works at full image size. This patch avoid it.

### DIFF
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -157,7 +157,7 @@ class OSFMContext:
             # Compute feature_process_size
             feature_process_size = 2048 # default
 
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @charnesp ! :pray:

smathermather:
Ahh, that explains some crashes I was seeing at a step I wasn't expecting the affected. Thanks!

PR #1274, opened 2021-04-26
Update las formatr for merged point clouds to 1.4

smathermather:


pierotofy:
This might need to be updated on the lasmerge command instead of PDAL. I don't think that function is called anywhere if I remember correctly.

pierotofy:
Any updates on this one?

pierotofy:
I'll close this for now, feel free to reopen in the future :+1:

PR #1275, opened 2021-04-26
Removed VTK, odm_cleanmesh module

pierotofy:
- [x] Removes the dependency to VTK
- [x] Removes the `odm_cleanmesh` module in favor of OpenMVS's ReconstructMesh, which works similarly if not better.


pierotofy:
Good to go!

smathermather:
Woah!

PR #1281, opened 2021-05-07
Remove dependency on GRASS GIS

pierotofy:
- Started rewriting cutline computation algorithm
- Rewrote cutline computation algorithm
- Remove GRASS dependency


smathermather:
Woah!

PR #1310, opened 2021-06-23
Adapted format to PEP8 and fixed some PyCharm warnings about unused code

p-vitt:
When looking at the code with PyCharm, it found some issues with formatting. I went through all files and fixed the low hanging fruits. For some more involved warnings, I added TODOs to keep track of them.

### DIFF
--- a/opendm/camera.py
+++ b/opendm/camera.py
@@ -1,5 +1,7 @@
-import os, json
+import os
+import json
 from opendm import log
+
 
 def get_cameras_from_opensfm(reconstruction_file):
     """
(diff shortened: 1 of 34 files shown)


pierotofy:
Thanks for the PR @p-vitt ! 

Unfortunately, automated linters like the one in PyCharm are not safe to just "run and forget". For example, running your branch gives me a:

```
  File "/code/run.py", line 14, in <module>
    from opendm.utils import double_quote, get_processing_results_paths
ImportError: cannot import name 'double_quote' from 'opendm.utils' (/code/opendm/utils.py)
```

This is because PyCharm thought that an import statement wasn't used in the module (but it was in fact needed by a module that depends on the module).

I'm open to improve the codebase to improve readibility, but please do not rely on IDEs or automated tools to do so. Make small incremental changes (one module at a time). Having a large number of changes (34 files!) makes it really difficult to review and discover possible regressions. :pray:

p-vitt:
I guess, trying to fix some stuff while setting up test bench is not working. I should get my stuff running first... Sorry for bothering. 

However, just to understand the problem: Is it valid design to import something just that others can import it from another place? Is this a construct to publish something that originates to an internal module in a more reasonable way for a caller?

Additionally, I'm not used to this whole pull request workflow. What would be a reasonable way to publish those PEP8 changes to the 34 files? Have one branch with 34 commits so that one can bisect through all those commits? In another project I was asked to squash those commits together into one big commit to keep the history clean. No offense, please, I'm just unsure how to do this and need a little guidance to get started or used to this project's culture.

pierotofy:
No worries, and great questions.

>  Is it valid design to import something just that others can import it from another place?

Yes.

> Is this a construct to publish something that originates to an internal module in a more reasonable way for a caller?

I sometimes use this pattern to avoid breaking code in other modules (e.g. a function was in `utils` but was then moved elsewhere, I can move it and still import it from the old module). As to whether this is a good practice it's up for debate ^_^

 > What would be a reasonable way to publish those PEP8 changes to the 34 files?

Don't change the code just "because PEP8". Does it improve readability for you and others? Make small incremental changes (don't run the entire project via a linter and call it good). Perhaps target a handful of modules first. Always test every code path after making such changes. :pray: 

In short, keep PRs small.

PR #1320, opened 2021-08-03
Add 3D error estimation in report

pierotofy:
This PR adds 3D error (relative) estimation to the PDF report, `stats.json`.


pierotofy:
image

PR #1321, opened 2021-08-04
Fix EXIF error while using GCPs

pierotofy:
When using GCPs and passing --optimize-disk-space the program will fail with: FileNotFoundError: [Errno 2] No such file or directory: ‘/var/www/data/e852b60f-abdf-47fc-a95d-d967f81b510b/opensfm/exif/DJI_0567.JPG.exif’


pierotofy:
Reported on https://community.opendronemap.org/t/error-in-processing-gcp-or-exif/8489

PR #1322, opened 2021-08-05
3D Error Function Fix

pierotofy:
When partial reconstructions are reconstructed, the td_errors function sometimes fails because of missing keys (fixed in https://github.com/OpenDroneMap/OpenSfM/tree/261)


pierotofy:
As reported on https://community.opendronemap.org/t/very-simillar-topic-cannot-process-dataset/8493/6

PR #1330, opened 2021-08-16
Update photo.py

ckato3:


pierotofy:
Merging into the `multispec` branch, thanks @ckato3 !

PR #1333, opened 2021-08-23
Update parameters of cap_resolution().

magnus-eigenvision:
Add parameter `gsd_scaling` instead of using `gsd_error_estimate` in an unintuitive way.


pierotofy (approved):

pierotofy:
Thanks @magnus-eigenvision ! Looks great 👍

PR #1337, opened 2021-08-31
Enable ORB --feature-type

Saijin-Naib:
Enable ORB as an optional feature type.
I'd love to get more folks testing this.

Here's some promising literature:  
https://mikhail-kennerley.medium.com/a-comparison-of-sift-surf-and-orb-on-opencv-59119b9ec3d0

The literature seems to indicate that pixel drift and reprojection errors are lower with ORB, and looking at the Report.PDF it certainly seems to agree with that assessment.

ORB Report.PDF:  
[report.pdf](https://github.com/OpenDroneMap/ODM/files/7086903/report.pdf)

SIFT Report.PDF:  
[report.pdf](https://github.com/OpenDroneMap/ODM/files/7086905/report.pdf)

However, I'm noticing that ORB seems to introduce bowling with Brown-Conrady that previously didn't occur with SIFT. I can't help but think that this is the increased precision conflicting with Rolling Shutter artifacts and the fact that this dataset was collected at 25mph without stops.
(ORB pictured as bottom Point Cloud)  
image

In terms of processing speed, on brighton_beach it saved about 1 minute with "high" --feature-quality. On mf_obriens, it saved about 20min. Memory consumption during feature detection also seems to be way more controlled, even with increased --min-num-features to 16000 and --feature-quality ultra.
brighton_beach with --feature-quality ultra? 10m processing time delta, even with ORB doing a DSM+DTM, not just DTM.
image

From reading this paper, I'm thinking ORB might be able to help with built-environment (particularly solar panels) reconstruction, since it seems to really like corners/edges, but seems to perform about as well (if not better) as SURF/HAHOG in more natural scenes:  
https://ieeexplore.ieee.org/document/8346440


Saijin-Naib:
It looks like we may need to suggest/switch to a different matching type when using ORB feature descriptors:  
https://docs.opencv.org/master/dc/dc3/tutorial_py_matcher.html

Either Brute Force with NORM_HAMMING, or FLANN with some tweaks.

pierotofy:
Awesome, thanks!

Saijin-Naib:
> Awesome, thanks!

Any ideas on the bowling and matchers issues, or let it be an option and see how others get on with it first?

pierotofy:
I'd let it be an option and see how it goes; the literature seems to agree that ORB can outperform SIFT in certain cases, but not others. So it's scene dependent. It remains to see which cases are best served by which detector.

I'm not entirely sure about the matching.

PR #1338, opened 2021-09-02
M1 Changes

pierotofy:
Necessary changes to build ODM on M1.

Switches the docker images to Ubuntu 21.04, since arm ports for GDAL & co are not available in the UbuntuGIS PPA.


pierotofy:
It's a go! Merging...

PR #1350, opened 2021-09-21
Merge Preview module, cutline fix

pierotofy:
- Add quickpreview module, fix cutline polygons edge-case (when a single polygon is created).


pierotofy:
The quick preview module is a cool experiment that creates fast previews of near-nadir photos (somewhat similar to map2dfusion outputs).

Untitled

Sheffield Cross (above) was captured with non-nadir shots, so obviously didn't yield great results, but took just a few minutes to compute and might be useful to get an idea of the dataset before actual processing.

PR #1353, opened 2021-10-06
update flying best practices link

mikegreen:
link was broken


pierotofy:
Nice catch! Thanks @mikegreen :+1:

PR #1355, opened 2021-10-11
Feat: user-defined boundary

pierotofy:
Adds the ability to specify a boundary polygon that limits the area of the reconstruction.

Work in progress.﻿


pierotofy:
Closes #852

pierotofy:
I think I will also add the ability to automatically generate vertical bounds based on the camera EXIFs extents / GSD to avoid the problem of reconstructing background objects (e.g. sky, clouds, or anything in line of sight far away) using the same logic.

pierotofy:
`--boundary <geojson>` --> Polygon specifying the reconstruction area
`--auto-boundary` --> Automatically calculates the polygon based on the camera shots convex hull (plus a buffer) to avoid reconstructing artifacts far in the distance.

Example input:

image

Example outputs (with `--crop 0` to disable additional cropping):

image

image

image


image

PR #1358, opened 2021-10-14
calculate the correct shot origin for gsd estimation

originlake:
Hi all,

I noticed that the shot origin is not calculated correctly in the current GSD estimation. 

The translation is the position of the origin of the world coordinate system expressed in coordinates of the camera-centered coordinate system instead of the position of the camera. So I add a small function to obtain the correct camera position based on the implementation in OpenSFM.

### DIFF
--- a/opendm/gsd.py
+++ b/opendm/gsd.py
@@ -2,6 +2,7 @@ import os
 import json
 import numpy as np
 import math
+import cv2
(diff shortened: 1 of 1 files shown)


pierotofy:
Good catch! Thanks @originlake :pray:

PR #1367, opened 2021-11-04
Fixed a bad link

Rdassy:
Changed previous URL https://docs.opendronemap.org/arguments.html which gave a 404 to https://docs.opendronemap.org/arguments/


Rdassy:
I also checked the other README links just to be thorough, should be all good!

pierotofy:
Great catch, thanks @Rdassy  :pray:

PR #1394, opened 2022-01-03
fix:  adding os dependent code for rerun-all directory deletion

coumbsek:
Should fix the 1388 issue (https://github.com/OpenDroneMap/ODM/issues/1388) by introducing os dependent code.
Tested on windows but could not test the linux version.


pierotofy:
Thanks @coumbsek :pray: !

PR #1399, opened 2022-01-06
Update requirements.txt

Saijin-Naib:
Update (almost) all requirements to current/latest.

Exception is laspy with breaking changes in v2.x branch, but this should be looked into as this branch seems it will address some pain points we have with large point clouds, memory usage, speed, etc.

Tested by running native in Ubuntu 20.04.3 LTS WSL2, compiling from latest upstream source, and using venv to install requirements.txt to then run.

Tested against brighton_beach data with full defaults from ./run.sh


pierotofy:
It seems like this is not building correctly (see the build details for exact errors).

In general however, we don't want to update dependencies and risking something to break (and stuff does break in Python quite frequently when you update packages), unless we have a reason to update. (Are there performance improvements? New features?)

Saijin-Naib:
A number of these are tagged with CVEs, which was my main motivation for looking at updating the various dependencies to begin with.

Yeah, I see the failing builders, but I'm confused that debian/local passes but others don't. I'll take a closer look and see if I can figure out what's happening there.

Many of the libraries are almost entirely bugfixes. Heavy stuff like Pillow looks like new features, formats/bitdepths, bugfixes, and a few optimizations, but I don't have benchmark data to tease out how meaningfully it impacts things for processing.

GDAL looks to have significant performance improvements, especially with Raterio in the 3.3.x and above tags, as well as some fixes to built-in modules and drivers.

Saijin-Naib:
Looks like this may fix the Windows builder for x86_64:  
https://www.ultraengine.com/community/topic/19423-component-too-big-for-incremental-ltcg-with-32-bit-toolset/

Windows build is failing a lot of sub-builds with:  
```
Component too big for incremental LTCG with 32-bit toolset, build without incremental LTCG; Consider switching to 64-bit toolset.
```

pierotofy:
> tagged with CVEs

Ah, yes it's good to keep up with security fixes, although having a CVE does not necessarily mean there's an exploitable vulnerability in the software.

Saijin-Naib:
Your x64 PoissonRecon fix fixed the Windows build test... Still working on figuring out what's wrong with the Docker one.

Saijin-Naib:
Tapping out on this one for now... It's beyond me.

Issue #1403, opened 2022-01-13
Possible Bug: SIFT_GPU processing seems to fail (v1.9.11 B49)


Saijin-Naib:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Windows Native

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

NA

### What is the problem?

```
[INFO]    Fast orthophoto is turned on, automatically setting --skip-3dmodel
[INFO]    Initializing ODM 2.7.0 - Wed Jan 12 19:18:11  2022
[INFO]    ==============
[INFO]    auto_boundary: True
[INFO]    boundary: {}
[INFO]    build_overviews: False
[INFO]    camera_lens: auto
[INFO]    cameras: {}
[INFO]    cog: True
[INFO]    copy_to: None
[INFO]    crop: 3
[INFO]    debug: False
[INFO]    dem_decimation: 1
[INFO]    dem_euclidean_map: False
[INFO]    dem_gapfill_steps: 3
[INFO]    dem_resolution: 5
[INFO]    depthmap_resolution: 640
[INFO]    dsm: False
[INFO]    dtm: False
[INFO]    end_with: odm_postprocess
[INFO]    fast_orthophoto: True
[INFO]    feature_quality: high
[INFO]    feature_type: sift
[INFO]    force_gps: False
[INFO]    gcp: None
[INFO]    geo: None
[INFO]    gps_accuracy: 10
[INFO]    ignore_gsd: False
[INFO]    matcher_neighbors: 8
[INFO]    matcher_type: flann
[INFO]    max_concurrency: 6
[INFO]    merge: all
[INFO]    mesh_octree_depth: 11
[INFO]    mesh_size: 200000
[INFO]    min_num_features: 8000
[INFO]    name: 951218a1-b597-450e-8144-5ecd979cbeda
[INFO]    optimize_disk_space: False
[INFO]    orthophoto_compression: DEFLATE
[INFO]    orthophoto_cutline: False
[INFO]    orthophoto_kmz: False
[INFO]    orthophoto_no_tiled: False
[INFO]    orthophoto_png: False
[INFO]    orthophoto_resolution: 5
[INFO]    pc_classify: False
[INFO]    pc_csv: False
[INFO]    pc_ept: True
[INFO]    pc_filter: 2.5
[INFO]    pc_geometric: False
[INFO]    pc_las: False
[INFO]    pc_quality: medium
[...]

PR #1404, opened 2022-01-14
Update gpu.py

Saijin-Naib:
Slightly buffer image dimensions to allow for graceful fallback to CPU from GPU pathway.


Saijin-Naib:
Mostly addresses #1403 , though GPU processing still seems to not happen when I would expect the hardware is sufficient.

pierotofy:
Looks great, thanks!

PR #1414, opened 2022-02-03
LOG.ODM_INFO for GPU detection

sbonaime:
LOG.ODM_INFO for GPU detection could be useful for GPU debug


pierotofy (approved):
Thanks @sbonaime :pray:

PR #1423, opened 2022-02-28
Handle partial reconstructions

pierotofy:
This PR adds the capability for ODM to handle partial reconstructions (cases when insufficient overlap leads to multiple reconstructions).

Partial reconstructions are now attempted to be merged and reconstructed as one.

﻿


pierotofy:
image

image

ckato3:
Excellent !!!
This is a feature I really wanted to have in order to consolidate my fragmented flights into one.

PR #1439, opened 2022-03-20
Account for GPS DOP over-estimation

pierotofy:
As the title says.﻿


smathermather:
I'm sure the hardware always tells the truth... .

hung-dx-amira:
@pierotofy @smathermather I'm having some trouble handling the height of DSM and I found this line of code; does it explain the duplication? Any support would be greatly appreciated.

PR #1444, opened 2022-03-23
mod:  forcing python to use utf8

coumbsek:
This should fix two issues as discussed [here](https://community.opendronemap.org/t/non-ascii-character-in-odm-install-path/11020)

- If there are non ascii characters in the ODM install path opencv loading fails thus preventing ODM from working
- If there are non ascci characters in projet path other errors are raised.

### DIFF
--- a/winrun.bat
+++ b/winrun.bat
@@ -3,7 +3,7 @@
 setlocal
 
 call win32env.bat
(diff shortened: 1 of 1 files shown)


pierotofy:
Excellent! Thank you!

PR #1447, opened 2022-03-28
reconfigured for point cloud operation

aurvis:


pierotofy:
I'm assuming this was opened by mistake?

PR #1455, opened 2022-04-25
Memory efficient, faster planar reconstruction, binary tracks file

pierotofy:
- Improves the runtime and memory requirements of planar reconstructions
 - Adds support for binary track files (reduces the `tracks.csv` file size by ~50%)﻿. Well it's no longer csv now, but we won't change the name.


pierotofy:
Ref https://github.com/OpenDroneMap/OpenSfM/compare/283...OpenDroneMap:284

JyothiKalyan:
how do i see content of this then?

PR #1466, opened 2022-05-31
add dmpdtm.py

smathermather:
Pull request with the intent of integrating some of the magic of DMPdtm from Josh @vonnonn's adaptation of Brad Chambers implementation of Mungus et al., 2014
https://github.com/vonnonn/poodles/blob/main/DMPdtm.py

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -14,6 +14,7 @@ from scipy import ndimage
 from datetime import datetime
 from opendm.vendor.gdal_fillnodata import main as gdal_fillnodata
 from opendm import log
+from opendm.dem import dmpdtm
 try:
     import Queue as queue
 except:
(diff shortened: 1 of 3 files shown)


pierotofy:
Nice! I wonder if there's a way to refactor this to not need panda/pdal python bindings. I'll have to set some time aside to look into this, unless somebody else beats me to it.

smathermather:
> Nice! I wonder if there's a way to refactor this to not need panda/pdal python bindings. I'll have to set some time aside to look into this, unless somebody else beats me to it.

You definitely have a couple geo folks (me and Josh) pretending to do CS here. MVP in my mind is having this as an option, as the results are so staggeringly better than any other PC filtering option. Help refactoring to remove any necessary dependencies very much appreciated.

smathermather:
Looks like pdal changed how it's handling Reader. I'll likely rewrite as pipeline for reading and filtering.

vonnonn:
Dang, wish I could help. I'll back in the saddle Sunday evening.

pierotofy:
#1565 brings in the pdal python bindings (and pandas), so perhaps it will make it easier to integrate this.

smathermather:
And pandas?! Oh yeah... . Ok, lets get this done.

smathermather:
For now I'll close this in favor of the other classification work. At some point I'll compare the two.

PR #1470, opened 2022-06-16
--dem-resolution note

Saijin-Naib:
Explain that it is capped to at least 2x the GSD estimate to clarify usage and reduce user confusion.


pierotofy (approved):
:+1:

PR #1467, opened 2022-06-02
OpenSfM Update + Rolling Shutter correction

pierotofy:
Updates OpenSfM, ceres.

WIP for rolling shutter.

PR #1471, opened 2022-06-16
Hotfix for bundle_compensate_gps_bias

smathermather:
Seeing problems in lots of datasets. We should get this in to the main branch ASAP.


pierotofy (approved):
:+1: was going to get included with #1467 but we can expedite it.

smathermather:
> was going to get included with #1467 but we can expedite it.

Yeah, kinda figured that. But now that I'm looking for this problem I'm seeing it everywhere. :sweat_smile:

PR #1474, opened 2022-06-18
Add mavic air 2 readout time

nyalldawson:
As measured using https://github.com/OpenDroneMap/RSCalibration


pierotofy (approved):
Thanks @nyalldawson !!

nyalldawson:
@pierotofy thank you for the fun weekend project to do with my son! 👍

Issue #1415, opened 2022-02-03
Possible Bug: geo location file always overwrites opk


originlake:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

ubuntu

### What is the problem?

The file format of geo location file given in documents is below
```
<projection>
image_name geo_x geo_y [geo_z] [omega (degrees)] [phi (degrees)] [kappa (degrees)] [horz accuracy (meters)] [vert accuracy (meters)] [extras...]
...
```
The current way of updating will set all the optional values to `None` in the ODM_Photo object. There is no way to provide different GPS xyz data with geo.txt but still use the opk values extracted from image metadata. 

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

When this feature is implemented, the opk value is not used, so this behavior is fine. But after adding the triangulation algorithm, this might need to be carefully handled. It's also not possible to provide horizontal/vertical accuracy without overwriting the opk value.

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

[Type answer here]

originlake:
I can help with this, but need to understand what's the expected behavior. Should we overwrite the optional values to None if they are not given?  If only x and y are given, should we use z from the image or set it to none? What about opk, and horizontal/vertical accuracy. I think x,y,z, and horizontal/vertical accuracy are tied together, so it makes sense to overwrite all of them together, but for opk, I guess they need to be considered separately.
Also maybe we can add a specific value that indicates I don't want to overwrite the opk, something like none for example
```
EPSG:4326
image_name -89 32 256 None None None 0.01 0.01 
...
```

pierotofy:
I think that would be a good approach (just use `NaN` instead of `None`, which is the same syntax we have for GCP files).

Issue #1473, opened 2022-06-17
geo.txt should accept yaw/pitch/roll values, not omega/phi/kappa


pierotofy:
We currently set omega/phi/kappa values when reading from `geo.txt` files, but that's incorrect. We should set ypr angles and convert them for end users.

originlake:
@pierotofy I'm working on #1415(finally get a chance to do it), figure I could add this as well. One problem is that will the ypr value in geo.txt take the camera make into consideration? As the ypr is treated differently in different make. If not need to consider this, call `compute_opk` after updating the geo entry should do the work.

https://github.com/OpenDroneMap/ODM/blob/157a2eb90a652f33fa8c91f4c0bb6c06ecb20388/opendm/photo.py#L426-L431

pierotofy:
> If not need to consider this, call compute_opk after updating the geo entry should do the work. 

Yep! :+1:

pierotofy:
Thanks for looking into this!

PR #1476, opened 2022-06-20
Geo-location file update

originlake:
Fixed issues #1415 and #1473
- Accept yaw/pitch/roll value instead of omega/phi/kappa
- Empty or `NaN` yaw/pitch/roll inputs will not overwrite the default EXIF data.


pierotofy (approved):
:+1: this looks great! Thanks @originlake!

PR #1477, opened 2022-06-20
Fix for RasterioIOError BigTIFF

smathermather:


pierotofy:
```
  File "/code/opendm/orthophoto.py", line 132, in compute_mask_raster
    with rasterio.open(output_raster, 'w', get_orthophoto_vars(args), **rast.profile) as dst:
NameError: name 'args' is not defined
```

I'm not sure you can pass the orthophoto args to rasterio.open in that way. :smile:

smathermather:
> I'm not sure you can pass the orthophoto args to rasterio.open in that way.

Fair. :laughing: . We can just statically set it. Probably shouldn't try pull requests between taking the kid to gymnastics, eating dinner, and wrangling a baseball game.

pierotofy:
Works great now, thanks!

PR #1482, opened 2022-06-22
fix: osfm elevation check

### LINKED ISSUE #1480: Possible Bug: Altitude is ignored if one of the exif altitude happens to be 0
Having a dataset that always generates a tilted model no matter how I adjust the arguments. After investigation, I noticed that one of my images has altitude 0.0, which set `has_alt` to false due to the below logic. 

https://github.com/OpenDroneMap/ODM/blob/7f7bc332ba5a96ec53448cd32d1320d7ae788074/opendm/osfm.py#L151-L156

originlake:
Fix #1480. By default, `photo.altitude` will be None if no altitude in photo EXIF. And I didn't find a place to explicitly set its value to 0 to indicate it doesn't exist. I assume it's safe to check if it's None.

### DIFF
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -148,17 +148,25 @@ class OSFMContext:
                 photos = reconstruction.photos
 
             # create file list
+            num_zero_alt = 0
             has_alt = True
             has_gps = False
             with open(list_path, 'w') as fout:
                 for photo in photos:
                     if photo.altitude is None:
                         has_alt = False
+                    elif photo.altitude == 0:
+                        num_zero_alt += 1
                     if photo.latitude is not None and photo.longitude is not None:
                         has_gps = True
 
                     fout.write('%s\n' % os.path.join(images_path, photo.filename))
             
+            # check 0 altitude images percentage when has_alt is True
+            if has_alt and num_zero_alt / len(photos) > 0.05:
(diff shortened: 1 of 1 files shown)


pierotofy:
This is a tricky one; a lot of camera vendors will set the tag to zero when they don't have altitude information (rather than not setting it at all). There's no standard, and mileage varies by camera vendor.

If the goal is to detect whether a dataset *mostly* has altitude information (e.g. 99% of images have altitude information, but a few images don't), (which sound like the case in #1480?) a better approach might be to sample the number of photos and use altitude data if a certain threshold is met (e.g. if at least 95% of photos have altitude info, then use altitude info).

originlake:
> This is a tricky one; a lot of camera vendors will set the tag to zero when they don't have altitude information (rather than not setting it at all). There's no standard, and mileage varies by camera vendor.

I don't know this before, in this case, it is quite tricky, checking the percentage is a better idea. The dataset I mentioned in #1480 actually has all the images with valid altitudes, but some of the altitudes are around 0, (-0.1,0,0.1, etc.), and the others are collected at different elevation levels. Some of the cameras are not nadir, which causes the tilting in the model when align with orientation_prior. 

I think the logic could be, 
if altitude is None, do not use altitude
if more than 5% of images have altitudes 0, do not use altitude

pierotofy:
Yes this looks great! Thanks for taking the time to fix/improve this!! :pray:

PR #1485, opened 2022-06-24
Add readout for TeraCube TeraCube_One Mobile Phone

Saijin-Naib:
My current mobile phone that I use for all of my ground-based data collection and reconstruction.

Not a fantastic sensor (Samsung S5K2L8SX 12MP), but device has full source available at least and OpenCamera writes full Y/P/R into EXIF :D

Example:  
image


pierotofy:
Excellent, thanks!

PR #1486, opened 2022-06-29
INFO message for camera make/model/readout

Saijin-Naib:
This currently prints once per image in the Task so it gets noisy on the console.
image
I could use some help cleaning this up, if possible.

### DIFF
--- a/opendm/rollingshutter.py
+++ b/opendm/rollingshutter.py
@@ -30,6 +30,7 @@ def make_model_key(make, model):
     return ("%s %s" % (make.strip(), model.strip())).lower().strip()
 
 warn_db_missing = {}
+info_db_found = {}
 
 def get_rolling_shutter_readout(make, model, override_value=0):
     global warn_db_missing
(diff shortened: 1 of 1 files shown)


pierotofy:
There; how's this one? https://github.com/OpenDroneMap/ODM/pull/1486/commits/4a9ca4c28332f93c8742dc89067864fa742449ed

Saijin-Naib:
Beautiful, thank you so much! I was really struggling, haha.

The only other thing I can think of at this juncture that I don't know how to handle would be mixed-camera Tasks and ensuring that each camera is represented once by the appropriate WARN/INFO message.

So for instance if I had a Task that had TeraCube, GoPro, and Fuji data, I'd expect to see two INFO messages (TeraCube and GoPro) and one WARN (Fuji).

Does your fix address that mixed-camera usecase?

pierotofy:
> Does your fix address that mixed-camera usecase?

It does!

Saijin-Naib:
I'm totally happy with this! If you're happy (and the CI/CD is happy) can we press merge?

PR #1493, opened 2022-07-07
Faster band alignment, minor fixes, thermal improvements

pierotofy:
Still testing some of these changes.


pierotofy:
Good to go!

PR #1494, opened 2022-07-07
Remove odm_orthophoto.tif from transfer

smathermather:
It is silly to zip up, transfer, and unzip odm_orthophoto.tif when it is unused in merge process.


pierotofy (approved):
Nice, thanks!

PR #1496, opened 2022-07-10
Check cutline products

smathermather:


smathermather:
That is a lot of changes, so careful review appreciated. Currently running individually on a bunch of submodels that are at varying conditions of complete.

pierotofy:
Thanks, I will take a look. Can you describe why the changes are needed? What's the intent?

smathermather:
There are lots of failure cases for partially complete submodels, not the least of which is my recent pull request which doesn't download the odm_orthophoto.tif, but odm_orthophoto.py looks for it in reruns. This is step one of making reruns and failure cases more robust with split-merge.

The challenge here is all the failure cases are with massive datasets that are unreasonable (and usually not possible) to share. So: better for me to just write the fixes (probably true anyway).

pierotofy:
Would these changes be needed if `odm_orthophoto.tif` was present?

Would reverting https://github.com/OpenDroneMap/ODM/commit/26feef4fdbd65d8f949bacf4c79b8f6bbeca7d96 just fix it? (For future runs that is)

pierotofy:
One issue with the following:

```
if not io.file_exists(tree.odm_orthophoto_tif) and not io.file_exists(tree.odm_orthophoto_render):
```

Assume a (normal, non split-merge) dataset has already run to completion. When calling `./run.sh [...] --rerun-from odm_orthophoto`, `check_orthophoto_or_rerun()` evaluates to `True` (OK), but the statement will evaluate to `False` (`odm_orthophoto.tif` exists), thus the orthophoto will *not* get updated (Incorrect)

My tips for handling rerun logic without going crazy :smile: 

1. Every command that generates something needs to check *only* what it produces, plus the `rerun` variable. So if a command creates `example.txt`, check `if exists(example.txt) or rerun: then: execute command; else: warn('example.txt already exists'`. Avoid multiple and/or logic whenever possible (especially do not check for multiple outputs). 
2. Do not assume args imply previous rerun behaviors (e.g. you cannot know if a person has omitted certain parameters between reruns); these might actually be desired (rerunning with different parameters). So if possible, don't change "global" logic based on the value of `args.orthophoto_cutline`. (It's OK to check for `args.orthophoto_cutline` if the statement is concerned with the re-run of the cutline command only).
3. Simple is good (and repetition, copy/paste is also OK).

This can get tricky in certain scenarios, but follow 1 and 2 and I think this can be managed.

smathermather:
> Would reverting [26feef4](https://github.com/OpenDroneMap/ODM/commit/26feef4fdbd65d8f949bacf4c79b8f6bbeca7d96) just fix it?

No. The logic is wrong and insufficiently robust. If split merge runs locally, or cutlines are otherwise chosen for a job and the process produces an odm_orthophoto.tif but fails before producing the cutline version, we have nothing in place now that makes sure that if the cutline flag is thrown, the correct outputs are created. Reruns for cutlines are currently broken.

This also fixes failures with the update from [26feef4](https://github.com/OpenDroneMap/ODM/commit/26feef4fdbd65d8f949bacf4c79b8f6bbeca7d96). Not shipping the unclipped ortho is the correct behavior and saves on bandwidth, time, and storage, but is not the core concern.

Fundamentally, the problem with processing lots of data is that failures happen for all sorts of reasons: network, storage, RAM, user error, spot instances (not that I use them), and so the logic checks for long-running and re-running long-running processes have to be really good to be practical. My current month-long project with 10s of k of images covering a full city is currently using 2.4TB of storage, is stalled before the merge stage due to most of the above challenges. And so, with this pull request, I'm embodying in code some of the work-around that I have been doing to ensure that no one every has to worry about this with big datasets, and so split-merge becomes just a little more turnkey.

smathermather:
> * Every command that generates something needs to check _only_ what it produces, plus the `rerun` variable. So if a command creates `example.txt`, check `if exists(example.txt) or rerun: then: execute command; else: warn('example.txt already exists'`. Avoid multiple and/or logic whenever possible (especially do not check for multiple outputs).
> * Do not assume args imply previous rerun behaviors (e.g. you cannot know if a person has omitted certain parameters between reruns); these might actually be desired (rerunning with different parameters). So if possible, don't change "global" logic based on the value of `args.orthophoto_cutline`. (It's OK to check for `args.orthophoto_cutline` if the statement is concerned with the re-run of the cutline command only).

Agreed on these two points. My original intent was a full refactor of the ODMOrthoPhotoStage, but I was trying to be less radical. But, with permission to adhere to the above, as far as I can tell, I would break out into separate functions orthophoto creation, from cutline/feathering, and maybe a separate function for georeferencing. Then each step in the process has its own product that is verifiable.

Issue #1488, opened 2022-06-29
Uneven raster cell size outputs


pierotofy:
Currently outputs from ODM have slightly uneven cell raster sizes:

 ```
>>> import rasterio; rasterio.open("/datasets/brighton2/odm_orthophoto/odm_orthophoto.tif").transform
Affine(0.04999088883357982, 0.0, 576647.5603842067,
       0.0, -0.0499903086638842, 5188220.140716478)
>>> 
```

This could be improved.

**Comment by originlake** [comment]
I think the odm_orthophoto renders the orthophoto in the exact raster size provided. It calculates the orthophoto height and width by 
https://github.com/OpenDroneMap/odm_orthophoto/blob/6a7e81b7dda75e2b71d762f4b6ac89ed97e118c7/src/OdmOrthoPhoto.cpp#L383-L389
Noticed that the ceiling operation there will change the bounds of the orthophoto. But when output corner file, it still uses the original bounds. And later, gdal_translate will inject the old bounds into the orthophoto, which makes the cell size uneven.
Therefore, after the ceiling, the bounds should be changed to 
```
bounds.xMax = bounds.xMin + width/resolution_
bounds.yMin = bounds.yMax - height/resolution_
```

But this still won't fix the problem, because gdal_translate has to calculate the scale from bounds again, and floating-point error is inevitable in such a way. I corrected one orthophoto with the above bounds, it generates the transform like below
```
Data axis to CRS axis mapping: 1,2
Origin = (728469.337334620999172,5118402.260040946304798)
Pixel Size = (0.050000000000008,-0.050000000000053)
```

If you can confirm that odm_orthophoto renders the exact raster size as provided, we can directly write the scale value using rasterio (I don't find options to do it with gdal_translate).

pierotofy:
Thanks for looking into this. Like you say, the core issue is that we manually set the bounds, and precision issues are inevitable.

`gdalwarp` is able to adjust/warp the cells and create evenly sized cells, but it's slower than `gdal_translate`.

A really good improvement would be to skip the entire `odm_orthophoto --> gdal_translate` chain and generate a georeferenced raster directly in `odm_orthophoto` via calls to GDAL's API.

originlake:
My point is that odm_orthophoto has already created evenly sized cells, based on how it calculates the orthophoto width and height. So we don't need to adjust/warp the cells, the only problem is then how we properly write the transform data into the orthophoto.

The current approach have two issues
[...]

PR #1499, opened 2022-07-12
update odm_orthophoto for more even raster cell size

originlake:
Solves #1488, comes along with https://github.com/OpenDroneMap/odm_orthophoto/pull/1

The SuperBuild part is not updated for new changes, need to modify it before merging, let me know what else is needed


pierotofy:
Thanks @originlake ! I'm going to merge this into `288` which has the updated SuperBuild odm_orthophoto reference. :+1:

PR #1501, opened 2022-07-13
Bugfix

originlake:
Handle the problem in my previous commit when data is not georeferenced, tested with banana


pierotofy:
Thanks!

PR #1509, opened 2022-07-24
Update images to Ubuntu 22.04

pierotofy:
Since Ubuntu 21.04 has reached end of life and it's starting to show some update issues during build.﻿


pierotofy:
Will revisit.

Issue #1366, opened 2021-10-30
radiometric-calibration doesn't work properly


lucadelu:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker using webodm on linux server

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Firefox/Chromium on Linux

### What is the problem?

Using Multispectral images coming from Mapir Camera and choosing `radiometric-calibration: camera` I'm not able to process the data.

The problem seems during the opensfm stage because it is looking for a .tif file but instead it is a .JPG image

~~~
[INFO]    running /code/SuperBuild/install/bin/opensfm/bin/opensfm export_visualsfm --points "/var/www/data/34a4d273-e578-49ca-a7fe-3feda8bc6217/opensfm"
Traceback (most recent call last):
File "/code/SuperBuild/install/bin/opensfm/opensfm/io.py", line 1221, in image_size_from_fileobject
with Image.open(fb) as img:
File "/usr/local/lib/python3.9/dist-packages/PIL/Image.py", line 2943, in open
raise UnidentifiedImageError(
PIL.UnidentifiedImageError: cannot identify image file <_io.BufferedReader name='/var/www/data/34a4d273-e578-49ca-a7fe-3feda8bc6217/opensfm/undistorted/images/2021_0922_145250_210.JPG.tif'>
~~~

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

Finishing the procedure for the reconstruction

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

You can try with this dataset https://drive.google.com/file/d/1sa-P_VzhsuGrAw4gb6n5ePU5qhDqm-zj/view?usp=sharing
I also attach the full log
[console.txt](https://github.com/OpenDroneMap/ODM/files/7445937/console.txt)

Gurkengewuerz:
Hey @lucadelu i tried the exact same thing. I opened a community thread last night and hopefully get a response. Did you get any further? Maybe you can also post more details at the community post. 

https://community.opendronemap.org/t/odm-stops-processing-with-cannot-identify-image-file/9739

lucadelu:
[...]

Issue #1238, opened 2021-03-04
Support of 8bit multi band JPEG + 16bit single band TIFF images


zgyivi:
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker as part of WebODM

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Firefox 86 on Ubuntu Linux

### What is the problem?

DJI Multispectral camera provides 8bit RGB JPG files with EXIF data + 5 separate 16bit single band TIFF images
When I try to process such a set OpenMVS/DensifyPointCloud reports error, because the undistorted RGB images
are converted to 16bit 3band TIFF format and it cannot process that.

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

Undistorting should keep 8bit format for multiband images even if there are 16bit single band files in the set.

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

The problem occur on the image set which was shared by mapninja on community page
[DJI Phantom 4 Pro Multipectral w/o RTK](https://community.opendronemap.org/t/dji-phantom-4-pro-multipectral-w-o-rtk/4978?u=zgyivi)
Only the JPGs can be processed properly, but when all images added processing fails as described above.
Consol log shows ([link to whole log](https://drive.google.com/file/d/1RdW8SGt7cC2HfRbV_bH9DV8ZPDp9ECpY/view?usp=sharing)): 
[INFO]    Initializing ODM - Sat Feb 27 23:14:17  2021
[INFO]    ==============
[INFO]    build_overviews: False
[INFO]    camera_lens: auto
[INFO]    cameras: {}
[INFO]    crop: 3
[INFO]    debug: False
[INFO]    dem_decimation: 1
[INFO]    dem_euclidean_map: False
[INFO]    dem_gapfill_steps: 3
[INFO]    dem_resolution: 5
[INFO]    depthmap_resolution: 640
[INFO]    dsm: False
[INFO]    dtm: False
[INFO]    end_with: odm_report
[INFO]    fast_orthophoto: False
[INFO]    feature_quality: high
[INFO]    feature_type: sift
[INFO]    force_gps: False
[INFO]    gcp: None
[INFO]    geo: None
[INFO]    gps_accuracy: 10
[INFO]    ignore_gsd: False
[...]

PR #1512, opened 2022-07-25
Support for RGB+Multispec image inputs

pierotofy:
Adds support for processing mixed RGB+Multispectral image inputs (e.g. Phantom 4 Multispectral files), you can now throw JPEGs (8bit) and TIFFs (16bit) and the dataset will process correctly.

You can also provide redundant data (RGB JPEGs + R/G/B TIFFs) and ODM will handle that gracefully.

Still need to add support for better radiometric calibration in these cases (byte to float conversion will currently still fail). Will try to improve that in the upcoming days.


pierotofy:
There, look at those beautiful float32 values derived from 8bit RGB JPEGs and mixed with 16bit TIFFs.

image

pierotofy:
Fixes #1366, #1238, #1512

PR #1518, opened 2022-07-31
Chunked median smoothing for odm_dem stage

originlake:
Hi,

For dem generation, median smoothing could take around half of the processing time, due to it being processed in a single thread.
This implementation allows applying the kernel-based filters in chunks so that we can have multi-core performance. It has some limitations
1. It should use multithreading along with GIL-released filters to have memory efficiency and multi-core performance. 
2. It supports most padding modes in the filters, but methods like 'wrap' in the median filter will not give identical results.
3. Behavior with no data value is unpredictable (depends on the filter's implementation). It currently follows the old strategy

I used `joblib.parallel` to achieve multi-threading, as `opendm.concurrency.parallel_map` is not able to return results. `joblib.parallel` should be able to handle ctrl-c gracefully as it describes. Let me know if need to follow the implementation in `opendm.concurrency`.

Some benchmarks

dsm size | Single filter | -- | Chunked filter(16 workers) | -- | Chunked filter(8 workers) |--| Chunked filter(2 workers) |--| Chunked filter(1 workers)| -- 
-- | -- | -- | -- | -- | -- |-- | -- | -- | -- | -- |
-- | Time | Peak Ram | Time | Peak Ram | Time | Peak Ram | Time | Peak Ram | Time | Peak Ram
28116x22706 | 4m26.8s | 8881.1mb | 28.4s | 10024.7mb | 38.8s | 10020.7mb | 2m18.7s | 10024.5mb | 4m32.0s | 10013.4mb
14058x11353 | 1m37.3s | 2850.6mb | 20.6s | 2682.6mb | 23.8s | 2691.1mb | 56.3s | 2696.7mb | 1m39s.75 | 3319.7mb
7029x5676 | 23.0s | 839.1mb | 2.4s | 814.7mb | 3.3s | 810.0mb | 12.0s | 807.5mb | 23.0s | 805.4mb
3514x2838 | 5.9s | 353.0mb | 0.7s | 359.9mb | 1.0s | 348.8mb | 3.1s | 346.0mb | 6.0s | 344.6mb

Note that the time includes reading and writing GeoTIFF, depending on the profile of the GeoTIFF, it could have different results. For example, in the above benchmarks, I used the final dsm.tif as input, which seems to read and write faster compared to the raw DSM generated in the pipeline. So you may observe slower results when checking the processing time directly in odm_dem stage.

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -5,6 +5,8 @@ import numpy
 import math
 import time
 import shutil
+import functools
+from joblib import delayed, Parallel
 from opendm.system import run
 from opendm import point_cloud
 from opendm import io
@@ -270,7 +272,7 @@ def create_dem(input_point_cloud, dem_type, output_type='max', radiuses=['0.56']
                 '"{tiles_vrt}" "{geotiff}"'.format(**kwargs))
 
     if apply_smoothing:
-        median_smoothing(geotiff_path, output_path)
+        median_smoothing(geotiff_path, output_path, num_workers=max_workers)
         os.remove(geotiff_path)
     else:
         os.replace(geotiff_path, output_path)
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @originlake ! This looks very promising (and awesome!). I will review it soon. :pray:

pierotofy:
Looks great, thank you @originlake ! :clinking_glasses:

PR #1519, opened 2022-08-02
288

konradkern:
Please add this rolling shutter correction.  'undefined undefined': 15, # Evo 2 Pro  Not sure how to complete undefined.  I think the flashing light used was 2k hertz, so the shutter read-out should be 30.


pierotofy:
Hey @konradkern :hand: can you share the image you captured with the sensor? 

Undefined means that the camera does not have a camera make / model tag (which might be problematic).

konradkern:
I have attached the image I provided in the pull request.  The original
measurement was done by Daniel Debosi and posted to the Open Drone Map
forum.  I believe he used the 2 khz flashing light technique also.

https://community.opendronemap.org/t/phantom-4-photo-request-rolling-shutter/11777/55?u=daniel.dobosi

Konrad
ph 714-658-4972

pierotofy:
You'll need to get the original image (the one posted on the forum has been resized/edited by the forum software, so all tags including the important make/model tags have been removed).

PR #1524, opened 2022-08-11
Adding DJI Mavic 2 Pro

DE-CaTeR:
Adding rolling shutter readout for DJI Mavic 2 Pro (16:10 and 4:3)

### DIFF
--- a/opendm/rollingshutter.py
+++ b/opendm/rollingshutter.py
@@ -14,6 +14,8 @@ RS_DATABASE = {
     'dji fc7203': 20, # Mavic Mini v1
     'dji fc3170': 27, # DJI Mavic Air 2
     'dji fc3411': 32, # DJI Mavic Air 2S
+    
+    'hasselblad l1d-20c': lambda p: 47 if p.get_capture_megapixels() < 19 else 56, # DJI Mavic 2 Pro (at 16:10 => 16.8MP 47ms, at 4:3 => 19.9MP 56ms)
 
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Thanks @DE-CaTeR ! :pray:

kikislater:
Thank you for this ! 
Some recommend to use 3:2 image size. Any values for this ratio ? @DE-CaTeR @pierotofy 

Reference : https://topodrone.com/upload/iblock/946/User%20Manual%20TOPODRONE%20DJI%20Mavic%202%20Pro%20L1_L2%20RTK_PPK%20[FULL].pdf
Page 4
> 2. Set correct image ratio. Go to the image menu, tap the second
tab. Use the Image Ratio 3:2 value, it allows to use all square of
camera’s matrix. Do not use 16:9 Ratio! Check another values
of this menu – all defaults.

kikislater:
Additionnal info : 

> For example, DJI Mavic 2 Pro has the readout time of 62 milliseconds. If the drone moves at the speed of 10 m/s and at the altitude of 100 meters above the ground, the vertical distortion caused by the rolling shutter effect will be 23.88 pixels. The camera will shift to almost 57 cm during the image scanning time. 

Source: https://topodrone.com/blog/917/

kikislater:
https://github.com/OpenDroneMap/ODM/issues/1537

DE-CaTeR:
Hi Sorry the comment in the code is wrong. But the code is still fine - native 16:9 Image has 19.9 MP and 3:4 has 16.8MP
I uploaded the two testpictures.
MP17_DJI_0814_47lines_16_10
MP20_DJI_0813_56lines_4_3

Issue #1314, opened 2021-07-06
Native MacOS Support


pierotofy:
At some point we should add native support for MacOS.

PR #1530, opened 2022-09-06
Apple Native

pierotofy:
Changes that allow ODM to compile/run natively on MacOS. Tested on M1 so far. I will test on Intel soon too.

I'm debating how to best package this for deployment, and how to automate M1 builds (GitHub runners do not provide M1 runners, we would need to rent one, which I'd rather not...).

image

One needs to install dependencies via homebrew, including:

```
homebrew install tbb@2020 gdal python@3.8 [...]
```

I didn't write down the full list (ups).


pierotofy:
Closes #1314

pierotofy:
Runs end-to-end:

image

pierotofy:
Compilation steps:

Install:

 * XCode
 * Homebrew

Then run:

```bash
brew install cmake gcc@12 python@3.8 tbb@2020 eigen gdal boost cgal libomp
brew link tbb@2020

python3.8 -m pip install virtualenv
python3.8 -m virtualenv venv
source venv/bin/activate
pip install -r requirements.txt

cd SuperBuild
mkdir build
cmake ..
make -j8
```

Finally:

```bash
cd /tmp
pip download GDAL
tar -xpzf GDAL-3.5.1.tar.gz
cd GDAL-3.5.1
python setup.py build_ext --gdal-config /opt/homebrew/bin/gdal-config
python setup.py build
python setup.py install
```

Smoke test:

```
./run.sh --help
```

pierotofy:
Code signing./ notarization is also a terrible mess, I'm unable to see how this could be automated with ease. Several tools that *should* allow one to make an automated build require UI interaction to allow permission / ask passwords (e.g. access to the keychain).

pierotofy:
Good to go.

ckato3:
In Native Install (MacOS),      . /run.sh      is not running.

It stops at "from opensfm.sensors import sensor_data".

I think it is because there are no files or folders for opensfm, sensors, sensor_data, etc. in the odm folder.

I am new to python, so I don't know how to solve this problem.
What should I do?

======
Coffee241:ODM ckato$ ./run.sh --help
  File "/Users/ckato/ODM/run.py", line 12, in <module>
    from opendm import system
  File "/Users/ckato/ODM/opendm/system.py", line 15, in <module>
    from opendm import types                                                                                                    # WC++
  File "/Users/ckato/ODM/opendm/types.py", line 18, in <module>
    from opendm.photo import ODM_Photo
  File "/Users/ckato/ODM/opendm/photo.py", line 25, in <module>
    from opensfm.sensors import sensor_data
ModuleNotFoundError: No module named 'opensfm'
======

pierotofy:
The recommended way to run ODM on Apple is by using docker. :+1:

ckato3:
I understand that Python beginners should use a docker...

I got an M1 Mac and am considering using docker to run ODM.
But during the docker build, gcc is killed during the OpenSfM build.
I think it is because my M1 Mac has only 8 GB memory.

I will let you know the detailed situation later, since I cannot use M1 Mac until the end of the week.

winstonford:
I get same error "ModuleNotFoundError: No module named 'opensfm'
Following docs here https://opensfm.org/docs/building.html I believe I broke path to python.
Now on smoke test I get error:
"No module named 'numpy.core._multiarray_umath'"

Intel Core i5
MacOS 12.6.2 
Homebrew 3.6.18
package-id: com.apple.pkg.CLTools_Executables
version: 13.4.0.0.1.1651278267

PR #1533, opened 2022-09-19
Add AI background removal

### LINKED ISSUE #1528: Automatic Background Removal
Similarly to how we create sky masks via `--sky-removal`, we could also include a model to perform background removal (useful for modeling 3D objects) via a `--bg-removal` flag.

pierotofy:
Adds automatic background removal via a novel `--bg-removal` which implements the method published by Daniel Gatis from https://github.com/danielgatis/rembg

Closes #1528


pierotofy:
image

image

pierotofy:
Fixed a bug with image masks on GPU too, so `--sky-removal` should work better as well.

We now use all features from the images (ignoring the masks) so OpenSfM should work better also in these cases.

image

No post-cropping required.

Compared to original:

image

Issue #1537, opened 2022-09-28
Wrong rolling shutter applied for Mavic 2 Pro 4:3 ratio


kikislater:
### How did you install ODM? (Docker, installer, natively, ...)?

Docker

### What is the problem?

Wrong rolling shutter applied for 4:3 image 

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.

Assuming that the rolling shutter varies with the height of the image, the code for mavic 2 pro is only applied for 3:2 and not 4:3 as specified in the code comment. 

![DJI Image size](https://djiphotoacademy.com/wp-content/uploads/2020/10/Aspect-Ratio-for-Phantom4Pro-and-Mav-Air-768x823.jpg)

[Mavic 2 Pro], 20MP, 5472×3648 = **native** ratio of 3:2.

```
$ python -c "print(4864*3648/1000000)"
17.743872
```

So in 4:3 ratio 16:9 readout will be applied. 4:3 and 3:2 have same height so theoretically same sensor readout 

### How can we reproduce this? What steps did you do to trigger the problem? If this is an issue with processing a dataset, YOU MUST include a copy of your dataset uploaded on Google Drive or Dropbox (otherwise we cannot reproduce this).

Just check the code ...

PR #1538, opened 2022-09-28
M2P rolling shutter correction. Solve issue #1537

kikislater:
https://github.com/OpenDroneMap/ODM/issues/1537


pierotofy:
Thanks @kikislater ! 🙏

PR #1544, opened 2022-10-17
Multi-GPU support

pierotofy:
Adds support for using multiple graphics cards during depthmap computation via OpenMVS.


pierotofy:
Adds `--sm-no-align` (credits @smathermather )

PR #1550, opened 2022-11-07
OpenMVS stability improvements, GCP warning checks

pierotofy:
- More memory resiliant openmvs
- GCP warning checks


sbonaime:
@pierotofy Maybe ODM should remove GCP with less than 3 tagged images in addition to the warning ?

PR #1551, opened 2022-11-14
3.0.0 release

pierotofy:
This is going to be a release with a few breaking changes, marking thus a major 3.x release.

I'm going to leave this as a draft for a while to hear feedback on the proposed changes:

 - [x] Removed `--resize-to` which was deprecated long ago (in favor of `--feature-quality`).
 - [x] Removed `--depthmap-resolution`, which is redundant with `--pc-quality`.
 - [x] Removed `--pc-geometric`, which is now enabled by default. The point cloud results with `pc-geometric` are much sharper and accurate. While it does increase runtime, very few users remember know/remember to check this flag. If there's a case for reducing the runtime later on, I'd rather add a `--skip-geometric-estimation` flag or something (change from default off to on).
 - [x] Removed `--texturing-data-term`, which I believe is not really used by many. A 3D model is almost always better textured using `gmi`.
 - [x] Removed `--texturing-outlier-removal-type`, which I think it's always good to leave turned on, and `gauss_clamping` (the default) is already the best working method.
 - [x] Removed `--texturing-tone-mapping` which I think nobody ever touches (defaults to `none`).
 - [x] Removed `--verbose`, which was toggling on only a few extra printout statements. We now always print all available information. There might be a case in the future for adding a `--quiet` flag, which prints no messages and is probably a better logic.
 - [x] Removed `--debug`, which was used in only two places.
 - [x] Removed `--time`; we now always generate a `benchmark.txt` file by default.

Furthermore, I removed the visibility filtering in the point cloud processing (which was obliterating thin structures in many datasets). We still have the `odm_filterpoints` in place to help remove outliers and I find that the completeness of the reconstruction outweighs the outliers that are introduced by the removal of this step. Due to the better completeness, it also improves the look of orthophotos in general.

Do not merge yet.


Saijin-Naib:
No complaints from me, though I have seen (rare) cases where area worked better than GMI (roadways with moving vehicles being one I remember). Everything else sounds excellent.

I think especially the removal of the deprecated/conflicting quality flags will be a huge win and stop a bunch of confusion.

smathermather:
[Cross posted on the forum](https://community.opendronemap.org/t/opendronemap-3-0/13732). Most of these changes look pretty benign to me. The one I care about most is `pc-geometric` being default on (yay!) and visibility filtering being off. I'll be super interested to see how much this improves towers and similar.

Saijin-Naib:
I've got a number of ground-level collects I can test these changes against, so that will be interesting to see the changes for sure.

Miner-Jeff:
I am happy to re-run some of my own previously run data sets or some of the sample datasets on an early windows installer version if it helps.

ckato3:
Is it possible to leave the "--orthophoto-no-tiled" option?
I'm using IgorPro, a commercial, general-purpose powerfull data analysis software, to import orthophoto output of ODM for advanced image statistical analysis.
This analysis software cannot read tiled-GeoTIFF format and requires them to be in striped-GeoTIFF format.

Is it possible to leave the "--debug" option?
I'm using this flag to output additional debug information about multispectral analysis.

pierotofy:
Good point on compatibility with some software, we'll keep `--orthophoto-no-tiled`. 

On the `--debug`, which information do you currently need and how are you using it?

Saijin-Naib:
> Good point on compatibility with some software, we'll keep `--orthophoto-no-tiled`. 
> 
> On the `--debug`, which information do you currently need and how are you using it?

Could all the debug and info messages we had just always be printed to the console?

ckato3:
Thank you for keeping the option "--orthophoto-no-tiled".

======
When additional debug messages are huge, having all of these messages always output to the console would result in a huge console log file, which would adversely affect execution time and storage space.
With the --debug option, additional debug messages can be toggled on and off depending on the situation.

======
Additional debugging information includes ......

(1) The start and completion times of each step in the ODM calculation are appended to the console log to keep track of the progress of the calculation in REAL TIME.   Similar information would be written to "benchmark.txt", but the console output is easier to keep track of in REAL TIME.

(2) Information from the drone's sensors.   For example, if the drone is hit by a gust of wind during aerial photography, the yaw, low, and pitch angles and ground speeds (X Y Z) will be disturbed, resulting in poor quality photos, so this information is usefull to check  if a gust is present or not.

======
To use the additional debugging information, we use the following command

bash docker run -it --rm -v .............       |& tee odmlog.txt | grep -i 'KEYWORD'

"tee" is used to split the console log into file output and console output, and "grep" is used to display only the necessary additional debugging information in the console output. The additional debugging information is put on a single line and includes a 'KEYWORD' for each type of debugging information.

smathermather:
> I am happy to re-run some of my own previously run data sets or some of the sample datasets on an early windows installer version if it helps.

Reach out via DM on the forums and I can give you access to a test instance. Now Windows installer yet, AFAIK, but I can facilitate testing.

pierotofy:
Good to go.

PR #1552, opened 2022-11-15
Static masking

smathermather:
Not sure how best to include MIT license within the repo, so move/rename/fix as appropriate.


pierotofy (approved):
A license.txt file suffices. :+1:  Thanks!

PR #1555, opened 2022-11-26
Fix improper depthmap resolution parameter

pierotofy:
Found this while performing some tests with higher `orthophoto-resolution` parameters. It turns out we should not set `--min-resolution` when calling OpenMVS, as we push the threshold for scaling images lower than the recommended value of `640` (OpenMVS's default).


Saijin-Naib:
Huh... Do you think this might help with low-resolution (thermal?) datasets?

PR #1556, opened 2022-11-27
GSD scale factor improvements, looser point visibility checks

pierotofy:
- Less aggressive point visibility filtering
- GSD image scale factor always 2**-N


pierotofy:
Rationale for GSD to always apply scaling by a factor is two, is that different `orthophoto-resolution` parameters can end up affecting point cloud quality in some cases, irrespective of `--pc-quality`, which is not the desired behavior.

This might increase memory usage slightly, but should also provide for sharper orthophotos (and reduce the need for --ignore-gsd).

Point cloud visibility was removed a few commits ago, but after running more datasets it's clear that we need it. I've increased the threshold to have it be less aggressive so it should be an improvement over the previous release.

PR #1558, opened 2022-11-29
upgrade laspy to 2.3.0 with laszip support

originlake:
The latest laspy allows us to directly read and write laz files, hence we can get rid of pdal conversion between las and laz.
It also brings many new designs that could be helpful for future development, such as chunked read/write, and header-only read.

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -50,35 +50,12 @@ def rectify(lasFile, debug=False, reclassify_threshold=5, min_area=750, min_poin
     start = datetime.now()
 
     try:
-        # Currently, no Python 2 lib that supports reading and writing LAZ, so we will do it manually until ODM is migrated to Python 3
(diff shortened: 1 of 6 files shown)


pierotofy:
Thanks @originlake ! :pray: this looks great!

pierotofy:
Ugh, I had to revert this unfortunately @originlake; it seems that the laspy[laszip] package is not available on ARM (and our docker images need to work on both x64 and ARM). This issue was highlighted during the build:

```
#19 909.5 ERROR: Could not find a version that satisfies the requirement laszip<0.2.0,>=0.1.0; extra == "laszip" (from laspy[laszip]) (from versions: 0.0.1)
#19 909.5 ERROR: No matching distribution found for laszip<0.2.0,>=0.1.0; extra == "laszip"
```

We might need to wait for laspy[laszip] support for ARM before this can be merged into master.

originlake:
OK, I didn't take ARM into consideration. Will check if there is any workaround.

PR #1560, opened 2022-12-02
upgrade laspy to 2.3.0 with laszip support

originlake:
Used a different dependency which is lazrs (a rust implementation that has arm64 support directly, described [here](https://laspy.readthedocs.io/en/latest/installation.html))

Lazrs does not support points with waveforms (I don't think it would be a problem here), but it benefits from multi-threaded compression/decompression. 

I didn't test the full build on arm64 (qemu build is quite slow), but able to use an existing arm64 ODM image to install laspy[lazrs] and used it to load a laz file.

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -50,35 +50,12 @@ def rectify(lasFile, debug=False, reclassify_threshold=5, min_area=750, min_poin
     start = datetime.now()
 
     try:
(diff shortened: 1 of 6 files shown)


pierotofy (approved):
Awesome! Looks good. Thank you for taking the time to find an alternative lib for ARM. :pray:

PR #1561, opened 2022-12-05
Add Sensor Readout: FUJIFILM X-T2

Saijin-Naib:
35ms for the FUJIFILM X-T2 Sensor Readout
ISO:ISO-12,800
Shutter:1/10,000s (Forced Electronic Shutter)


pierotofy:
Thanks! I added some missing commas.

Saijin-Naib:
How embarrassing 😬

Thanks for the save, as always!

Issue #1557, opened 2022-11-29
Use larger sensor database


pierotofy:
We could improve OpenSfM support for various cameras by using a more comprehensive database of CCD width information, like the one that DroneDB uses (which is an aggregate of OpenSfM + AliceVision).

PR #1565, opened 2022-12-13
Automatic outputs alignment via CODEM

### LINKED ISSUE #1564: Possible Bug: Qhull precision error: Initial simplex is flat
### How did you install ODM? (Docker, installer, natively, ...)?
Docker, Alpine Linux Latest-Stable
### What is the problem?
On a few (seemingly random) datasets, reconstruction will fail with an error in QHull about degenerate hyperplane(s). Each time there is nothing obviously wrong with the data, on the contrary, they're usually exceptionally well-captured data. The solution always is to completely strip EXIF Geolocation metadata, and then the reconstruction will succeed.
There should not be this error... I can't figure out exactly what is happening, but it seems like something with co-incident camera positions.
### How can we reproduce this? What steps did you do to trigger the problem? If this is an issue with processing a dataset, YOU MUST include a copy of your dataset uploaded on Google Drive or Dropbox (otherwise we cannot reproduce this).
Example dataset:  
https://hub.dronedb.app/r/roofergaming/r4gzbklfs8ym8mjo
Example dataset thread:  
https://community.opendronemap.org/t/cannot-process-dataset-other-projects-without-problems/13268?u=saijin_naib
Something of note:  
```
I removed the GPS data from every 6th picture and the calculation was done without any problems.
```
Reconstruction with no EXIF Geolocation Metadata:  
image

pierotofy:
This PR adds the ability to automatically register a reconstruction (and resulting orthophotos, point clouds, 3D models, etc.) to a reference DSM (GeoTIFF, LAS or LAZ) by using CODEM (https://github.com/NCALM-UH/CODEM, credits to Craig Glennie and the group at CRREL). 

The details of the method are available at https://github.com/NCALM-UH/CODEM/blob/main/docs/details.md

Usage:

* Pass a file via `--align <path-to-tif-or-las-laz>`.
* Place a `align.tif`, or `align.laz` or `align.las` in the project root (similar to GEO and GCP files).

Sample output report:

image

This could be particularly useful for repeat flights of the same area where automatic alignment allows for better change detection analysis, without needing GCPs.

Also fixes #1564


pierotofy:
Implements #1557 via https://github.com/OpenDroneMap/OpenSfM/commit/cca015d1b11c276d1ff61526b0d460f6544c7c7b

pierotofy:
Passes test suite. Ready to go.

smathermather:
Woah. I wonder if some interaction between this approach and the alignment step in split merge might be meaningful.

PR #1567, opened 2022-12-14
Added video2dataset module

HeDo88TH:
This PR adds the module for creating the dataset from a video.
The algorithm used is able to discard out of focus frames and too similar ones.

Input parameters:

-   `input` = path to input video file
-   `output` = path to output directory
-   `start` = start frame index
-   `end` = end frame index
-   `output_resolution` = Override output resolution (ex. 640x480)
-   `blur_percentage` = discard the lowest X percent of frames based on blur score (allowed values from 0.0 to 1.0)
-   `blur_threshold` = blur measures that fall below this value will be considered 'blurry' (to be used in exclusion with -bp)
-   `distance_threshold` = distance measures that fall below this value will be considered 'similar'
-   `frame_format` = frame format (jpg, png, tiff, etc.)
-   `stats_file` = Save statistics to csv file
-   `internal_width` = We will resize the image to this width before processing
-   `internal_height` = We will resize the image to this height before processing


pierotofy:
Nice; how can we test this?

opendm/video2dataset/parameters.py
```diff
@@ -0,0 +1,89 @@
...
+        self.frame_format = args["frame_format"]
+        self.output_resolution = tuple(map(int, args["output_resolution"].split("x"))) if args["output_resolution"] else None
+        self.stats_file = args["stats_file"] if args["stats_file"] else None
+
+        # We will resize the image to this size before processing
+        self.internal_width = 800
+        self.internal_height = 600
+
+    # Validate parameters
+    def ValidateParameters(self, args):
```
pierotofy: Why are there two almost identical `ValidateParameters` functions?

opendm/video2dataset/requirements.txt
```diff
@@ -0,0 +1,3 @@
+opencv-python
```
pierotofy: Don't need a requirements.txt file.

pierotofy:
Can you add more info on `distance_threshold`? What are good values? What units?

opendm/video2dataset/parameters.py
HeDo88TH: Old files got pushed

opendm/video2dataset/requirements.txt
HeDo88TH: Old files got pushed

HeDo88TH:
> Can you add more info on `distance_threshold`? What are good values? What units?

Technically it's the average distance between detected features of the two images. We use the iterative Lucas-Kanade method with pyramids and we only consider good features that were found in both the images.

HeDo88TH:
Here is an example of how you can use the Video2Dataset class:


```
# Import the required modules
from video2dataset import Parameters, Video2Dataset

# Set the input and output parameters
params = Parameters()
params.input = "./input_video.mp4"
params.output = "./output_dir"

# Set the start and end frames
params.start = 0
params.end = 100

# Set the blur percentage and distance threshold
params.blur_percentage = 0.2
params.distance_threshold = 3

# Create an instance of the Video2Dataset class and process the video
v2d = Video2Dataset(params)
v2d.ProcessVideo()
```
This example will process the video file input_video.mp4, discard the 20% lowest quality frames based on blur score, discard too similar frames and save the remaining frames as images in the `output_dir` directory. The frames will be saved in JPEG format and will have the same resolution as the original video. The processing will start from the first frame and end at the 100th frame.

I found that 0.2 (20%) as `blur_percentage` and 3 as `distance_threshold` are good candidates for default values.

pierotofy:
Test code:

```
# Import the required modules
from opendm.video.video2dataset import Parameters, Video2Dataset

# Set the input and output parameters
params = Parameters()
params.input = "/datasets/house/House.mp4"
params.output = "/datasets/house/images"

# Set the start and end frames
params.start = 0
params.end = 100

# Set the blur percentage and distance threshold
params.blur_percentage = 0.2
params.distance_threshold = 3

# Create an instance of the Video2Dataset class and process the video
v2d = Video2Dataset(params)

```

```
  File "test_video.py", line 19, in <module>
    v2d.ProcessVideo()
  File "/code/opendm/video/video2dataset.py", line 121, in ProcessVideo
    print("Calculated threshold is {}".format(self.blur_checker.threshold))
AttributeError: 'PercentageBlurChecker' object has no attribute 'threshold'
```

pierotofy:
Future improvements (no need to implement these right away, just notes):

 - [x] Black frame detection (many videos have a black/fade transition frames and should be discarded): http://ffmpeg.org/ffmpeg-all.html#blackdetect
 - [x] SRT geotagging (something similar to https://community.opendronemap.org/t/how-to-create-geotagged-image-frames-from-video/14077/2 but without explicitly calling exiftool, done via Python)
 - [x] Process multiple video files inputs
 - [x] EXIF tag population for Image Width, Image Height, Camera Make / Model (are they available from the video?), Orientation, Create Date, Date/Time Original (each frame being offset by the correct number of seconds)
 - [x] For subsecond timestamps we should also populate GPS time (EXIF SubSecTime). https://github.com/OpenDroneMap/ODM/blob/master/opendm/photo.py#L267
 - [x] Add a parameter to choose the maximum number of video frames (e.g. `1000` should extract at most 1000 video frames). Needs to work in combination with the other parameters.
 - [x] Remove percentage blur checker which is not useful for ODM.
 - [x] Have the function return a list of created video frame filenames instead of True | False.
 - [x] On error, throw Exceptions instead of returning True | False.

HeDo88TH:
Added:

- Multiple input files https://github.com/OpenDroneMap/ODM/pull/1567/commits/9efea5a966d38feacf94c5e560b2f12128d9d83e#diff-52ea5021193a4183fb81d7f6ada887ba60e09127e8f1a858479d190fcd211b0fR45
- Black frame detection https://github.com/OpenDroneMap/ODM/pull/1567/commits/9efea5a966d38feacf94c5e560b2f12128d9d83e#diff-c5021b4277cc5fb00d212285263803faaa6910e172357c3ea4e7f58f0e79b812R126
- SRT file parser https://github.com/OpenDroneMap/ODM/pull/1567/commits/9efea5a966d38feacf94c5e560b2f12128d9d83e#diff-8590c2ed3f7c5aed9c180e7378d330b12e662c2594995fa9504fb3e48efb50aaR1
- Exif tag population from video metadata and SRT file https://github.com/OpenDroneMap/ODM/pull/1567/commits/9efea5a966d38feacf94c5e560b2f12128d9d83e#diff-52ea5021193a4183fb81d7f6ada887ba60e09127e8f1a858479d190fcd211b0fR201

The new parameters are the following:

- `input` now accepts a list of file paths separated by commas
- `black_ratio_threshold` and `pixel_black_threshold` follow the same method of https://ffmpeg.org/ffmpeg-filters.html#toc-blackdetect. The first determines the `picture_black_ratio_th`, how many black pixels in a frame make it count as a black frame (percentage). The latter is the `pixel_black_th` used to calculate the absolute threshold for considering a pixel "black": 
`absolute_threshold = luminance_minimum_value + pixel_black_th * luminance_range_size`
- `use_srt` make the program look for a srt file with the same name of the video file and try to use it for frame metadata. You can specify an `utc_offset` to account for timezones in srt frame timestamps (defaults to "local")
- EXIF tag population. Are now supported: `datetime_original`, `datetime_digitized`, `datetime`, `pixel_x_dimension`, `pixel_y_dimension`, `orientation`, `software`, `model`; (SRT ONLY): `exposure_time`, `focal_length`, `f_number`, `latitude`, `longitude`, `altitude`, `photographic_sensitivity`.

HeDo88TH:
NOTE: I patched the `exif` python package to support the writing of the `photographic_sensitivity` (ISO) tag. 
The edited `requirements.txt` contains the git link of the patched repo. (PR is pending)

pierotofy:
Could have used Pillow (which is already a dependency) to write EXIF. https://github.com/python-pillow/Pillow/issues/4935

HeDo88TH:
> Could have used Pillow (which is already a dependency) to write EXIF. [python-pillow/Pillow#4935](https://github.com/python-pillow/Pillow/issues/4935)

Yes, I will switch!

opendm/video/video2dataset.py
```diff
@@ -1,51 +1,18 @@
...
-        self.frame_format = "jpg"
-        self.output_resolution = None
-        self.stats_file = None
-
-        # We will resize the image to this size before processing
-        self.internal_width = 800
-        self.internal_height = 600
-
+from opendm.video.parameters import Parameters
+import pymediainfo
```
pierotofy: Can we not include this? It requires libmediainfo, which is going to be a pain to get to work on Windows.

opendm/video/video2dataset.py
```diff
@@ -0,0 +1,295 @@
...
+        elapsed_time = start_time + datetime.timedelta(seconds=(self.frame_index / video_info.frame_rate))
+
+        # Set datetime_original
+        img.datetime_original = elapsed_time.strftime('%Y:%m:%d %H:%M:%S')
+        img.datetime_digitized = elapsed_time.strftime('%Y:%m:%d %H:%M:%S')
+        img.datetime = elapsed_time.strftime('%Y:%m:%d %H:%M:%S')
+        img.pixel_x_dimension = frame.shape[1]
+        img.pixel_y_dimension = frame.shape[0]
+        img.orientation = video_info.orientation if video_info.orientation is not None else Orientation.TOP_LEFT
+        img.software = "Video2Dataset"
```
pierotofy: `ODM`

opendm/video/srtparser.py
```diff
@@ -0,0 +1,124 @@
...
+        self.filename = filename
+        self.data = []
+        self.utc_offset = utc_offset
+
+    def get_entry(self, timestamp):
+        if not self.data:
+            self.parse()
+
+        for entry in self.data:
+            if entry["timestamp"] <= timestamp:
```
pierotofy: Assuming self.data is sorted from least to most recent, which I think it is by reading `parse`, doesn't this end up returning always the first RST entry?

Can you link the test data you're using?

opendm/video/video2dataset.py
HeDo88TH: I was working and testing on windows, it works pretty well. What alternative do you propose?

pierotofy: Ah, never mind; they ship binaries with the wheel. It might be OK then.

pierotofy:
Tested, ready to go!

PR #1568, opened 2022-12-15
Update rollingshutter.py

twchambers:
Added calibration value for Mavic Pro (=64). Value is valid for both Mavic Pro and Mavic Pro Platinum, as both use camera 'DJI FC220'


pierotofy (approved):
Thanks @twchambers ! :pray:

PR #1572, opened 2022-12-20
Add --feature-quality help warning

pierotofy:
Should help people that are confused as to why choosing `Ultra` is a bad idea for GPU usage.


smathermather:
Yeah. Good idea.

Issue #1571, opened 2022-12-17
Possible regression in OpenMVS, geometrically consistent views.


smathermather:
Here is the failure I am seeing. It's not the most enjoyable test dataset with 2k camera positions, but happy to share as needed.

Ubuntu 20.04, docker install, FWIW.

```
Geometric-consistent estimated depth-maps 1448 (72.11%, 27m13s, ETA 10m)...
Geometric-consistent estimated depth-maps 2008 (100%, 27m13s529ms)

===== Dumping Info for Geeks (developers need this to fix bugs) =====
Child returned 1
Traceback (most recent call last):
File "/code/stages/odm_app.py", line 81, in execute
self.first_stage.run()
File "/code/opendm/types.py", line 386, in run
self.next_stage.run(outputs)
File "/code/opendm/types.py", line 386, in run
self.next_stage.run(outputs)
File "/code/opendm/types.py", line 386, in run
self.next_stage.run(outputs)
[Previous line repeated 1 more time]
File "/code/opendm/types.py", line 365, in run
self.process(self.args, outputs)
File "/code/stages/openmvs.py", line 117, in process
raise e
File "/code/stages/openmvs.py", line 103, in process
run_densify()
File "/code/stages/openmvs.py", line 99, in run_densify
system.run('"%s" "%s" %s' % (context.omvs_densify_path,
File "/code/opendm/system.py", line 109, in run
raise SubprocessException("Child returned {}".format(retcode), retcode)
opendm.system.SubprocessException: Child returned 1

===== Done, human-readable information to follow... =====

[ERROR]   Uh oh! Processing stopped because of strange values in the reconstruction. This is often a sign that the input data has some issues or the software cannot deal with it. Have you followed best practices for data acquisition? See https://docs.opendronemap.org/flying/
```

pierotofy:
Would be interesting to patch openmvs.py and disable the geometric consistency to see if it processes through.

Also would be really helpful to narrow down a smaller subset of images that causes the crash.

smathermather:
Currently running through a local split-merge with `--split 200` to get sub 500-image submodels. We'll see if there are particular ones that fail.

Saijin-Naib:
> Currently running through a local split-merge with `--split 200` to get sub 500-image submodels. We'll see if there are particular ones that fail.

Is this stable enough to work almost like a binary bisect of datasets? If so, this needs to go into my mental toolkit...

smathermather:
[...]

PR #1580, opened 2023-01-10
add --pc-skip-geometric

smathermather:
Possible fix for https://github.com/OpenDroneMap/ODM/issues/1571

That issue isn't fully tested (downtime during the holidays, what?), but given others are having issues with large datasets on the forum, I think it is worth adding this flag.


pierotofy:
Thanks!

smathermather:
Midnight pull requests with no closing quote... . Thanks!

PR #1582, opened 2023-01-11
Add --texturing-single-material

### LINKED ISSUE #1525: Add option to export OBJs using a single material
Add a `--texturing-single-material` option that post-processes the 3D model output to use a single material (something similar to https://github.com/theFroh/imagepacker/blob/master/objuvpacker.py, but with support for multispectral TIFFs inputs, thus using rasterio instead of PIL).

pierotofy:
Closes #1525

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -454,6 +454,13 @@ def config(argv=None, parser=None):
                         help=('Keep faces in the mesh that are not seen in any camera. '
                               'Default:  %(default)s'))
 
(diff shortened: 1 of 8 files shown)


pierotofy:
From:

image

To:

image

Packing program can also be invoked as a standalone module:

```
[piero:/code/opendm/objpacker] python3 objpacker.py /datasets/brighton2/odm_texturing/odm_textured_model_geo.obj
```

smathermather:
woohoo!

PR #1583, opened 2023-01-17
won centimeter

smathermather:
Context:
https://github.com/OpenDroneMap/ODM/issues/1584


pierotofy:
Some discussion in #1584

PR #1590, opened 2023-01-28
Better support for AeroVironment Quantix

pierotofy:
This PR improves support for AeroVironment Quantix drones. It's a bit hackish because band information does not seem to be available in EXIF meta (unless test images from https://community.opendronemap.org/t/quantix-mapper-multispectral-data/6706 got postprocessed, but doesn't seem like it), so we set the band name from the filename.

This drone has two sensors, a RGB camera and GRN camera, of different image resolutions, which are physically offset. https://www.avinc.com/images/uploads/product_docs/Quantix_Mapper_Datasheet_07222021.pdf

The current image alignment method might be able to get some results, but I would expect alignment issues. Investigation into camera rigs might be required here for mixing/matching the images from these two sensors.

I would recommend to process the images from sensors separately if alignment issues arise. Also, when processing from the multispectral sensor camera, I find that setting `--primary-band nir --sfm-algorithm planar --skip-band-alignment --fast-orthophoto` is a must, as the Green and Red bands seem to have a very weak signal. Perhaps this was just an issue with the test dataset. OpenMVS also kept crashing for some reason (thus `--fast-orthophoto`).

In short, I think this sensor might still require more work, but this should at least give users a chance to process some data and start experimenting.

### DIFF
--- a/SuperBuild/cmake/External-OpenSfM.cmake
+++ b/SuperBuild/cmake/External-OpenSfM.cmake
@@ -25,7 +25,7 @@ ExternalProject_Add(${_proj_name}
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
   GIT_REPOSITORY    https://github.com/OpenDroneMap/OpenSfM/
-  GIT_TAG           303
+  GIT_TAG           304
   #--Update/Patch step----------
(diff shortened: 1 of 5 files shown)


pierotofy:
image

pierotofy:
image

PR #1606, opened 2023-02-17
Feature/admin64bit

coumbsek:
As discussed in [this topic](https://community.opendronemap.org/t/install-for-all-users-native-windows/14805/4) I made two modifications in the innosetup configuration file:
- Install program as a 64bit program because it is one. (Currently it appears under the 32bit programs install on the 64bit computer in the registry)
- All the program to be install for all users with elevated privileged if launched in command line with `/ALLUSERS` flag


pierotofy:
Thanks @coumbsek! 🙏

Issue #1608, opened 2023-02-19
Request For Comments: Set DEM Product Scaling Factor to 1


Saijin-Naib:
### How did you install ODM? (Docker, installer, natively, ...)?
Docker, but this affects all platforms

### What is the problem?
Currently, we cap DEM products to 2x the GSD of the dataset.

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.
The main historical reasons for this have been to more accurately reflect the precision of our reconstruction, and to help keep resource consumption for this part of the pipeline under control.

I believe that at this point, with all the incredible work to reconstruction completeness, precision, accuracy, multi sub-resolution depthmaps, geometrically consistent depthmaps, rolling shutter correction, enhanced filtering, etc. that we no longer need to cap the DEM products to 2x dataset GSD, and that in fact, we are ready to have DEM products able to be created at 1:1.

### How can we reproduce this? What steps did you do to trigger the problem? If this is an issue with processing a dataset, YOU MUST include a copy of your dataset AND task output log, uploaded on Google Drive or Dropbox (otherwise we cannot reproduce this).
![image](https://user-images.githubusercontent.com/19295950/219909620-234859f9-962a-4805-8ff2-84771d85e113.png)
Brighton Beach Road - v3.0.4
```
Options:	crop: 0, dem-resolution: 0.1, dsm: true, dtm: true, feature-quality: ultra, min-num-features: 128000, orthophoto-resolution: 0.1, pc-filter: 0, pc-quality: ultra, primary-band: green, rolling-shutter: true
```

![image](https://user-images.githubusercontent.com/19295950/219909649-46a6b6d7-79e8-476f-a3a5-01ea5d594f66.png)
Brighton Beach Road - v3.0.4 1:1 GSD Patch
```
Options:	crop: 0, dem-resolution: 0.1, dsm: true, dtm: true, feature-quality: ultra, min-num-features: 128000, orthophoto-resolution: 0.1, pc-filter: 0, pc-quality: ultra, primary-band: green, rolling-shutter: true
```
Brighton Beach Road -v3.0.4
![image](https://user-images.githubusercontent.com/19295950/219909660-089d534c-dbbb-4a88-acb5-4428771eb78a.png)

I'm seeing a significant increase in real resolved detail with @smathermather patch.

[...]

PR #1609, opened 2023-02-19
Alternatives to fixed gsd_scaling for DEMs

smathermather:
Reference https://github.com/OpenDroneMap/ODM/issues/1608

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -536,7 +536,7 @@ def config(argv=None, parser=None):
                         metavar='<float>',
                         action=StoreValue,
                         type=float,
-                        default=10,
+                        default=5,
                         help='DSM/DTM resolution in cm / pixel. Note that this value is capped to 2x the ground sampling distance (GSD) estimate. To remove the cap, check --ignore-gsd also.'
                              ' Default: %(default)s')
 
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -536,7 +536,7 @@ def config(argv=None, parser=None):
                         metavar='<float>',
                         action=StoreValue,
                         type=float,
-                        default=5,
+                        default=10,
                         help='DSM/DTM resolution in cm / pixel. Note that this value is capped to 2x the ground sampling distance (GSD) estimate. To remove the cap, check --ignore-gsd also.'
                              ' Default: %(default)s')
 
--- a/stages/odm_dem.py
+++ b/stages/odm_dem.py
@@ -31,7 +31,7 @@ class ODMDEMStage(types.ODM_Stage):
 
         # It is probably not reasonable to have accurate DEMs a the same resolution as the source photos, so reduce it
         # by a factor!
-        gsd_scaling = 2.0
+        gsd_scaling = 1.0
 
         resolution = gsd.cap_resolution(args.dem_resolution, tree.opensfm_reconstruction, 
                                         gsd_scaling=gsd_scaling,
--- a/stages/odm_dem.py
+++ b/stages/odm_dem.py
@@ -31,7 +31,10 @@ class ODMDEMStage(types.ODM_Stage):
 
         # It is probably not reasonable to have accurate DEMs a the same resolution as the source photos, so reduce it
         # by a factor!
-        gsd_scaling = 1.0
+        if args.pc_quality in {'ultra'}
+            gsd_scaling = 1.0
+        else:
+            gsd_scaling = 2.0
 
         resolution = gsd.cap_resolution(args.dem_resolution, tree.opensfm_reconstruction, 
                                         gsd_scaling=gsd_scaling,


opendm/config.py
```diff
@@ -536,7 +536,7 @@ def config(argv=None, parser=None):
                         metavar='<float>',
                         action=StoreValue,
                         type=float,
-                        default=5,
+                        default=10,
```
pierotofy: I'm not sure I understand; why is this being changed?

smathermather: Perhaps I am wrong on the current logic, but my assumption is that if someone runs something with e.g. dsm+dtm, they'll get an orthophoto at the default 5cm and their DEM products at 10cm because of the existing gsd_scaling variable. If folks are used to these defaults and depend on them to avoid crashing with their available memory, we might not want to change that behavior.

pierotofy: So there's multiple things to consider (which I think might have not been considered here and #1608 ):
 * `gsd_scaling` applies only in regard to the *maximum* allowed resolution per GSD cap
 * `--dem-resolution` (capped, sometimes by an estimate that uses `gsd_scaling`) always sets the output raster resolution (width of the raster cells). If you want ortho+dtm+dsm to have the same raster resolution, they need to be both `5`. Changing it to `10` will decrease your DEMs resolution by half.
 * `--dem-resolution` *also* affects the search radius for binning points. See `odm_dem.py` and `radius_steps` for how that works.

Maybe the real intent is to tweak/decrease the `radius_steps`, which can sometimes bring on more details if your point cloud is dense enough, but not always, it's not a perfect science and benchmarks should be run on a variety of datasets (with different point densities / capture altitude) to establish if improvements are a net-positive across datasets.

smathermather: I'm not sure how this changes my understanding, but perhaps there are still elements I am missing. Let's dive in and find out what I'm missing:

> gsd_scaling applies only in regard to the maximum allowed resolution per GSD cap

Understood. It is the cap we are asserting needs changed. If I fly a photo set at 1cm GSD, set my orthophoto resolution to 1cm or finer, and ODM estimates ~1cm GSD, I would like to have the capacity without using `ignore-gsd` to set my DEM resolution to the maximum available resolution.

> --dem-resolution (capped, sometimes by an estimate that uses gsd_scaling) always sets the output raster resolution (width of the raster cells). If you want ortho+dtm+dsm to have the same raster resolution, they need to be both 5. Changing it to 10 will decrease your DEMs resolution by half.

This is the current default behavior as I understand it with datasets of 5cm: orthophoto ends up 5cm and DEM ends up 10cm because it currently hits the gsd_scaling cap. But, thinking about it further, changing the defaults only makes sense for that narrow set of cases, so I will revert this change in the pull request.

> --dem-resolution also affects the search radius for binning points. See odm_dem.py and radius_steps for how that works.

I am familiar, and have tuned those parameters for specific datasets when necessary, usually after post-processing the point cloud by using pc2dem.py for fussier elevation model datasets.

> Maybe the real intent is to tweak/decrease the radius_steps, which can sometimes bring on more details if your point cloud is dense enough, but not always, it's not a perfect science and benchmarks should be run on a variety of datasets (with different point densities / capture altitude) to establish if improvements are a net-positive across datasets.

The intent is to match the current state of the art with respect to available resolution, which given our current quality of point clouds is IMO attainable. Tuning of radius steps is already user configurable, and the defaults seem generally fine. Now, if we need to change those in response to resolution cap changes then that's a different set of constraints, and we need to finally have benchmark datasets sufficient in variety. Which... that's on my list.

pierotofy: > changing the defaults only makes sense for that narrow set of cases 

:+1: 

> given our current quality of point clouds 

Well, I think it depends on the settings. For example, this is brighton beach (master) run with:

``` 
./run.sh /datasets/brighton --dem-resolution 0.00001 --dsm
```

image


And with the proposed patch:

image

Now, if I also double the `radius_steps` I get a better result:

image

But is it really?

master:

image

patched (+ radius_steps increase):

image

The point being: I think we should totally work toward setting `gsd_scaling` to `1`, but I don't think we can/should just change `2` to `1` and be done with it.

pierotofy: A possible approach might involve the computation of actual average point spacing in the point cloud to set a better value to radius_steps, rather than using output resolution as a starting point.

smathermather: Hmm. Agreed this is a good counter point, and building logic for discerning point cloud density makes sense, but I think filling real data is a useful pre-step to improving interpolation, i.e. actually tying to `pc-quality` first. `gsd_scaling` of 1 on both sides here, both `pc-quality` of `ultra`. The one on the left has some extra radius steps that make nominal improvements:

image

Propose, in advance of a method for discerning point density, we test with different values for gsd_scaling that are dependent on pc-quality that roughly replicate what exists, but raises the ceiling at higher pc-quality values.

smathermather: Modified the pull request to only allow gsd_scaling of 1.0 with Ultra.

pierotofy:
Closing in favor of #1610

PR #1610, opened 2023-02-20
Point spacing estimation

pierotofy:
This PR brings in changes to FPCFilter so that we can compute point cloud density estimates.

It also ties in such density estimates into the generation of DEMs, thus disassociating resolution from radius parameters.

This makes it possible to render higher resolution rasters without affecting the radius. It also no longer limits the `gsd_scaling` factor.

Related #1609 #1608

### DIFF
--- a/SuperBuild/cmake/External-FPCFilter.cmake
+++ b/SuperBuild/cmake/External-FPCFilter.cmake
@@ -8,7 +8,7 @@ ExternalProject_Add(${_proj_name}
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
   GIT_REPOSITORY    https://github.com/OpenDroneMap/FPCFilter
-  GIT_TAG           main
+  GIT_TAG           305
   #--Update/Patch step----------
   UPDATE_COMMAND    ""
   #--Configure step-------------
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -376,6 +376,6 @@ def window_filter_2d(arr, nodata, window, kernel_size, filter):
 def get_dem_radius_steps(stats_file, steps, resolution, multiplier = 1.0):
     radius_steps = [point_cloud.get_spacing(stats_file, resolution) * multiplier]
     for _ in range(steps - 1):
-        radius_steps.append(radius_steps[-1] * math.sqrt(2)) # sqrt(2) is arbitrary, maybe there's a better value?
+        radius_steps.append(radius_steps[-1] * math.sqrt(2))
     
     return radius_steps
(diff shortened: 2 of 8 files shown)


pierotofy:
Current master with defaults:

image

This PR with defaults:

image

This PR with `--dem-resolution 0.000001`:

image

Which looks a bit noisy, but I'd say that's expected given the point cloud density...

pierotofy:
I'd encourage to test this one as a possible solution to #1608 @smathermather and @Saijin-Naib, in particular the method for point cloud density estimation, which I opted to do in a somewhat novel way but has not been tested. (Check if the `density` value in `odm_filterpoints/point_cloud_stats.json` makes sense if something weird happens).

smathermather:
I probably won't have time to test as I'm traveling this week, but @Saijin-Naib -- there will be a node at 3002 (already added and permissions shared with you) if you've got capacity to throw some dataset through.

It's not ready yet, but should be within the hour (it's building).

Saijin-Naib:
Yeah, absolutely, I'll get on that throughout the week

pierotofy:
A lot of changes, I'm noticing improvements to orthophotos across the board as well, but more testing is needed before I feel confident to merge.

Saijin-Naib:
This makes me happy!

I have all of our bolded ODMData datasets enqueued for processing, so we have decent coverage coming.

Any non-bolded anyone wants to see?

pierotofy:
sheffield_park_1, sand_key would be interesting.

pierotofy:
Biggest change should be a noticeable decrease in average relief displacement around buildings:

Master:

image

This PR:

image

Saijin-Naib:
This is incredibly exciting, as I get this as a point of feedback quite consistently.

Extrapolating from here, I'd expect to improve reconstruction on solar panels as well, but we've not had any datasets submitted for that type of scenery, unfortunately.

Saijin-Naib:
I'm still collecting test results, but at first blush, your changes have massively improved the appearance of tree canopies, which is always nice to note.

Saijin-Naib:
Quick Comparison with some key areas of difference on mf_old_orchard (4.56 cm/px GSD):  
```
crop: 0, dem-resolution: 0,1, dsm: true, dtm: true, feature-quality: ultra, orthophoto-resolution: 0.1, pc-filter: 0, pc-quality: ultra
```
image
image
image
image
image
image
image
image

Some quick things that seem to have improved:
- Tree Canopies
- Straight Lines (crop rows, traffic markers, building edges, etc)
- Relief/Extrusion (individual drainage tiles are visible)

It seems almost like what currently is Ultra ends up being more like Medium or Low in comparison to experimental v3.0.5 4a0b60b.

Rest of the datasets are processing up still.

atobiese:
Great work, thanks to all of you for this. I will run testing with this last commit. I tried to share the discussion regarding the  scaling factor from the forum to @pierotofy, but does not seem like odm-discourse is set up for sharing dm's. We discussed the improvements in dem's with removal of the gsd cap compare to that of the orthophoto, use case is DEM over snow. Looking forward to testing and will get back with results.

PR #1612, opened 2023-02-25
fix outdated docs for angles in geo file

zfb132:
According to the [question](https://community.opendronemap.org/t/geolocation-files-versus-exif-and-omega-phi-kappa-versus-pitch-roll-yaw/9905/2), [\[issue 1473\]](https://github.com/OpenDroneMap/ODM/issues/1473), [\[pr 1476\]](https://github.com/OpenDroneMap/ODM/pull/1476) and the following code, the `omega`, `phi`, `kappa` should be replaced to `yaw`, `pitch`, `roll`.

https://github.com/OpenDroneMap/ODM/blob/92cab06a5140730ce166fc22ea18aebbc9bc6093/opendm/geo.py#L39-L43

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -485,11 +485,11 @@ def config(argv=None, parser=None):
                         action=StoreValue,
                         default=None,
                         help=('Path to the image geolocation file containing the camera center coordinates used for georeferencing. '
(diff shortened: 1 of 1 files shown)


pierotofy:
Good catch, thanks @zfb132 ! :pray:

PR #1614, opened 2023-03-01
Remove laspy

### LINKED ISSUE #1600: Remove LASPY
There's no need to use LASPY, we already have PDAL (including python bindings), and the only module that uses LASPY is the rectify module.

HeDo88TH:
Closes #1600 

Tested on Linux. The code has been modified to use PDAL instead of Laspy, which has resulted in the elimination of one dependency. However, I was unable to test the code path where the `write_extra_dimensions` parameter is set to `True`. @pierotofy, could you kindly provide a test case for this scenario? Thank you!

Please note that the auto-formatter has made some edits to the `commands.py` file by removing duplicate spaces. Please disregard these changes. The only modification I made was to the `rectify` method, where I removed the laspy workaround for macOS.

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -49,13 +49,11 @@ def classify(lasFile, scalar, slope, threshold, window):
 def rectify(lasFile, reclassify_threshold=5, min_area=750, min_points=500):
     start = datetime.now()
 
-    pcFile = lasFile
-
     try:
 
         log.ODM_INFO("Rectifying {} using with [reclassify threshold: {}, min area: {}, min points: {}]".format(lasFile, reclassify_threshold, min_area, min_points))
         run_rectification(
-            input=pcFile, output=pcFile, \
+            input=lasFile, output=lasFile, \
             reclassify_plan='median', reclassify_threshold=reclassify_threshold, \
             extend_plan='surrounding', extend_grid_distance=5, \
             min_area=min_area, min_points=min_points)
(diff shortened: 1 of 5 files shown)


opendm/dem/commands.py
```diff
@@ -83,7 +73,7 @@ def create_dem(input_point_cloud, dem_type, output_type='max', radiuses=['0.56']
                 decimation=None, keep_unfilled_copy=False,
                 apply_smoothing=True):
     """ Create DEM from multiple radii, and optionally gapfill """
-    
+
```
pierotofy: Please remove any formatting changes that are not relevant to the work?

opendm/dem/commands.py
```diff
@@ -107,18 +97,18 @@ def create_dem(input_point_cloud, dem_type, output_type='max', radiuses=['0.56']
     RES_FLOOR = 64
     if w < RES_FLOOR and h < RES_FLOOR:
         prev_w, prev_h = w, h
-        
+
```
pierotofy: Same.

opendm/dem/commands.py
```diff
@@ -107,18 +97,18 @@ def create_dem(input_point_cloud, dem_type, output_type='max', radiuses=['0.56']
...
         if w >= h:
             w, h = (RES_FLOOR, int(math.ceil(ext_height / ext_width * RES_FLOOR)))
         else:
             w, h = (int(math.ceil(ext_width / ext_height * RES_FLOOR)), RES_FLOOR)
-
```
pierotofy: Same (you get the idea)...

pierotofy:
> Please note that the auto-formatter [...]

Please disable it and clean up any changes. It makes it difficult to review what actually changed. :pray:

pierotofy:
`write_extra_dimensions` can be removed as a parameter, it's used only for debug. (You'll need to also remove any references to it, e.g. in `def run_rectification(**kwargs):` and up the call stack).

HeDo88TH:
> `write_extra_dimensions` can be removed as a parameter, it's used only for debug. (You'll need to also remove any references to it, e.g. in `def run_rectification(**kwargs):` and up the call stack).

All done 👍

pierotofy:
:+1:

PR #1618, opened 2023-03-08
Add support docker user

### LINKED ISSUE #1413: ODM processing from docker container run with -u parameter
### How did you install OpenDroneMap? (Docker, natively, ...)?

Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Firefox 96, Manjaro Linux

### What is the problem?

I cannot finish full procession of orthophoto when I run docker container with docker's user different from root (`-u 1000:1000`).
```
docker run -u 1000:1000 -ti --rm -v "${PWD}":/datasets opendronemap/odm:latest --project-path /datasets project_small
```
The problem is described in more detail in the thread:
https://community.opendronemap.org/t/odm-docker-processing-fails-while-runing-the-same-from-webodm-works/10295/5

The ODM processing should succeed, output files should be generated in `/datasets` directory with ownership as a user of ID 1000 and group 1000.

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

Run processing of any dataset in docker container using `-u` option as shown above.

HeDo88TH:
Closes #1413

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,5 +51,12 @@ RUN bash configure.sh installruntimedepsonly \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
   && bash run.sh --help \
   && bash -c "eval $(python3 /code/opendm/context.py) && python3 -c 'from opensfm import io, pymap'"
+
+# Switch to appuser
+RUN adduser --disabled-password --gecos "" --uid 1000 appuser
+RUN chown -R appuser /code
+RUN chown -R appuser /usr/local
+USER appuser
+
 # Entry point
 ENTRYPOINT ["python3", "/code/run.py"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,7 @@ RUN bash configure.sh installruntimedepsonly \
 RUN adduser --disabled-password --gecos "" --uid 1000 appuser
 RUN chown -R appuser /code
 RUN chown -R appuser /usr/local
+RUN chown -R appuser /tmp
 USER appuser
 
 # Entry point
--- a/gpu.Dockerfile
+++ b/gpu.Dockerfile
@@ -46,5 +46,13 @@ RUN bash configure.sh installruntimedepsonly \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
   && bash run.sh --help \
   && bash -c "eval $(python3 /code/opendm/context.py) && python3 -c 'from opensfm import io, pymap'"
+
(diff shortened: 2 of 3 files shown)


pierotofy:
* `portable.Dockerfile` hasn't been updated.
 * The current changes will double the image size (  2.71GB from ~1.63GB)
 * This will cause "permission denied" errors while mounting volumes belonging to any directory that doesn't belong to uid `1000` (on macOS) and any other Linux user that is not the first/default user (plus any distro that doesn't start uids with 1000).

If we didn't need to mount volumes, this would be fine, but we do.

Honestly I'm not even sure there's a good solution here, aside from creating a separate image that allows rootless execution.

HeDo88TH:
> * `portable.Dockerfile` hasn't been updated.
> * The current changes will double the image size (  2.71GB from ~1.63GB)
> * This will cause "permission denied" errors while mounting volumes belonging to any directory that doesn't belong to uid `1000` (on macOS) and any other Linux user that is not the first/default user (plus any distro that doesn't start uids with 1000).
> 
> If we didn't need to mount volumes, this would be fine, but we do.
> 
> Honestly I'm not even sure there's a good solution here, aside from creating a separate image that allows rootless execution.

The increase in size is inexplicable. We can remove the `--uid 1000` and try it, I will do some more testing.

For example, Bitnami uses ad-hoc rootless docker images: 
https://docs.bitnami.com/tutorials/work-with-non-root-containers
https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-containers/#root-and-non-root-containers

HeDo88TH:
I got an explanation for the increase in size. The ownership of the files and the subsequent `USER` statement should be put before the compilation. I'm working on some changes to address it.

pierotofy:
You just need to execute everything in the same RUN statement.

Every RUN statement adds a new layer in docker; and by changing every file with `chown` the RUN statements are creating new layers with twice the size.

But this is not even the most pressing point; what about:

>  This will cause "permission denied" errors while mounting volumes belonging to any directory that doesn't belong to uid 1000 (on macOS) and any other Linux user that is not the first/default user (plus any distro that doesn't start uids with 1000).

If this cannot be figured out, it's unnecessary to work on the other points.

I *personally* don't think there's a good solution.

pierotofy:
Closing as the concept of root-less execution for a container that needs read access to a mounted volume, with files created by different users is a no-go.

One can still create their own docker image by following the modifications made in this PR.

PR #1619, opened 2023-03-13
Fix pc rectify userdata

### LINKED ISSUE #1602: pc-rectify obliterates "views" property in the point cloud
This yields an improper survey map in the report. The "views" field is stored in the UserData dimension.

image

HeDo88TH:
Closes #1602 

As title says, the `UserData` dimension was not copied over to the new point cloud.


pierotofy (approved):
:+1: Excellent, thank you!

PR #1621, opened 2023-03-17
Add Autel Nano+ rolling shuitter value

lurenzzzz:
'autel robotics xl724': 29, # Autel Nano+
MAX_0013


pierotofy:
Awesome, thanks @lurenzzzz !

Issue #1622, opened 2023-03-20
Improve some warnings messages


pierotofy:
Some warning messages are unnecessarily alarming and confusing to users. They could use improvements.

Issue #1620, opened 2023-03-16
Possible Bug: Rolling Shutter Correction unable to handle non-Geolocated Imagery


Saijin-Naib:
### How did you install ODM? (Docker, installer, natively, ...)?
Docker, latest pull as of 2023-03-15.

### What is the problem?
Consistent failure to process a sample dataset with RSC enabled, though it processes fine with same settings otherwise (and RSC disabled). Sensor readout is accurate and present in our database.

### What should be the expected behavior? If this is a feature request, please describe in detail the changes you think should be made to the code, citing files and lines where changes should be made, if possible.
If RSC can't work with non-geolocated imagery, a graceful fallback or fixed up error message that is more meaningful would be helpful.

### How can we reproduce this? What steps did you do to trigger the problem? If this is an issue with processing a dataset, YOU MUST include a copy of your dataset AND task output log, uploaded on Google Drive or Dropbox (otherwise we cannot reproduce this).
Console Log:  
```
[INFO]    Initializing ODM 3.0.5 - Thu Mar 16 05:06:28  2023
[INFO]    ==============
[INFO]    3d_tiles: False
[INFO]    align: None
[INFO]    auto_boundary: True
[INFO]    auto_boundary_distance: 0
[INFO]    bg_removal: True
[INFO]    boundary: {}
[INFO]    build_overviews: False
[INFO]    camera_lens: auto
[INFO]    cameras: {}
[INFO]    cog: True
[INFO]    copy_to: None
[INFO]    crop: 3
[INFO]    dem_decimation: 1
[INFO]    dem_euclidean_map: False
[INFO]    dem_gapfill_steps: 3
[INFO]    dem_resolution: 5
[INFO]    dsm: False
[INFO]    dtm: False
[INFO]    end_with: odm_postprocess
[INFO]    fast_orthophoto: False
[INFO]    feature_quality: ultra
[INFO]    feature_type: hahog
[INFO]    force_gps: False
[INFO]    gcp: None
[INFO]    geo: None
[INFO]    gltf: True
[INFO]    gps_accuracy: 10
[INFO]    ignore_gsd: False
[INFO]    matcher_neighbors: 64
[INFO]    matcher_type: bow
[INFO]    max_concurrency: 8
[INFO]    merge: all
[INFO]    mesh_octree_depth: 11
[INFO]    mesh_size: 200000
[INFO]    min_num_features: 128000
[...]

Issue #884, opened 2018-08-31
Extract / Classify Buildings


pierotofy:
It has been suggested at the drone panel at FOSS4G to add the ability to classify point clouds to extract buildings contours / footprints. A raster approach could also be interesting.

pierotofy:
I've started working on this.

PR #1623, opened 2023-03-20
Automated point classification and other improvements

pierotofy:
Adds completely automated point cloud classification for:
 * Ground
 * Vegetation
 * Buildings

Some points remain in the "unclassified" class (mostly vehicles, but also some near-ground building points which are hard to distinguish).

image


pierotofy:
Fixes #1620

pierotofy:
Fixes #1622

pierotofy:
Closes #884 :tada:

PR #1627, opened 2023-03-29
Fix strange python launcher error when folder has spaces

### LINKED ISSUE #1569: Windows installation in non-default path causes error
> installed at a non-default path (in my user dir) and tried to run run.bat but it failed because it couldn't find geos_c.dll at ./venv/Library/bin/geos_c.dll .  Copying it from the shapely site-packages to that location solved it though

HeDo88TH:
This PR fixes the behavior of innosetup to use the windows short path when it contains spaces.
The issue seems to be related to https://github.com/pypa/setuptools/issues/216 and several others. I tested the same exact setup with newer python versions and they work fine. Unfortunately we are stuck with python 3.8 due to breaking changes in newer versions.

Bonus: added `vcpkg`, `venv`, `python38`, `dist` and `innosetup` folders to gitignore.

Closes #1569

### DIFF
--- a/.github/workflows/test-build-prs.yaml
+++ b/.github/workflows/test-build-prs.yaml
@@ -66,12 +66,6 @@ jobs:
       uses: jwlawson/actions-setup-cmake@v1.13
       with:
         cmake-version: '3.24.x'
-    - name: Extract code signing cert
-      id: code_sign
-      uses: timheuer/base64-to-file@v1
-      with:
-        fileName: 'comodo.pfx'
-        encodedString: ${{ secrets.CODE_SIGNING_CERT }}
     - name: Install venv
       run: |
         python -m pip install virtualenv
@@ -84,10 +78,8 @@ jobs:
         rmdir SuperBuild\build /s /q
       shell: cmd
     - name: Create setup
-      env: 
-        CODE_SIGN_CERT_PATH: ${{ steps.code_sign.outputs.filePath }}
       run: |
-        python configure.py dist --code-sign-cert-path $env:CODE_SIGN_CERT_PATH
+        python configure.py dist
     - name: Upload Setup File
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/test-build-prs.yaml
+++ b/.github/workflows/test-build-prs.yaml
@@ -58,17 +58,38 @@ jobs:
       with:
         python-version: '3.8.1'
         architecture: 'x64'
+    - uses: Jimver/cuda-toolkit@v0.2.4
+      id: cuda-toolkit
+      with:
+        cuda: '11.4.0'
     - name: Setup cmake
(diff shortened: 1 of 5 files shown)


pierotofy:
Thanks. I'm a bit hesitant to accept this as a solution, it's my understanding that you're simply forcing users to install ODM to a path without spaces (via shortpath)? I also have concerns about that "Delete Folder" statement (potentially dangerous) and the ASCII encoding requirement (will this even work with non-ASCII characters in the path?)

If this issue was fixed in newer versions of Python, can't we just backport the patch?

For example, have you attempted to apply https://github.com/pypa/setuptools/pull/2697 ?

HeDo88TH:
We abandoned the short path workaround in favour of merging the folders `python38` and `venv\Scripts`. 
It works and test pipeline runs!

pierotofy:
Need to wait for https://gitlab.com/gitlab-org/gitlab/-/issues/402616 to test this (it's causing the ODM builds to fail).

```
2023-03-29T18:11:09.9728890Z     Performing download step (download, verify and extract) for 'ext_eigen'
2023-03-29T18:11:10.0050628Z     -- Downloading...
2023-03-29T18:11:10.0053940Z        dst='D:/a/ODM/ODM/SuperBuild/build/mvstexturing/elibs/ext_eigen/src/eigen-3.3.2.tar.gz'
2023-03-29T18:11:10.0054871Z        timeout='none'
2023-03-29T18:11:10.0056023Z        inactivity timeout='none'
2023-03-29T18:11:10.0057117Z     -- Using src='https://gitlab.com/libeigen/eigen/-/archive/3.3.2/eigen-3.3.2.tar.gz'
2023-03-29T18:11:10.8564153Z     -- verifying file...
2023-03-29T18:11:10.8565675Z            file='D:/a/ODM/ODM/SuperBuild/build/mvstexturing/elibs/ext_eigen/src/eigen-3.3.2.tar.gz'
2023-03-29T18:11:10.8657097Z     -- MD5 hash of
2023-03-29T18:11:10.8657908Z         D:/a/ODM/ODM/SuperBuild/build/mvstexturing/elibs/ext_eigen/src/eigen-3.3.2.tar.gz
2023-03-29T18:11:10.8659324Z       does not match expected value
2023-03-29T18:11:10.8660476Z         expected: '02edfeec591ae09848223d622700a10b'
2023-03-29T18:11:10.8661183Z           actual: '7b8ab470dbf9d22bf4adef60f9f8f4f2'
```

Not related to the changes here.

HeDo88TH:
> Need to wait for https://gitlab.com/gitlab-org/gitlab/-/issues/402616 to test this (it's causing the ODM builds to fail).

They messed up some hashes 😢

pierotofy:
Everything checks out, thanks!

PR #1631, opened 2023-04-14
Photometric Keypoint Refinement

pierotofy:
Test code for an implementation of keypoint refinement.

It's currently too slow to make it into master, but could be improved.

It tends to improve relative accuracy measures by ~30% in a variety of datasets.

Original keypoints (red) vs. corrected (blue):

image


Saijin-Naib:
That is incredible!

Would it be viable behind a flag that Defaults to Off and isn't in any preset?

We can then let folks who need that precision get it after they're read up on it, and we can get more real-world tests while a refinement happens later?

pierotofy:
Folks that want to test it out can rebuild from this branch (or pull the relevant changes from ODM/OpenSfM). :+1:

pierotofy:
Not enough bandwidth to further evaluate this approach.

PR #1632, opened 2023-04-17
Add  camera-lens fisheye_opencv, better support for Parrot Sequoia

pierotofy:
Adds support for `--camera-lens fisheye_opencv`, which is a more advanced model for fisheye distortion.

I tested this on a set of Parrot Sequoia datasets and improves the results drastically.

I would expect this to improve other fisheye cameras datasets as well.


Saijin-Naib:
I need to test this against the data from the Parrot CHUCK platforms, which have been broken for a long time.

Would it make sense to replace the default fisheye handler with this?

pierotofy:
> Would it make sense to replace the default fisheye handler with this? 

Time will tell.

PR #1638, opened 2023-04-23
Automatic pc-tile (remove --pc-tile)

pierotofy:
ODM already automatically chooses the best method to launch OpenMVS (with pc-tile as a fallback), so there's little reason to have `--pc-tile` as an option.


Saijin-Naib:
What do you think about the possible use-case of a user forcing it to enabled to try and skirt SWAP/PAGE and speed things along, despite it being an automated fall-back for when things fail during densification?

I've been using it in pretty much all my presets for the above reason.

pierotofy:
That's a valid use case, but that means you've also been spending more time than necessary in the OpenMVS step when the dataset did fit in memory.

I'm of the opinion that this stuff should be automatic (to the extent possible).

PR #1640, opened 2023-05-01
Edge Inpainting

pierotofy:
This PR significantly improves RGB orthophoto rendering by using a novel edge inpainting technique.

Edges of building and trees are often clearly marked by pixel intensity differences. This is due to different images being used to color high rise areas vs ground areas. This is very common with low altitude (drone) flights.

image

Mvs-texturing performs seam leveling on the textures atlases which addresses this issue at the mesh level, but does not help when the mesh is projected orthographically and viewed from above.

To identify such trouble areas, I perform a post-processing step after orthophoto rendering has completed to compute a gradient on the depthmap calculated during rendering (which is free, since we already compute it). Steep changes in elevation closely match object boundaries and it's trivial to apply a threshold to identify areas that belong to such edges:

image

image

image

From here I borrow from the existing body of research on image inpainting (http://www.olivier-augereau.com/docs/2004JGraphToolsTelea.pdf) which is able to fill in gaps in images and perform inpainting at the edges. The operation is executed on 1024x1024 chunks and multithreaded for fast and memory constrained execution.

image

The resulting orthophoto has significantly improved aliased edges which improve overall effective and perceived quality.

anim

anim2

The method is parameter free and generalizes to any georeferenced scene.


Saijin-Naib:
This is phenomenal!

PR #1646, opened 2023-05-04
Update DJI Mavic Mini v1 RSC

aeburriel:
This drone has different rolling shutter timings for 4/3 and 16:9 ratios.


pierotofy:
Thanks @aeburriel !

PR #1648, opened 2023-05-08
Direct odm_orthophoto georeferencing

pierotofy:
This PR adds support for direct raster georeferencing in the `odm_orthophoto` module, thus removing the need for an extra `gdal_translate` invocation.


pierotofy:
Interestingly, this does not have the improvement in runtime performance that I was expecting; most of the computation time seems to be spent inside GDAL for compression and tiling (not file I/O). It's still a bit faster, but not much. Oh well. Code looks cleaner on the Python side at least.

Hamza01541 (approved):

PR #1651, opened 2023-05-17
add opensfm arg: matching_order_neighbors

rexliuser:
Added "--matching_order_neighbors" argument for opensfm to allow faster feature matching if images are in sequence eg. extracted from video.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -153,6 +153,13 @@ def config(argv=None, parser=None):
                         default=0,
                         type=int,
                         help='Perform image matching with the nearest images based on GPS exif data. Set to 0 to match by triangulation. Default: %(default)s')
+    
+    parser.add_argument('--matcher-order-neighbors',
+                            metavar='<positive integer>',
+                            action=StoreValue,
+                            default=0,
+                            type=int,
+                            help='Perform image matching with the nearest images based on image order. Set to 0 to disable. Default: %(default)s')
 
     parser.add_argument('--use-fixed-camera-params',
                         action=StoreTrue,
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -246,6 +246,9 @@ class OSFMContext:
                 "triangulation_type: ROBUST",
                 "retriangulation_ratio: 2",
             ]
+            
+            if args.matcher_order_neighbors > 0:
+                config.append(f"matching_order_neighbors: {args.matcher_order_neighbors}")
 
             if args.camera_lens != 'auto':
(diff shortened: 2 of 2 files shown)


pierotofy:
Hey thanks for the PR @rexliuser, does the slowness happen when using the video processing functionality that's built-in into ODM? I wonder if we can simply set matching_order_neighbors to a good default value automatically instead of requiring the user to manually specify a value when video files are present.

rexliuser:
No. I tried the built-in video processing but I found it quite unreliable for my need, i.e. for certain videos, sometimes only <10 images or no images extracted at all, then the rest of ODM failed. So at last I resorted to my own methods to extract images from video before running ODM.

All in all, I'm feeding only jpg to ODM.

The slowness refers to when i use many images eg. >500 extracted from video myself, the FLANN matching is extremely slow as it does exponentially lots of unnecessary matching on far-away image pairs.

I did some online search and tested matching_order_neighbors to 2 and result is usually good. Though setting to 3-5 may be more robust for shaky videos and it does affect computation time by much. May need more trial on this.

pierotofy:
> sometimes only <10 images or no images extracted at all,

Huh, that's strange, would be interesting to understand and possibly fix this as well (are you able to share a test video that exhibits this issue?)

I wonder if we should then automatically add `matching_order_neighbors` to any dataset that lacks georeferencing information (default it to `3` or `5` as you suggested); slowness in reconstructing non-georeferenced data has been an issue for a while (not just for video).

pierotofy:
Nop, can't do automatic detection (it's just not reliable, as lots of non-georeferenced datasets are non-sequential in nature).

Thanks @rexliuser !

rexliuser:
> > sometimes only <10 images or no images extracted at all,
> 
> Huh, that's strange, would be interesting to understand and possibly fix this as well (are you able to share a test video that exhibits this issue?)
> 

I tried this video of an antique online: 
https://v-cg.etsystatic.com/video/upload/s--fNFl8qNK--/ac_none,c_crop,du_15,h_960,q_auto:good,w_720,x_0,y_209/5D1FD7E9-94BC-44B2-B90D-B358E262EF95_m3dpzf

I used docker, and it only extracted 2 frames from this 15s video, after that ODM failed with not enough images. I didn't dig into this, but i suspect it's related to each frame being too similar and didn't pass threshold and get rejected.

pierotofy:
:+1: I lowered the blur threshold and that seemed to get the dataset processed. I've opened #1655 to fix it.

PR #1653, opened 2023-05-18
Adds support for Zenmuse XT Thermal images

pierotofy:
Adds support for Zenmuse XT thermal images (with radiometric calibration).

image


pierotofy:
@holyCowMp3 while I was making these changes I noticed that https://github.com/OpenDroneMap/ODM/blob/master/opendm/thermal_tools/dji_unpack.py#L16 has the meta vars hard-coded, I don't have test images from an enterprise mavic 2 so I'm leaving them there, but note that we now have functions to extract those values (`extract_raw_thermal_image_data`). Perhaps even remove/condensate the Pillow code that unpacks the raw thermal image, since the exiftool function I wrote in this PR should be able to handle them.

If you have a dataset to share I can run some tests and update the code accordingly.

sbonaime:
@pierotofy Do you think thermal images from Zenmuse XT2 could be use with this commit ? How can I check ? 
Thanks

pierotofy:
Maybe, try it out and see if you get reasonable values.

PR #1660, opened 2023-05-24
max_tile safety check in DEM creation

pierotofy:
Sometimes the `commands.create_dem` function will hang forever until it crashes due to bad reconstructions causing the program to attempt to generate DEM rasters of huge dimensions.

This PR adds a conservative heuristic that blocks the process before it attempts to do the impossible.


pierotofy:
* Updates untwine to the latest version
* Embeds GCP info in GeoJSON format instead of GML to point cloud's VLR records

PR #1663, opened 2023-06-07
typo radiometric-calibration and  radiometric_calibration

sbonaime:
typo radiometric-calibration and  radiometric_calibration.


pierotofy:
Thanks @sbonaime :pray:, but I think `radiometric_calibration` is correct, it's not a typo. You would not be able to run code that has a property with a dash (`-`) in the name, as dashes are not allowed by the Python language for property/variable names.

You can verify this by building the code and trying to running the program with the proposed changes.

PR #1667, opened 2023-06-18
Added DJI Phantom4 rtk srt parsing

vinsonliux:
Added DJI Phantom4 rtk srt file parsing


pierotofy:
Thanks @vinsonliux ! Looks great.

PR #1671, opened 2023-06-30
Update rollingshutter.py

udaf-mcq:
This was done using Autel Evo II 8k camera at 8000 X 6000 pixels, shutter speed at 1/8000 second.

Photographing a flashing LED powered and controlled by an Arduino Nano  using code from WebODM rolling shutter site.


pierotofy:
Thanks @udaf-mcq !

PR #1676, opened 2023-07-15
update cuda ver

rexliuser:
update cuda image cuda:11.2.0 -> 11.2.2 in gpu.Dockerfile as original image no longer available.


pierotofy:
Thanks @rexliuser !

PR #1683, opened 2023-07-29
Update docs to reflect dem resolution defaults

smathermather:
Also ignore ignore-gsd, but also don't advertise it in orthophoto resolution. Replaces https://github.com/OpenDroneMap/docs/pull/176#issuecomment-1656550757


Saijin-Naib (approved):
Looks great to me.

Something about the existing help text for ignore-gsd in general doesn't read well to me, but I need to audit what it does to make sure I'm not just misunderstanding. It almost seems to imply that it would be a *good* option to pass for memory-limited systems, which it really is not.

Saijin-Naib (approved):
This, I think, reads much more true to the behavior and might help some folks not step into this particular stumbling block.

Saijin-Naib (approved):

PR #1697, opened 2023-09-08
Compress GCP data before VLR inclusion

### LINKED ISSUE #1691: Pdal is broken on big VLRs
Hello everyone,
ODM is broken on large GCP files. I used 17 gcp point and then error. But wehn i tried the without gcp it is working. 
image

I searched the error of the pdal. And i found it, it just supports for 65535 bytes for version between 1 and 1.3. It must be used v1.4 for large gcp files.
image

Pdal use version 1.2 as a default.
image

It can be changed in ODM to version 1.4 after some tests.

pierotofy:
Should fix #1691 by first compressing the GCP data, and then checking that the size is within margins for VLR inclusion.

### DIFF
--- a/stages/odm_georeferencing.py
+++ b/stages/odm_georeferencing.py
@@ -5,6 +5,7 @@ import pipes
 import fiona
 import fiona.crs
 import json
+import zipfile
 from collections import OrderedDict
 from pyproj import CRS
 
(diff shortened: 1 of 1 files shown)


DiegoContelli:
Hi my friend, thank you very much I will test it, it also worked here when I removed the --writers.las.vlrs parameter from the odm_georeferencing file in the WebODM\resources\app\apps\ODM\stages folder

PR #1701, opened 2023-09-29
Fix start/end date on Windows and enforce band order normalization

### LINKED ISSUE #1699: Fix capture start/end date on Windows
Discussion: https://community.opendronemap.org/t/capture-start-and-end-times/17666/5

pierotofy:
Fixes #1699


pierotofy:
This PR also enforces band order normalization, relying on the `band_index` provided by manufacturers provides inconsistent outputs, e.g. the Mavic 3M puts the RGB band at index 5, MicaSense at 1, plus countless other permutations found in the wild. It's a pain for users to then sort out why "RedEdge" is the first channel band when opened in QGIS (or worse, assume that ODM doesn't work because RGB is not the default band).

We now enforce this band order:

 * RGB
 * NIR
 * RedEdge
 * LWIR

Which is pretty common/standard.

Saijin-Naib:
Perfect. Follows the wavelength order too.

pierotofy:
This will also improve multispectral alignment, because in cases where the manufacturer chose "RedEdge" as index 0, RedEdge was automatically chosen as the primary band (unless overriden via `--primary-band`, which few people use), giving poor alignment results.

We now automatically favor RGB, or a Blue band if available, otherwise the first band in the list.

Saijin-Naib:
Can we favor green over blue for alignment? That should be higher SNR than blue on these sensors.

pierotofy:
Sure!

Saijin-Naib:
I don't know how to modify to test, but most sensors mimic human vision and green is the most responsive channel, plus slightly longer wavelength gives less noise on CCD and CMOS (blue is too close to UV), but not so long focusing and scatter becomes problematic (NIR/red-edge)

So, it should be slightly better but I can't prove it out 😑

PR #1702, opened 2023-10-03
Altum-PT support

### LINKED ISSUE #1686: Handle Altum-PT Thermal band
### How did you install ODM? (Docker, installer, natively, ...)?

Latest Docker image pulled from dockerhub.

### What is the problem?

Processing Altum-PT data with thermal band, when using ```--radiometric-calibration camera+sun```, software crashes.

Expect it to either leave this channel alone, or, convert the values from milliKelvin to Celsius.

In opendm/thermal.py in the dn_to_temperature function, change an "if" statement (line 38) so that it will handle both Altum models correctly:

```if photo.camera_make == “MicaSense” and photo.camera_model == “Altum”:```
to
```if photo.camera_make == “MicaSense” and “Altum” in photo.camera_model:```

This will solve this specific problem, adding support for Altum-PT.
To make the code more robust, the "else" case (line 57) for all unrecognized camera makes and models still is not functioning well and may need attention. (see attached log).

### How can we reproduce this? What steps did you do to trigger the problem? If this is an issue with processing a dataset, YOU MUST include a copy of your dataset AND task output log, uploaded on Google Drive or Dropbox (otherwise we cannot reproduce this).

Attached one Altum-PT thermal band image if you are able to test this function directly.
Otherwise full dataset available at https://shepard-altumpt-test.s3.us-west-2.amazonaws.com/  Ensure that channel 6 files have been removed before running, as they cause ODM to fail in a different manner.  Run equivalent of ```docker run -ti --rm -v C:/Users/scrawford/Documents/data:/datasets opendronemap/odm --project-path /datasets subset --dsm --dtm --pc-las --pc-quality high --radiometric-calibration camera+sun```

[IMG_0010_7.zip](https://github.com/OpenDroneMap/ODM/files/12371019/IMG_0010_7.zip)
[odm_crash_log.txt](https://github.com/OpenDroneMap/ODM/files/12371048/odm_crash_log.txt)

pierotofy:
Should close #1686

### DIFF
--- a/opendm/thermal.py
+++ b/opendm/thermal.py
@@ -35,12 +35,12 @@ def dn_to_temperature(photo, image, images_path):
         # Every camera stores thermal information differently
         # The following will work for MicaSense Altum cameras
         # but not necessarily for others
-        if photo.camera_make == "MicaSense" and photo.camera_model == "Altum":
+        if photo.camera_make == "MicaSense" and photo.camera_model[:5] == "Altum":
             image = image.astype("float32")
             image -= (273.15 * 100.0) # Convert Kelvin to Celsius
             image *= 0.01
             return image
-        elif photo.camera_make == "DJI" and photo.camera_model == "ZH20T":            
+        elif photo.camera_make == "DJI" and photo.camera_model == "ZH20T":
             filename, file_extension = os.path.splitext(photo.filename)
             # DJI H20T high gain mode supports measurement of -40~150 celsius degrees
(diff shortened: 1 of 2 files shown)


pierotofy:
Processed with `--radiometric-calibration camera+sun`:

image

pierotofy:
Thanks for providing a test dataset @gnosys-scrawford and suggestions for the fix!

PR #1708, opened 2023-10-14
Add extra report file op, disable snap builds

pierotofy:
Discussion on report issue: https://community.opendronemap.org/t/permission-denied-wont-write-report/17891

- Also disables Snap builds: https://community.opendronemap.org/t/how-many-people-use-snapcraft/17788


pierotofy:
Related: #1709

pierotofy:
* Removes WSL/WSL2 commands/instructions, which have been broken for years also. These were from a time before we had native Windows support.

John-Nagle:
Nobody used the snap version because the documentation didn't mention it prominently. It's easy to install. No need for docker. No need for admin access. Works fine on current Ubuntu. I'd like to have the snap version back.

smathermather:
@John-Nagle -- we need someone to take on the snap build for it to continue. We don't currently have a maintainer who can fix it when it periodically breaks, and we require that in order to maintain it as a build.

PR #1712, opened 2023-10-28
Adds support for DSP SIFT

### LINKED ISSUE #1710: Port DSP SIFT, remove min-num-features
DSP SIFT performs better than SIFT and is already implemented in other software like COLMAP. https://arxiv.org/pdf/1412.8556.pdf

It's just waiting to be ported.

We should also remove the concept of "min-num-features", and replace it with a "max-num-features" which sets an upper limit on the number of features. We currently iterate the feature extraction process by lowering the contrast thresholds to achieve a target number of features, but this doesn't necessarily find better features.

pierotofy:
Closes #1710, although leaves `min-num-features` in place for the time being.

### DIFF
--- a/SuperBuild/cmake/External-OpenSfM.cmake
+++ b/SuperBuild/cmake/External-OpenSfM.cmake
@@ -25,7 +25,7 @@ ExternalProject_Add(${_proj_name}
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
   GIT_REPOSITORY    https://github.com/OpenDroneMap/OpenSfM/
-  GIT_TAG           dsp
+  GIT_TAG           330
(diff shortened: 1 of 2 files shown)


pierotofy:
This descriptor/detector looks very promising, although it's slower by a factor of ~2x.

For comparison, this is the sparse point cloud graph / match graph for Brighton Beach processed with `--feature-type dspsift`:

image

image

Compared to SIFT:

image

image

pierotofy:
I also still need to port and test the affine shape estimation, I'm not sure if that will bring additional improvements.

So far it seems to be able to handle forest areas / poor overlap datasets with much better results. Would love for people to help test before I complete the rest and merge it.

pierotofy:
Closes #1713

pierotofy:
image

PR #1716, opened 2023-11-04
Fix fast_floor in FPC Filter, Invalid PLY file (expected 'property uint8 views')

pierotofy:
Fixes a fault in FPC Filter which was causing the program to stall forever in certain reconstructions when using `--pc-sample` , caused by a bad floor implementation.


pierotofy:
Also fixes the `Invalid PLY file (expected 'property uint8 views')` error which several people have encountered, due to a fault in ODM's calls to OpenMVS's subscene handling.

Saijin-Naib:
Do you think this is worthy of a new release of Windows? I think this is a nice fix up.

pierotofy:
Probably yes.

Issue #1718, opened 2023-11-07
Document appropriate rerun-from places when changing settings


smathermather:
Per: https://community.opendronemap.org/t/changed-flag-now-where-do-i-restart-from/18126, it's difficult for even some advanced users to know what point to re-run from when trying out different settings. Per João's request, we could establish this as a table, but I lean toward instead/also documenting this in the command documentation themselves so it's always visible.

smathermather:
https://github.com/OpenDroneMap/ODM/pull/1719

pierotofy:
| **Option**                              | **Restart Stage**     |
| --------------------------------------- | --------------------- |
| **3d-tiles**                            | PostProcess           |
| **align**                               | Georeferencing        |
| **auto-boundary**                       | Point Filtering       |
| **auto-boundary-distance**              | Point Filtering       |
| **bg-removal**                          | Load Dataset          |
| **boundary**                            | Point Filtering       |
| **build-overviews**                     | Orthophoto            |
| **camera-lens**                         | Load Dataset          |
| **cameras**                             | Load Dataset          |
| **cog**                                 | DEM                   |
| **copy-to**                             | PostProcess           |
| **crop**                                | Georeferencing        |
| **dem-decimation**                      | DEM                   |
| **dem-euclidean-map**                   | DEM                   |
| **dem-gapfill-steps**                   | DEM                   |
| **dem-resolution**                      | DEM                   |
| **dsm**                                 | DEM                   |
| **dtm**                                 | DEM                   |
| **end-with**                            | N/A                   |
| **fast-orthophoto**                     | Point Filtering       |
| **feature-quality**                     | Structure From Motion |
| **feature-type**                        | Structure From Motion |
| **force-gps**                           | Structure From Motion |
| **gcp**                                 | Load Dataset          |
| **geo**                                 | Load Dataset          |
| **gltf**                                | Texturing             |
| **gps-accuracy**                        | Load Dataset          |
| **help**                                | N/A                   |
| **ignore-gsd**                          | Structure From Motion |
| **matcher-neighbors**                   | Structure From Motion |
| **matcher-order**                       | Structure From Motion |
| **matcher-type**                        | Structure From Motion |
| **max-concurrency**                     | N/A                   |
| **merge**                               | Merge                 |
| **mesh-octree-depth**                   | Meshing               |
| **mesh-size**                           | Meshing               |
| **min-num-features**                    | Structure From Motion |
| **no-gpu**                              | N/A                   |
[...]

PR #1719, opened 2023-11-08
Rerun from notes

smathermather:
Per https://github.com/OpenDroneMap/ODM/issues/1718, an attempt to document appropriate rerun-from places.

### DIFF
--- a/opendm/config.py
+++ b/opendm/config.py
@@ -248,7 +248,7 @@ def config(argv=None, parser=None):
                     action=StoreTrue,
                     nargs=0,
                     default=False,
-                    help='Use a full 3D mesh to compute the orthophoto instead of a 2.5D mesh. This option is a bit faster and provides similar results in planar areas. Default: %(default)s')
+                    help='Use a full 3D mesh to compute the orthophoto instead of a 2.5D mesh. This option is a bit faster and provides similar results in planar areas. (If rerunning, restart "From Orthophoto") Default: %(default)s')
 
     parser.add_argument('--skip-3dmodel',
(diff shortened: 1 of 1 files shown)


smathermather:
Some of these are known to be correct, some are best guesses. Not all flags are noted, especially those that would require a complete or nearly complete rerun of the toolchain.

pierotofy:
This seems a bit verbose. In my opinion this belongs to the documentation.

p.s. the missing guide 2nd has a table with all the rerun stages; feel free to double check from there.

But the real issue is: people don't know/remember which stage to restart from if they change an option.

Ideally, ODM would already know which settings were used in the previous run (we currently don't), compare the previous with the current options, and automatically pick the restart point.

E.g.

`./run.sh /datasets/brighton`

`./run.sh /datasets/brighton --orthophoto-resolution 2` --> rerun from orthophoto stage without needing to specify: `--rerun from odm_orthophoto`. That would remove the need to know the stages.

smathermather:
I don't agree on verbosity (although suggestions for how to shorten and remain clear quite welcome), and agree wrt docs, but if implemented here it is in the docs and always in the docs, rather than drifting apart from the docs.

Agree on functionality, but that's a different pull request altogether. Happy to make this obsolete with such a pull request in the future (should be easier now that we have json to parse), but not today.

Folks regularly ask for this info. I'd love for it to be in docs and up-to-date with the code by default.

(Edit: I'll edit from that table. My copy happily remained in Zambia with Dr. Samboko. I was too lazy this evening to dig up my digital copy but can review tomorrow or Thursday.)

PR #1724, opened 2023-11-19
Feat: track decimation

pierotofy:
This PR brings the changes suggested from https://github.com/mapillary/OpenSfM/pull/641 into ODM.

On small datasets this seems to decrease relative accuracy, so perhaps this might need to be adjusted to only be used on larger datasets (how large?)

Please help test, in particular:
 * How does relative/absolute accuracy change relative to the current baseline?
 * How do the final outputs look relative to the current baseline?
 * What are the runtime differences for the OpenSfM stage?

### DIFF
--- a/SuperBuild/cmake/External-OpenSfM.cmake
+++ b/SuperBuild/cmake/External-OpenSfM.cmake
@@ -24,8 +24,8 @@ ExternalProject_Add(${_proj_name}
   STAMP_DIR         ${_SB_BINARY_DIR}/stamp
   #--Download step--------------
   DOWNLOAD_DIR      ${SB_DOWNLOAD_DIR}
-  GIT_REPOSITORY    https://github.com/OpenDroneMap/OpenSfM/
-  GIT_TAG           330
+  GIT_REPOSITORY    https://github.com/pierotofy/OpenSfM/
+  GIT_TAG           trackdec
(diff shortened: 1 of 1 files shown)


smathermather:
What sizes do you want / need tested?

pierotofy:
A variety would be best (small, medium, large), but the main concern is just trying to answer the bullet points I mentioned above.

smathermather:
Cool. I've got it running on a 1900 dataset. Seems like a decent size (larger than most test datasets) without taking overly long to run (should be done tomorrow afternoon). All defaults, one running on current and one on a fresh build of this branch.

smathermather:
So far, I don't think I see any computational time differences in OpenSfM:
image

smathermather:
Only difference I see so far is progress bar differences for the same stage:
image

smathermather:
Definitely got some weird errors in the northeastern corner:

[Kakuma_17A_Flight_02-report.pdf](https://github.com/OpenDroneMap/ODM/files/13432151/Kakuma_17A_Flight_02-report.pdf)
[Kakuma_17A_Flight_02-trackdec-report.pdf](https://github.com/OpenDroneMap/ODM/files/13432152/Kakuma_17A_Flight_02-trackdec-report.pdf)


image

trackdec left, current right

image

PR #1725, opened 2023-11-27
Render DEM tiles using RenderDEM

### LINKED ISSUE #1722: Limit number of parallel pdal processes for DEM generation
Sometimes launching DEM generation with PDAL results in an out-of-memory error because we launch too many processes at once:

```
[INFO]    Point cloud bounds are [minx: 373925.68, maxx: 377208.26] [miny: 3858571.96, maxy: 3861177.08]
[INFO]    DEM resolution is (130257, 164130), max tile size is 4096, will split DEM generation into 121 tiles
...
Killed
```

Even though a computer is capable of processing the dataset.

pierotofy:
This PR replaces the PDAL calls to generate DEM tiles with a separate C++ module ([RenderDEM](https://github.com/OpenDroneMap/RenderDEM)), which performs the same task using PDAL internals, but more efficiently. Because it performs point cloud reads only once, it reduces memory usage by an order of magnitude, while speeding up processing by at least 20% (often more, depends on number of cores, size of point cloud).

Closes #1722

When reading PLY files it also skips the PDAL reader and reads them using an optimized PLY reader (same as OpenPointClass), which is orders of magnitude faster (useful for the 2.5D mesh generation part).


smathermather:
Oh, heck yeah.

PR #1731, opened 2023-12-07
C++ median smoothing filter

pierotofy:
This PR moves the median smoothing filter to C++ via FastRasterFilter (https://github.com/OpenDroneMap/FastRasterFilter), which is now both faster and more algorithmically correct, since we properly apply the median filter using a sliding window with padding (compared to the current implementation, which does not implement padding).

image

### DIFF
--- a/SuperBuild/CMakeLists.txt
+++ b/SuperBuild/CMakeLists.txt
@@ -251,6 +251,17 @@ externalproject_add(odm_orthophoto
                       ${WIN32_CMAKE_ARGS} ${WIN32_GDAL_ARGS}
 )
 
+externalproject_add(fastrasterfilter
+    DEPENDS           eigen34
+    GIT_REPOSITORY    https://github.com/OpenDroneMap/FastRasterFilter.git
+    GIT_TAG           main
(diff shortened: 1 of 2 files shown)


pierotofy:
On brighton beach for a small raster this is ~25% faster, I fully expect this to be much faster for larger rasters and orders of magnitude faster for corridor type datasets, which have large chunks of empty space (we now detect empty tiles and skip the filtering for those areas).

pierotofy:
Oh yeah.

```
time ./fastrasterfilter "/datasets/large/mesh_dsm.tif" --output "/datasets/large/median.tif" --window-size 512 --radius 5 --co TILED=YES --co BIGTIFF=IF_SAFER --co COMPRESS=DEFLATE
Input: /datasets/large/mesh_dsm.tif
Size: 45999x96885
Blocks: 17100
Smoothing...1000...2000...3000...4000...5000...6000...7000...8000...9000...10000...11000...12000...13000...14000...15000...16000...17000...17100... done
Wrote /datasets/large/median.tif

real	5m47.263s
```

On a scrappy core i5 laptop (4 cores).

PR #1736, opened 2024-01-10
modify upper limit of band count from 8 to 10

joanlyq:
MicaSense RedEdge-MX/Dual sensors has 10 bands usually. 

Based on modifying files in docker container and the two posts in the community, it seems to work fine just by changing the upper limit. 

https://community.opendronemap.org/t/problems-with-multispectral-10-band-16-bit-images/14718

https://community.opendronemap.org/t/problems-while-processing-micasense-rededge-mx-dual-data-10-bands-with-webodm/11622

### DIFF
--- a/contrib/types.py
+++ b/contrib/types.py
@@ -1,0 +1,468 @@
+import os
+import shutil
+import warnings
+import numpy as np
+from opendm import get_image_size
+from opendm import location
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @joanlyq , I have merged this contribution as part of #1738 :pray:

Issue #1737, opened 2024-01-13
Skip missing multi-band images instead of failing


pierotofy:
Currently if a multispectral dataset lacks all the images for a band, the reconstruction will fail:

```
[ERROR]   Multi-camera setup detected, but band "RedEdge" (identified from "IMG.TIF") has only 54 images (instead of 53), perhaps images are missing or are corrupted. Please include all necessary files to process all bands and try again.
[INFO]    Found 213 usable images
```

We should make it so that the reconstruction continues, emit a warning, and discard the files with missing bands instead.

smathermather:
This is a huge usability improvement for multispectral users. +1

PR #1738, opened 2024-01-15
Ignore multispectral band groups that are missing images

pierotofy:
Closes #1736, #1737

If one or more images are missing, the console will show something like:

```
[INFO]    Found 29 usable images
[WARNING] Multi-camera setup detected, but band "Blue" (identified from "IMG_0278_1.tif") has only 5 images (instead of 6), perhaps images are missing or are corrupted.
[WARNING] Excluding IMG_0277_5.tif
[WARNING] Excluding IMG_0277_3.tif
[WARNING] Excluding IMG_0277_2.tif
[WARNING] Excluding IMG_0277_4.tif
[INFO]    New image count: 25
```

### DIFF
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -64,7 +64,6 @@ class OSFMContext:
                             "Check that the images have enough overlap, "
                             "that there are enough recognizable features "
                             "and that the images are in focus. "
-                            "You could also try to increase the --min-num-features parameter."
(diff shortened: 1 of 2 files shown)


smathermather:
Lots of much happier multispectral users with cameras that struggle to write to their cards fast enough are collectively breathing a sigh of relief.

PR #1744, opened 2024-02-01
Update comment on mvs_tmp_dir

kielnino:
Hi, I'm brand new to this project, so I'm not sure if I've understood the code correctly. But going through it, I noticed that the mvs_tmp_dir is defined but never used.

### DIFF
--- a/stages/mvstex.py
+++ b/stages/mvstex.py
@@ -108,6 +108,13 @@ class ODMMvsTexStage(types.ODM_Stage):
                     'labelingFile': '-L "%s"' % r['labeling_file'] if r['labeling_file'] else ''
                 }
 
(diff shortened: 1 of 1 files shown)


kielnino:
Oh maybe its just not documented exactly.
I think the [texrecon](https://github.com/OpenDroneMap/mvs-texturing/blob/c831cf75db754e2e8288a51d0a44f43b8824fe7a/apps/texrecon/texrecon.cpp#L51) creates the tmp directory, so its actually needed to check for it.

kielnino:
I have updated the comment accordingly. If this change is negligible, I'm fine if this PR is closed.

pierotofy:
Thanks for the PR! texrecon does create the directory, but will stop execution if a directory with files already exists (that's why we make sure to remove it). The comment does add clarity to the purpose.

PR #1746, opened 2024-02-09
GLTF - obj2glb - Set extensionsUsed in all cases to be consistent with the GLTF standard

kielnino:
As the GLTF Definition [states](https://github.com/KhronosGroup/glTF/blob/main/extensions/README.md#extension-mechanics)

> All extensions used in a model are listed as strings in the top-level extensionsUsed array.

So `extensionsUsed` should be set in all cases. With this change `extensionsUsed` is set even if `rtc` is `(None, None)` (which is the default).

In our use case we didn't had georeferences so `rtc` was not set and the created `glb` file couldn't be loaded with the three.js GLTFLoader because `extensionsUsed` was not set.

### DIFF
--- a/opendm/gltf.py
+++ b/opendm/gltf.py
@@ -279,9 +279,10 @@ def obj2glb(input_obj, output_glb, rtc=(None, None), draco_compression=True, _in
     )
 
     gltf.extensionsRequired = ['KHR_materials_unlit']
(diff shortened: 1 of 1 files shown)


pierotofy:
Looks great, thanks @kielnino !

PR #1756, opened 2024-04-18
Fix PoissonRecon failed with n threads log message

andrewharvey:
The message was reporting failure with `n` threads and retrying with `n // 2`, however a few lines up `threads` was already set to `n // 2` representing the next thread count to try.

You'll see currently the log finishing with

```
[WARNING] PoissonRecon failed with 1 threads, let's retry with 0...
```

which demonstrates the issues.


pierotofy:
Nice catch @andrewharvey ! Thanks for the patch. 🙏

Issue #1757, opened 2024-05-09
DJI Mavic 2 Zoom video (mp4 + srt) processing problem. Missing longitude and latitude in the produced jpg exif data.


idimitrovski:
### How did you install ODM? (Docker, installer, natively, ...)?

Installer (Windows 10)

### What is the problem?

When processing mp4 files from DJI Mavic 2 Zoom drone, the first step of the processing produces jpg files without lon, lat exif data. This is due to non standard srt file structure (the longitude and latitude params in the srt file are different from the supported formats)

I already added support for the Mavic 2 Zoom srt files in the code. I just opened this issue so I can reference it in the pull request.

PR #1758, opened 2024-05-10
Support for DJI Mavic 2 Zoom srt files

idimitrovski:
This is a small fix regarding issue #1757 .  DJI Mavic 2 Zooms firmware has different srt file format format than other models. Also (since forever) it has a typo in the "longitude" property in the srt file so the files are generated with "longtitude" property. They haven't fixed it since the release of the drone so I doubt they ever will. This issue has been reported to them since 2019...


pierotofy (approved):

pierotofy:
Looks great! Thanks @idimitrovski !

Issue #1761, opened 2024-05-27
Video to image conversion GPS tagging stops working after a while


rawnsley:
### How did you install ODM? (Docker, installer, natively, ...)?

Running through Docker on MacOS

### What is the problem?

Input is a video from a DJI drone where I have extracted the embedded SRT stream into companion file. The video conversion creates frames for the whole video correctly and adds GPS EXIF data for the first 39 frames, but then doesn't it for any more frames after that. This corresponds to frame 410 in the video and there is nothing obviously unusual at that point in the SRT file.

### What should be the expected behavior?

All frames should be annotated with GPS EXIF data

### How can we reproduce this?

Command line:

```sh
docker run -ti --rm -v /my/projects/datasets opendronemap/odm --project-path /datasets test3
```
Output:

```
[INFO]    Initializing ODM 3.5.1 - Mon May 27 07:09:37  2024
[INFO]    ==============
[INFO]    3d_tiles: False
[INFO]    align: None
[INFO]    auto_boundary: False
[INFO]    auto_boundary_distance: 0
[INFO]    bg_removal: False
[INFO]    boundary: {}
[INFO]    build_overviews: False
[INFO]    camera_lens: auto
[INFO]    cameras: {}
[INFO]    cog: False
[INFO]    copy_to: None
[INFO]    crop: 3
[INFO]    dem_decimation: 1
[INFO]    dem_euclidean_map: False
[INFO]    dem_gapfill_steps: 3
[INFO]    dem_resolution: 5
[INFO]    dsm: False
[INFO]    dtm: False
[INFO]    end_with: odm_postprocess
[INFO]    fast_orthophoto: False
[INFO]    feature_quality: high
[INFO]    feature_type: dspsift
[INFO]    force_gps: False
[INFO]    gcp: None
[INFO]    geo: None
[INFO]    gltf: False
[INFO]    gps_accuracy: 3
[INFO]    ignore_gsd: False
[INFO]    matcher_neighbors: 0
[INFO]    matcher_order: 0
[INFO]    matcher_type: flann
[INFO]    max_concurrency: 10
[INFO]    merge: all
[INFO]    mesh_octree_depth: 11
[INFO]    mesh_size: 200000
[INFO]    min_num_features: 10000
[INFO]    name: test3
[INFO]    no_gpu: False
[INFO]    optimize_disk_space: False
[INFO]    orthophoto_compression: DEFLATE
[INFO]    orthophoto_cutline: False
[...]

PR #1765, opened 2024-06-10
Speed-up undistortion by 2x

pierotofy:
Speeds up the undistortion step by at least 2x, using caching to avoid recomputing camera mappings.

https://github.com/OpenDroneMap/OpenSfM/commit/5c9fe5be9f5a0d0f18af18167de6d130e995a931


pierotofy:
- Exports the camera ID(s) with shots.geojson.
- Exports camera UV undistortion mappings (added in the 352 branch). These can be used by downstream applications to undo the image undistortion (for example, to map points from a reconstruction camera back to the original images). Description of the format is in the OpenSfM source.

pierotofy:
Improves #1761

PR #1774, opened 2024-07-05
Fixed GPU docker build

### LINKED ISSUE #1773: Docker GPU image not up to date due to failed Guthub Action
I've noticed that the Docker GPU image had failed to build and publish since 3.4.0 and so is now out of date with the master branch.

This was the PR that seemed to cause the build to first fail https://github.com/OpenDroneMap/ODM/actions/runs/7846051413

ehallein:
missing libav libraries, install ffmpeg to fix
Fixes #1773

### DIFF
--- a/.github/workflows/publish-docker-gpu.yaml
+++ b/.github/workflows/publish-docker-gpu.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     timeout-minutes: 2880
     steps:
     - name: Checkout
--- a/gpu.Dockerfile
+++ b/gpu.Dockerfile
@@ -21,7 +21,7 @@ RUN bash configure.sh clean
 
 ### Use a second image for the final asset to reduce the number and
 # size of the layers.
-#FROM nvidia/cuda:11.2.2-runtime-ubuntu20.04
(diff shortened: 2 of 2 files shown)


pierotofy:
Looks good to me, thank you @ehallein !

ehallein:
seems to have fixed the gpu building but the normal build failed now?https://github.com/OpenDroneMap/ODM/actions/workflows/publish-docker.yaml
Which is odd because those files haven't changed. Are you able to run publish-docker.yaml again @pierotofy to see if it still fails. its using "self-hosted", so not sure if there is some issues there?  Also odd that those builds have been taking 14 hours to run?

pierotofy:
The failure is unrelated to this PR. 

`2024-07-06T07:15:24.8132346Z #24 7.579 Installing lsb_release
2024-07-06T07:16:03.9635500Z #24 46.72 E: Failed to fetch http://old-releases.ubuntu.com/ubuntu/pool/main/r/readline/readline-common_8.1-1_all.deb  Connection timed out [IP: 185.125.190.40 80]`

Just a hiccup with Ubuntu's servers (happens quite a lot actually).

It will probably build fine at the next rebuild. 14 hours is unfortunately how long it takes to build the images (and reason we have to use a self-hosted runner) because we build both amd64 and arm images, and ARM compilation happens via emulation, which is super-slow.

PR #1775, opened 2024-07-07
Update OpenStreetMap provider in gdal2tiles.py

danieldegroot2:
I would like to suggest using the now preferred https://tile.openstreetmap.org URL instead of the current one (`{s}.`), see

https://github.com/openstreetmap/operations/issues/737

Also, changes attribution URL to HTTPS link (use direct links).


pierotofy:
Thanks @danieldegroot2

PR #1779, opened 2024-07-16
Add libtbb2 to gpu.Dockerfile

fri-K:
libtbb2 is need for gpu docker container
see nodeodm bug #219


pierotofy:
Nice, this needs confirmation that it works correctly with CUDA before merging.

ehallein:
Tested and builds and runs fine. However this bug only seemed to appear when using webODM/nodeODM, as it ran fine using ODM directly, so not sure if fixing it here will solve the downstream issue in NodeODM?

pierotofy:
Thanks for testing and confirming! 

It might be an issue with cgroups and docker-compose, in which case perhaps it's an issue with how WebODM brings the node online.

I'll merge this for now since it should fix ODM.

Thank you!

ehallein:
ok, once the nodeODM image is built I'll test it

ehallein:
tested with odm_data_aukerman and it's now working on nvidia GPU in webODM

PR #1782, opened 2024-07-22
Update logic to allow split default when path set

smathermather:


pierotofy:
Thanks, I've fixed the default use case (where `split-image-groups` is set to None).

Compare the fix to the original proposed change, test, then feel free to merge. :clinking_glasses:

smathermather:
Make sense, thanks! Testing now.

smathermather:
Tested and merged.

PR #1788, opened 2024-08-07
Las resolution tied to GSD

smathermather:


stages/odm_georeferencing.py
```diff
@@ -125,13 +125,26 @@ def process(self, args, outputs):
 
                 stages.append("transformation")
                 utmoffset = reconstruction.georef.utm_offset()
+
+                # Establish appropriate las scale for export
+                las_scale = 0.001
+                filtered_point_cloud_stats = tree.path("odm_filterpoints", "point_cloud_stats.json")
+                if os.path.isfile(filtered_point_cloud_stats):
+                    with open(filtered_point_cloud_stats, 'r') as stats:
```
pierotofy: * Would place any file I/O operation inside a try//catch block.
 * We should check for a threshold before triggering a change in scale (currently this will change every time)
 * Spacing should be in 1/10 increments, not `spacing / 10`. E.g. 0.001, 0.0001, 0.00001, etc. not 0.35893904

It's a good start!

smathermather: I addressed the first and the third in an update to the pull request. Sorry/not sorry for the function name. Probably doesn't need to be a function, but at least then it's somewhat clear what it's doing.

For the second one, I'm not sure what we are checking for. Do you mean on a rerun from an earlier stage or something else? I haven't profiled it, but relative to (waves hand at all the photogrammetry), I suspect this calculation is trivially inexpensive to recalculate if recalculation happens (I cannot detect the stage running slower than stock). But, as unsure I am what threshold we are checking for, I'm writing here largely from ignorance.

But, happy to address if I understand the concern. And thanks for the quick and thorough review!

smathermather: I should add, I have no idea if there's a stock function to do the log rounding. My googling on the issue was unproductive. It seems like it should be a common function type and have a common name, but it was easier to write than search for.

pierotofy: For the second one it's fine, you currently do a call to `min`, which does that implicitly, I would have done an `if threshold then ...` but that's just me. The end result is equivalent. Thanks!

smathermather: Makes sense. An if would be more explicit indeed.

Issue #1793, opened 2024-08-13
Feature suggestion: adding TIFFTAG_DATETIME to .tif outputs


stephenwinn16:
Hi, 

I'm looking into adding some metadata to the `.tif` outputs of ODM such that I have access to this information in a self-contained manner. Specifically, I would like the file metadata to include some estimate of the capture time (e.g. the mean of capture start/capture end times). I see [here](https://gdal.org/drivers/raster/gtiff.html) that geotiff files support the TIFFTAG_DATETIME tag (I use this info to e.g. get the capture time of satellite data I'm working with). Do you think it would be a case of adding an `-mo` argument to the GeoTIFF creation stage e.g. as is done below for the `-co` arguments? 

https://github.com/OpenDroneMap/ODM/blob/e6e4095844017d4f84d7d845e4d14780059703f6/stages/odm_orthophoto.py#L102

As well as something similar for the DSM/DTM creation steps if they happen. I'm happy to take a shot at adding this to the source code if I'm on the right track here.

pierotofy:
I think that could be useful and a pull request would be most welcome.

stephenwinn16:
Implemented in PR #1797.

PR #1797, opened 2024-08-20
Adding TIFFTAG_* to .tif outputs

stephenwinn16:
Following up on  #1793, I've added steps to the orthophoto and DSM/DTM generation to be able to include TIFFTAG information. In the current commit, I've hardcoded a datetime for testing purposes. I would like this to be replaced by a mean capture time based on the input image metadata (such as the "Capture Start" and "Capture End" that appear e.g. in the PDF report). Would anyone be able to point me to a good place to pull them from during the processing pipeline? 

Additionally, is the ODM version stored somewhere during the processing pipeline? Or should it always be read from the `VERSION` file as in `run.py`?

### DIFF
--- a/opendm/dem/utils.py
+++ b/opendm/dem/utils.py
@@ -1,4 +1,3 @@
-import rasterio
 
 def get_dem_vars(args):
     return {
@@ -9,10 +8,3 @@ def get_dem_vars(args):
         'BIGTIFF': 'IF_SAFER',
         'NUM_THREADS': args.max_concurrency,
     }
-
-def update_tags(dem_file):
-
-    # - Update the geotiff tags in place using rasterio
-    with rasterio.open(dem_file, 'r+') as rst:
-        rst.update_tags(TIFFTAG_DATETIME='2024:01:01 00:00+00:00')
-        rst.update_tags(TIFFTAG_SOFTWARE='OpenDroneMap')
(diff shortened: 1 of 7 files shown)


pierotofy:
Thanks @stephenwinn16,  note we should probably set the tags in/around https://github.com/OpenDroneMap/ODM/blob/master/stages/odm_postprocess.py#L38 similarly to how we set GCPs metadata.

You could find the precomputed start/end dates in `odm_report/stats.json`, but you have to be careful, because if a report is not generated (when `--skip-report` is used) the file might not be there. If not, you'll have to recompute from utc_time https://github.com/OpenDroneMap/ODM/blob/master/opendm/photo.py#L130 as a fallback.

Yes version is stored in `VERSION`. I would just re-read it, or create a function to read it and then optionally memoize it.

stephenwinn16:
Thanks @pierotofy for the comments. I've moved the lines into the `odm_postprocess` step and read the time information either from the `odm_report/shots.geojson` file or the `images.json` file depending on what exists. 

Do you see any other useful information to include with the [TIFFTAG fields](https://gdal.org/en/latest/drivers/raster/gtiff.html)?

pierotofy:
Thanks @stephenwinn16, I've made some changes to simplify the logic and avoid reading files from disk, I realized we already had a function to read the ODM version and that it was probably faster to read the UTC time from the photos object which is already in memory, I'll run some more tests and merge this shortly. :pray:

stephenwinn16:
Great, more efficient and much cleaner, thank you @pierotofy !

PR #1799, opened 2024-09-09
Include info on CODEM 

smathermather:


pierotofy:
Shouldn't credits be added in the docs/README ?

I just don't think the command line argument are the right place to credit.

smathermather:
https://github.com/OpenDroneMap/docs/pull/199

PR #1801, opened 2024-09-16
Fix incorrect file name in ODM_WARNING

ghost:
post_point_cloud_steps() incorrectly reports xyz file name when las file name is intended.


pierotofy:
Nice catch, thanks @mark-stevens !

PR #1811, opened 2024-11-06
Aligned the arm64 release name

HeDo88TH:
The Obj2Tiles arm release name now follows the same naming convention


cjgarson:
@HeDo88TH The build will fail on this unless the Obj2Tiles version is also changed to v1.0.13 or some logic is placed to account for the difference in versions/release tags, since v1.0.12 uses LinuxArm. Not sure how you want to handle--or if there will be any issues using the new version.

pierotofy:
Looks good to me!

PR #1828, opened 2025-02-18
added capability to generate vector polygon in DXF format to be used in AutoCAD LT

### LINKED ISSUE #1825: Create border polygon
During a course students told me that in CAD system to show correctly orthophoto it is needed a file with a polygon of the orthophoto's bounding box (just the extension). They also told me that proprietary software is creating this polygon and provide it with the image.

I searched a little bit but it looks like there is nothing like that right now, I can implement it, I was thinking to create a new function (and a new option in the cli) and add it after `args.orthophoto_cutline`, 

What do you think?

lucadelu:
Fix #1825

Do you need any test for this?

### DIFF
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -91,7 +91,7 @@ def generate_extent_polygon(orthophoto_file, output_file=None):
 
     Args:
         orthophoto_file (str): the path to orthophoto file
-        output_file (str, optional): the path to the gpkg file. Defaults to None.
+        output_file (str, optional): the path to the output file. Defaults to None.
     """
     def _create_vector(ortho_file, poly, format, output=None):
         if output is None:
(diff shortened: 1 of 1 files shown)


pierotofy:
It looks good, thanks @lucadelu 

Can you please describe in more details the use case for needing this extent file? I don't have a problem merging it, but I would like to understand which use case we're solving (the information is redundant, so why is it needed?)

lucadelu:
It was a request of a client, for what they told me, to get the right positioning and resolution of a raster in a CAD system they need to upload a vector layer and later tell to the raster to adapt to the vector polygon loaded before.

Since they would like to have DXF, in the last commit I also added DXF output, the sum of the two is something like 120Kb and it is really a fast operation.

pierotofy:
Ok, I understand, but should ODM be in charge of generating redundant files for an unspecified CAD software?

It's a rhetorical question, I'd be nice to hear @smathermather's thoughts on this.

smathermather:
I've been following, and as far as the geopackage with the extent, I think we need to make explicit what CAD package we are supporting for 2 reasons: users of that CAD package should know we support them and we should be at least announcing that support in the pull request and thus the releases; and we should know we are supporting them so that when we get bug reports later because we changed something that breaks that support, we will want to know where those questions are coming from and why. Maybe we can even prevent those breaks from happening, if there is some downstream community engagement.

As far as the DXF file, that feels like custom code bolted on to a core project, and don't feel inclined to support its inclusion. I could be swayed, but from what little signal I get in this request, I don't even get the sense the upstream customer knows why they want it, which definitely sets off my `this is custom code` warning flags.

lucadelu:
@smathermather ok I understood, what is not clear to me is, geopackage is it fine and not the DXF or you are incline to not support the inclusion of both of the output?

smathermather:
I would just say: choose one and let us know who we'll now be supporting as a CAD community. Choose the one that better serves your use case, and anyone who needs the data in an additional format will need to post-process.

lucadelu:
> I would just say: choose one and let us know who we'll now be supporting as a CAD community. Choose the one that better serves your use case, and anyone who needs the data in an additional format will need to post-process.

Done. Let me know if need anything else...

Saijin-Naib:
Knew it was AutoCAD 🤣

pierotofy:
:+1:, thanks for the changes and information @lucadelu. 

I'm still not 100% clear on what software we are supporting here; AutoCad is a suite of tools. Is this Civil 3D? Map 3D? LT? Please provide some more details, or better, have the client join the discussion.

I want to make sure we are solving the problem with the correct file before merging this. Other photogrammetry software for example provides TFW files. Is that the file we should be providing instead?

Related discussions: 

https://community.opendronemap.org/t/tiff-importing-to-autocad/15426/9
https://community.opendronemap.org/t/output-for-orthomosaic-on-jpw-for-georeferencing-on-cad/18972
https://community.opendronemap.org/t/how-to-export-a-coordinate-file-with-orthophotos-jgw/14333

lucadelu:
@pierotofy AutoCAD LT, I changed also the title of the pull request.

Maybe also TFW could be useful, as well .wld for png file (I have in mind to add this to PNG export but first let finish this one). They are using another proprietary photogrammetry software but they never spoke with me about TFW file.

pierotofy:
Thanks @lucadelu. Yes additional files can be considered as part of a separate PR. 👍

lucadelu:
Hi @pierotofy I have update my WebODM instance to last ODM and WebODM versions but DXF output is not in the output folder (the orthophoto is there). Is there anything else to do?

pierotofy:
You might need to include the path into https://github.com/OpenDroneMap/NodeODM/blob/master/libs/Task.js#L444 as well for it to show in WebODM's all assets archive.

lucadelu:
@pierotofy ok thanks, now it's clear.
Do you think it could be useful for everyone or you suggest to keep the change only in my running instance?

Saijin-Naib:
I think it might be welcome

PR #1837, opened 2025-03-05
added wld file to png and tfw file to geotiff

lucadelu:
Let's discuss about this changes, I saw several requests on the forum about these files, since it is not time and space consuming I think they are good improvements for ODM outputs.

### DIFF
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -161,24 +161,16 @@ def generate_extent_polygon(orthophoto_file, output_file=None):
 def generate_tfw(orthophoto_file):
     base, ext = os.path.splitext(orthophoto_file)
     tfw_file = base + '.tfw'
-    with rasterio.open(orthophoto_file) as ds:
-        transform = ds.transform
-        with open(tfw_file, 'w') as f:
-            # rasterio affine values taken by
-            # https://mharty3.github.io/til/GIS/raster-affine-transforms/
-            # pixel x size
-            f.write("%s\n" % transform.a)
-            # rotation y
-            f.write("%s\n" % transform.d)
-            # rotation x
-            f.write("%s\n" % transform.b)
-            # pixel y size (negative)
-            f.write("%s\n" % transform.e)
-            # x coordinate
-            f.write("%s\n" % transform.c)
-            # y coordinate
-            f.write("%s\n" % transform.f)
-    return True
+
+    try:
+        with rasterio.open(orthophoto_file) as ds:
+            t = ds.transform
+            with open(tfw_file, 'w') as f:
+                # rasterio affine values taken by
+                # https://mharty3.github.io/til/GIS/raster-affine-transforms/
+                f.write("\n".join([str(v) for v in [t.a, t.d, t.b, t.e, t.c, t.f]]) + "\n")
+    except Exception as e:
+        log.ODM_WARNING("Cannot create .tfw for %s: %s" % (orthophoto_file, str(e)))
 
 
 def post_orthophoto_steps(args, bounds_file_path, orthophoto_file, orthophoto_tiles_dir, resolution):
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -103,59 +103,51 @@ def generate_kmz(orthophoto_file, output_file=None, outsize=None):
     system.run('gdal_translate -of KMLSUPEROVERLAY -co FORMAT=PNG "%s" "%s" %s '
                '--config GDAL_CACHEMAX %s%% ' % (orthophoto_file, output_file, bandparam, get_max_memory()))    
 
-def generate_extent_polygon(orthophoto_file, output_file=None):
+def generate_extent_polygon(orthophoto_file):
     """Function to return the orthophoto extent as a polygon into a gpkg file
 
     Args:
         orthophoto_file (str): the path to orthophoto file
-        output_file (str, optional): the path to the output file. Defaults to None.
     """
-    def _create_vector(ortho_file, poly, format, output=None):
-        if output is None:
-            base, ext = os.path.splitext(ortho_file)
-            output_file = base + '_extent.' + format.lower()
-        # set up the shapefile driver
-        driver = ogr.GetDriverByName(format)
-
-        # create the data source
-        ds = driver.CreateDataSource(output_file)
-
-        # create one layer
-        layer = ds.CreateLayer("extent", srs, ogr.wkbPolygon)
-        if format != "DXF":
-            layer.CreateField(ogr.FieldDefn("id", ogr.OFTInteger))
-
-        # create the feature and set values
-        featureDefn = layer.GetLayerDefn()
-        feature = ogr.Feature(featureDefn)
-        feature.SetGeometry(ogr.CreateGeometryFromWkt(poly))
-        if format != "DXF":
-            feature.SetField("id", 1)
-        # add feature to layer
-        layer.CreateFeature(feature)
-        # save and close everything
-        feature = None
-        ds = None
+    base, ext = os.path.splitext(orthophoto_file)
+    output_file = base + '_extent.dxf'
 
     try:
         gtif = gdal.Open(orthophoto_file)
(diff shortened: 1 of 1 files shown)


pierotofy:
TFW and WLD files are equivalent (just the extension is different), I'm leaning toward just generating TFW files?

smathermather:
Same tendency. Also, it's been a decade since I've encountered software that uses WLD and not TFW.

lucadelu:
GDAL to understand something more from a PNG file would like to have .wld file 
https://gdal.org/en/stable/drivers/raster/png.html

gdalinfo output with tfw file

```
gdalinfo odm_orthophoto.png
Driver: PNG/Portable Network Graphics
Files: odm_orthophoto.png
       odm_orthophoto.png.aux.xml
Size is 2229, 2123
Image Structure Metadata:
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0, 2123.0)
Upper Right ( 2229.0,    0.0)
Lower Right ( 2229.0, 2123.0)
Center      ( 1114.5, 1061.5)
Band 1 Block=2229x1 Type=Byte, ColorInterp=Red
  Min=0.000 Max=247.000 
  Minimum=0.000, Maximum=247.000, Mean=107.130, StdDev=49.799
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=247
    STATISTICS_MEAN=107.13003773252
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=49.798670180326
    STATISTICS_VALID_PERCENT=65.01
Band 2 Block=2229x1 Type=Byte, ColorInterp=Green
  Min=8.000 Max=250.000 
  Minimum=8.000, Maximum=250.000, Mean=119.696, StdDev=43.391
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=250
    STATISTICS_MEAN=119.69593751376
    STATISTICS_MINIMUM=8
    STATISTICS_STDDEV=43.391056591561
    STATISTICS_VALID_PERCENT=65.01
Band 3 Block=2229x1 Type=Byte, ColorInterp=Blue
  Min=0.000 Max=255.000 
  Minimum=0.000, Maximum=255.000, Mean=91.407, StdDev=49.870
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=91.406659717226
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=49.870229084362
    STATISTICS_VALID_PERCENT=65.01
Band 4 Block=2229x1 Type=Byte, ColorInterp=Alpha
```

gdalinfo output with the wld file 

```gdalinfo odm_orthophoto.png
Driver: PNG/Portable Network Graphics
Files: odm_orthophoto.png
       odm_orthophoto.png.aux.xml
       odm_orthophoto.wld
Size is 2229, 2123
Origin = (576646.834813122870401,5188231.064078653231263)
Pixel Size = (0.057737069400000,-0.057743744200000)
Image Structure Metadata:
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (  576646.835, 5188231.064) 
Lower Left  (  576646.835, 5188108.474) 
Upper Right (  576775.531, 5188231.064) 
Lower Right (  576775.531, 5188108.474) 
Center      (  576711.183, 5188169.769) 
Band 1 Block=2229x1 Type=Byte, ColorInterp=Red
  Min=0.000 Max=247.000 
  Minimum=0.000, Maximum=247.000, Mean=107.130, StdDev=49.799
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=247
    STATISTICS_MEAN=107.13003773252
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=49.798670180326
    STATISTICS_VALID_PERCENT=65.01
Band 2 Block=2229x1 Type=Byte, ColorInterp=Green
  Min=8.000 Max=250.000 
  Minimum=8.000, Maximum=250.000, Mean=119.696, StdDev=43.391
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=250
    STATISTICS_MEAN=119.69593751376
    STATISTICS_MINIMUM=8
    STATISTICS_STDDEV=43.391056591561
    STATISTICS_VALID_PERCENT=65.01
Band 3 Block=2229x1 Type=Byte, ColorInterp=Blue
  Min=0.000 Max=255.000 
  Minimum=0.000, Maximum=255.000, Mean=91.407, StdDev=49.870
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=91.406659717226
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=49.870229084362
    STATISTICS_VALID_PERCENT=65.01
Band 4 Block=2229x1 Type=Byte, ColorInterp=Alpha
```

you can see differences in origin and pixel size.

TFW file is used by TIFF file 
https://gdal.org/en/stable/drivers/raster/gtiff.html

pierotofy:
Just rename the .tfw file to .wld if you need to use gdalinfo?

lucadelu:
Yes, it is possible; it is just a little bit less user friendly. 

As I said in the first post, I understand it is duplication but it doesn't affect calculation time or output size.

You suggest to remove `-co WORLDFILE=YES` option?

smathermather:
To clarify, the are two conventions for naming the file: *.wld and naming it according to image file extension. If the product is a PNG, then the correct world file extension alternative to wld is *.pgw, similarly *.jpw for jpg, etc..

*.tfw for png does not follow the standard, and I agree that having the user rename is not a good experience.

So, either we need logic to handle the variants per output type, or we need to just write a file with a wld extension.

I have a preference for the variants according to extension as it gives the end user some hint that the files are related, but could argue for either in this case.

pierotofy:
Thanks for everyone's feedback. We already have logic in this PR to generate the appropriate file extension depending on the file type (thanks @lucadelu). 

My main concern was about redundancy; yes it's just one file, but it's one file that we will be expected to support for a long time. I just like to know the reasons for why we include stuff. User-friendliness is important, so that's a good reason enough for me, since we already support PNG outputs :+1: 

Side note: as a mature project we have to be a bit careful about feature-creep; it's easy to just have ODM "do more one thing", but we should always ask if an operation is better suited to be done elsewhere downstream (WebODM for example, or via scripts outside of ODM). A monolithic design is not the way to go long term. 

Btw, just curious, what's the use case for using a georeferenced PNG rather than a GeoTIFF?

smathermather:
Yup. Agreed on being careful about core vs. downstream processing, and also that world files are common enough to be useful, especially for CAD users who often have workflows that use import processes dependent on them.

As far as a georeferenced PNG, it can be useful as a lighter-weight standard for images more widely supported by software than a complex geotiff, and have referencing info. But I suspect that in general, the use of a PNG over a TIFF is as much about preferences and expectations as functionality. And PNG guarantees certain things about compression, while TIFF can have so many things inside it's wrapper.

lucadelu:
> 
> Btw, just curious, what's the use case for using a georeferenced PNG rather than a GeoTIFF?

Some software like OpenLayers not support directly GeoTIFF but the support PNG setting the right extension in the world, without wld file you cannot get it.
Not related to orthophoto, but more general, sometimes it happen that you don't want to share the real values but only a visualization of those data so with png it is quiet to prepare and manage

smathermather:
> Not related to orthophoto, but more general, sometimes it happen that you don't want to share the real values but only a visualization of those data so with png it is quiet to prepare and manage

Ah, makes sense. Could easily apply to calibrated channels, like thermal and multispectral.

PR #1843, opened 2025-03-17
Update build scripts for Windows builds

fnuecke:
I've run into some issues while building for Windows on the current master / v3.5.5. Most of this boils down to me using VS 2022 and having had a version of Python installed that did not match the one bundled with the installer. But I've also run into one issue I'm genuinely not sure about.

- For the version mismatch issues, I've added some checks to save others some time in the future.
  - Automatically adjust paths for libs in OpenSfM based on MSVC toolkit version.
  - Update OpenCV dependency to fix compile errors in its protobuf dependency.
  - Add a check for the Python version PDALPython will use and fail if it's not 3.8.1 (otherwise it will build, but the .pyd will not be loadable at runtime).
- Added `python-dateutil` as requirement for the virtual environment, otherwise this would result in an import error at runtime for me.
- Most questionable change: added an early import for `cv2`. Without this, OpenCV initialization fails for me while trying to load its native lib. I traced this down to it breaking if running after PIL's `Image.py` imported its `_imaging` native lib. The error I get without this change is `ImportError: DLL load failed while importing cv2: A dynamic link library (DLL) initialization routine failed.`. No idea why this is; if anyone has an idea how to properly fix this, that would be much appreciated.


pierotofy:
Thanks @fnuecke ! I will wait on merging this since our build process uses an older version of VS (but we'll need to upgrade at some point and this PR will probably help us do so when it's time). 🙏

kmk142789 (approved):

PR #1844, opened 2025-03-19
Install and import cv2 for ai.py

kurkurzz:
Using bg filter throws an error "cv2 is not recognized". Turns out it was not even installed or imported. This is a quickfix that I have tested on my own runs.

### DIFF
--- a/opendm/ai.py
+++ b/opendm/ai.py
@@ -5,6 +5,7 @@ import zipfile
 import time
 import sys
 import rawpy
+import cv2
 
 def read_image(img_path):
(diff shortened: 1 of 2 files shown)


pierotofy:
Nice catch @kurkurzz ! The cv2 import was indeed missing. I've committed https://github.com/OpenDroneMap/ODM/commit/6114e5e934b025a6712e9953e184570f0a334926 to fix it instead of merging this because we shouldn't install opencv via requirements.txt, ODM compiles its own version of OpenCV during build time. :pray:

kurkurzz:
Cool, thank you.

PR #1848, opened 2025-03-28
fix bug of multiple video files and limit_files

### LINKED ISSUE #1847: video files and limit_files
with ODM 3.5.5
I found a bug while importing multiple video files with video2dataset.py.
limit_files is used for all images from all the input video files and not per video file.
I will fix this problem in a following PR

sbonaime:
fix #1847

### DIFF
--- a/opendm/video/video2dataset.py
+++ b/opendm/video/video2dataset.py
@@ -37,10 +37,11 @@ class Video2Dataset:
 
         self.global_idx = 0
 
-        output_file_paths = []
-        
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the PR @sbonaime :pray: 

I'm unsure about interpreting `--video-limit` as a limit that applies per video file; the original intent is to specify a cap on the total number of images (frames) that the reconstruction will have in *total*.

PR #1854, opened 2025-04-06
Declare gdal_proximity to handle nullcheck

sanmit99:
Fix for handling nullcheck on gdal_proximity.

Machine Details: 16x80
- Debian 5.10.223-1 (2024-08-10) x86_64 GNU/Linux
- NVIDIA T4 GPU: 16GB

Container Tested:
opendronemap/nodeodm:gpu

GDAL PIP INFO:
show GDAL
Name: GDAL
Version: 3.0.4
Summary: GDAL: Geospatial Data Abstraction Library
Home-page: http://www.gdal.org
Author: Howard Butler
Author-email: hobu.inc@gmail.com
License: MIT
Location: /usr/lib/python3/dist-packages
Requires: 
Required-by:

```
Traceback (most recent call last):
File "/code/run.py", line 65, in <module>
retcode = app.execute()
File "/code/stages/odm_app.py", line 118, in execute
raise e
File "/code/stages/odm_app.py", line 82, in execute
self.first_stage.run()
File "/code/opendm/types.py", line 470, in run
self.next_stage.run(outputs)
File "/code/opendm/types.py", line 470, in run
self.next_stage.run(outputs)
File "/code/opendm/types.py", line 470, in run
self.next_stage.run(outputs)
[Previous line repeated 5 more times]
File "/code/opendm/types.py", line 449, in run
self.process(self.args, outputs)
File "/code/stages/odm_dem.py", line 66, in process
commands.create_dem(
File "/code/opendm/dem/commands.py", line 187, in create_dem
compute_euclidean_map(tiles_vrt_path, emap_path, overwrite=True)
File "/code/opendm/dem/commands.py", line 213, in compute_euclidean_map
if gdal_proximity is not None:
NameError: name 'gdal_proximity' is not defined
```

PR #1855, opened 2025-04-08
Fallback gdal_proximity command to native path

sanmit99:
Following up to resolve proximity issue: https://github.com/OpenDroneMap/ODM/pull/1854

I have added the native path of gdal_proximity and ran it in subprocess. additional check is passed as well to make sure the path exists to the gdal_proximity script.

### DIFF
--- a/opendm/dem/commands.py
+++ b/opendm/dem/commands.py
@@ -30,8 +30,13 @@ except ModuleNotFoundError:
     # GDAL <= 3.2
     try:
         from osgeo.utils.gdal_proximity import main as gdal_proximity
-    except:
-        pass
+    except ModuleNotFoundError:
+        # GDAL <= 3.0
+        gdal_proximity_script = shutil.which("gdal_proximity.py")
+        if gdal_proximity_script is not None:
+            def gdal_proximity(args):
+                subprocess.run([gdal_proximity_script] + args[1:], check=True)
+
 
 def classify(lasFile, scalar, slope, threshold, window):
     start = datetime.now()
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the PR @sanmit99 ! This is a good workaround, but I don't think it addresses the root cause.

It seems that this issue only happens with the GPU builds of ODM. If you run:

```
root@4d9e2ef9523d:/code# python3 -c "from osgeo import utils;print(utils.__file__)"
/usr/lib/python3/dist-packages/osgeo/utils/__init__.py
```

From the latest CPU docker image, it works fine. But doesn't in the GPU images.

For some reason, the `osgeo_utils` package does not get installed in the system path (`/usr/lib/python3`). Understanding why that is and fixing it should be the proper fix.

sanmit99:
@pierotofy Yes indeed you are right, I had gone through osgeo util import import. utils itself wont import
```
python3
Python 3.8.10 (default, Mar 18 2025, 20:04:55) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import utils
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'utils' from 'osgeo' (/usr/lib/python3/dist-packages/osgeo/__init__.py)
>>> 
```
Note: the container has gdal 3.0.4

And as per investigation going through gdal versions i found out utils were first introduced in gdal3.2:
https://github.com/OSGeo/gdal/blob/release/3.2/gdal/swig/python/osgeo/utils/gdal_proximity.py

Prior versions had it in scripts
https://github.com/OSGeo/gdal/blob/release/3.1/gdal/swig/python/scripts/gdal_proximity.py

sanmit99:
> @pierotofy Yes indeed you are right, I had gone through osgeo util import import. utils itself wont import
> 
> ```
> python3
> Python 3.8.10 (default, Mar 18 2025, 20:04:55) 
> [GCC 9.4.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> from osgeo import utils
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ImportError: cannot import name 'utils' from 'osgeo' (/usr/lib/python3/dist-packages/osgeo/__init__.py)
> >>> 
> ```
> 
> Note: the container has gdal 3.0.4
> 
> And as per investigation going through gdal versions i found out utils were first introduced in gdal3.2: https://github.com/OSGeo/gdal/blob/release/3.2/gdal/swig/python/osgeo/utils/gdal_proximity.py
> 
> Prior versions had it in scripts https://github.com/OSGeo/gdal/blob/release/3.1/gdal/swig/python/scripts/gdal_proximity.py

This problem is with gdal versions. if we can either force to have bare minimum gdal version or else we can handle like i have shared in PR. 

:) hope this helps

pierotofy:
Thanks for investigating. Upgrading GDAL could be tricky. I ended up just refactoring the solution proposed here by defining a fallback function for gdal_proximity instead.

PR #1860, opened 2025-05-05
Compress PNGs generated with mvs-texturing

pierotofy:
Realized that we don't currently compress PNGs generated with mvs-texturing, but we really should to reduce disk space usage (~20% reduction for all PNG textures).

Related change: https://github.com/OpenDroneMap/mve/commit/90f13c18f92db6089c06c8c2451e283911f5f634


xinzheng798:
image
The core reason for the difference between the CC in the above figure and the texture images generated by ODM, in my opinion, lies in the inadequate quality of the generated texture patches and the bin packing algorithm.

PR #1861, opened 2025-05-10
Allow the usage of a different output directory in winpostinstall.bat

sylveon:
This is required for packaging ODM in a MSIX package, which are deployed to a read-only location. PSF can redirect writes to the package root to a writeable overlay directory to compensate for that.

PSF also supports start scripts so we can run `winpostinstall.bat` to patch up the path to where ODM is installed (usually under `C:\Program Files\WindowsApps`, but this can be changed). However those don't run under the FileRedirectionFixup, so we need to manually write to the overlay directory so that Python picks them up with the FileRedirectionFixup later.


pierotofy:
Thanks @sylveon ! This looks good to me.

PR #1880, opened 2025-07-07
Add gps-z-offset

pierotofy:
Adds the ability to offset GPS Z values directly via a command line parameter, rather than needing to create a geo.txt file manually.


smathermather:
Love it. A much desired feature.

PR #1882, opened 2025-07-07
Proposal for a Time-SIFT script

denis-f:
As discussed [here](https://community.opendronemap.org/t/doing-time-sift-multi-temporal-co-registration-without-gcps-with-odm/24708), I propose a script that allows for multi-temporal processing and intrinsic co-registration of all epochs without a posteriori processing thanks to our so-called 'Time-SIFT' method.

This is my first contribution to a github project, I even don't know how we do usually. Hope this works fine


pierotofy:
Thanks for sharing @denis-f ! 🙏 

The only thing to change/fix before merging this is to not include any images/datasets (no PNGs, JPEGs); they increase the size of the codebase and of the git repository.

denis-f:
Ok, I've put the example data elsewhere to save space on the repository and keep the example

pierotofy:
@smathermather did you do a squash commit?

htami1717:
‪+92 346 5557038‬

htami1717:
‏أو أعرف المواقع

PR #1885, opened 2025-07-16
Crops tif by cutline by default if cutline is calculated

smathermather:
cutline_crop

### DIFF
--- a/stages/odm_orthophoto.py
+++ b/stages/odm_orthophoto.py
@@ -133,11 +133,10 @@ class ODMOrthoPhotoStage(types.ODM_Stage):
                     reconstruction, tree, not outputs["large"])
 
                 # Generate feathered orthophoto also
-                if args.orthophoto_cutline and submodel_run:
-                    orthophoto.feather_raster(tree.odm_orthophoto_tif, 
-                            os.path.join(tree.odm_orthophoto, "odm_orthophoto_feathered.tif"),
-                            blend_distance=20
-                        )
+                orthophoto.feather_raster(tree.odm_orthophoto_tif, 
+                        os.path.join(tree.odm_orthophoto, "odm_orthophoto_feathered.tif"),
+                        blend_distance=20
+                    )
 
             else:
                 if io.file_exists(tree.odm_orthophoto_render):
--- a/stages/odm_orthophoto.py
+++ b/stages/odm_orthophoto.py
@@ -133,10 +133,11 @@ class ODMOrthoPhotoStage(types.ODM_Stage):
                     reconstruction, tree, not outputs["large"])
 
                 # Generate feathered orthophoto also
-                orthophoto.feather_raster(tree.odm_orthophoto_tif, 
-                        os.path.join(tree.odm_orthophoto, "odm_orthophoto_feathered.tif"),
-                        blend_distance=20
-                    )
+                if args.orthophoto_cutline and submodel_run:
+                    orthophoto.feather_raster(tree.odm_orthophoto_tif, 
+                            os.path.join(tree.odm_orthophoto, "odm_orthophoto_feathered.tif"),
+                            blend_distance=20
+                        )
 
             else:
                 if io.file_exists(tree.odm_orthophoto_render):
--- a/stages/odm_orthophoto.py
+++ b/stages/odm_orthophoto.py
@@ -125,12 +125,9 @@ class ODMOrthoPhotoStage(types.ODM_Stage):
                                     args.max_concurrency,
                                     scale=0.25)
                     
-                    if submodel_run:
-                        orthophoto.compute_mask_raster(tree.odm_orthophoto_tif, cutline_file, 
-                                            os.path.join(tree.odm_orthophoto, "odm_orthophoto_cut.tif"),
-                                            blend_distance=20, only_max_coords_feature=True)
-                    else:
-                        log.ODM_INFO("Not a submodel run, skipping mask raster generation")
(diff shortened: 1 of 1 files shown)


smathermather:
Slight API change, but if you're generating cutlines, you probably want the cut version instead of or in addition to the uncut version.

One point of confusion is whether the feathered version adds anything here outside a split-merge scenario.

pierotofy:
You're not cropping by cutline with these changes though (just rasterizing the cutline vector to a raster?).

The feathered version is only useful for split-merge scenarios imo, it's an optimization for speeding up the later merge stage.

smathermather:
Yeah, I wondered about the feathered portion. If someone wants to get a cutline though, the alpha values in the feathered ortho could also be useful to them. But, it does require some additional logic to use. I may write a module for that, as it is of general utility (it's remarkably difficult to find good FOSS tools for cutline / blend in the wild).

But as to whether we are generating the output expected, we do get clipped ortho from this:

```
ls odm_orthophoto/*.tif
odm_orthophoto/odm_orthophoto_cut.tif
odm_orthophoto/odm_orthophoto_feathered.tif
odm_orthophoto/odm_orthophoto.original.tif
odm_orthophoto/odm_orthophoto.tif
```

image

An argument could be made that maybe odm_orthophoto_cut replace odm_orthophoto, but I'm not sure. Or it could be an additional download via WebODM if we add support for it there, alternatively.

pierotofy:
It's an interesting discussion about redundant data; the feathered output can be computed post processing (via a script) without too much trouble from the original orthophoto + cutline vector. It's just computationally expensive and that's why it's computed for split-merge purposes (to avoid having the primary node needing to perform this computation itself during merge). Instead, it was designed to have submodel nodes compute the feathered product, so that merge can be really fast on the primary node. Note that during merge, both the original orthophoto and the feathered orthophoto are needed (otherwise you'll get cutline edges in the merged result), it's a two pass process.

I agree that there's a general lack of simple open tools for doing large scale merging (thus why everything here was built from scratch), in the spirit of modularity I do think this could be an opportunity to improve the tools for doing this kind of task, but I'm not sure ODM is the (ideal) place where this should happen.

pierotofy:
WebODM for example could compute/export the feathered orthophoto on the fly, with some modifications, without needing to store additional data.

There's actually been some interest in having a "merge" functionality within WebODM (e.g. imagine taking two tasks which are already aligned, pressing merge and boom, now you have a single task with the two tasks merged into single products).

smathermather:
> There's actually been some interest in having a "merge" functionality within WebODM (e.g. imagine taking two tasks which are already aligned, pressing merge and boom, now you have a single task with the two tasks merged into single products).

Yes. Now that align is there, that would be a sensible addition to WebODM.

> I agree that there's a general lack of simple open tools for doing large scale merging (thus why everything here was built from scratch), in the spirit of modularity I do think this could be an opportunity to improve the tools for doing this kind of task, but I'm not sure ODM is the (ideal) place where this should happen.

Agreed that ODM shouldn't be the tool for large scale merging tool. But if ODM is outputting a cutline, I assert that it should provide a cut version of the ortho, and potentially the pre-calculated feathering as well, if it's relatively inexpensive to do so.

I'm currently rerunning on batches across a 50,000 images subset (of a 113k dataset). It appears to add 1 minute to a 35 minute process for each. So: <3% penalty if an only if you enable the flag.

pierotofy:
I'll just link to here https://github.com/OpenDroneMap/ODM/pull/1837#issuecomment-2702625559 and ask to consider why this isn't something that can be done as a post processing step via a script, but the final decision is not mine, so I'll leave it at that.

smathermather:
> I'll just link to here [#1837 (comment)](https://github.com/OpenDroneMap/ODM/pull/1837#issuecomment-2702625559)

Anything can be done as a post processing step. We could say that about vegetation indices, alignment to other datasets, clipping to an existing boundary, etc.. We can always raise that question. And that question is useful, but only in the context of the other points raised in that comment:

Any pull request needs to meet a few criteria, that includes increased long term maintenance and complexity as weighted against core functionality vs. post-processing. With that context, I'm not sure how that post is relevant to this pull request. 

This pull request decreases complexity and doesn't introduce new types of output. I also don't see it as a custom use case as implied.

A few differences to that issue along a few axes:
* I see it as a good affordance for anyone throwing the flag for the same reasons that split merge uses it -- it puts the onus of using the pipelines on the machine already doing most of the processing. Split merge isn't the only tooling or workflow where this is useful, and we shouldn't assume that it is.
* It is not additional code, nor additional output types.
* Nominally it's less code and less complexity
* It is computationally a nominal increase
* And that nominal increase affects a small number of users.

Importantly, it also increases the utility of that flag and distributes compute closer to the data source, which is the behavior I think we generally want to see in a core utility.

pierotofy:
> This pull request decreases complexity and doesn't introduce new types of output. I also don't see it as a custom use case as implied.

Mm, it increases computation time and yes, it does introduce new file(s). Now the expectation will be that when I turn on `--orthophoto-cutline` that I will get cutline raster(s). I know you suggested replacing odm_orthophoto.tif, but that's not advisable, you're breaking split-merge as well as a contract with your users.

But that's the beauty of open source, you can include whatever you want, you don't need to ask, this is a matter of whether you value keeping things fast, lean and modular or create monoliths for *others*, I prefer the former and that's just personal preference.

smathermather:
Closing in favor of https://github.com/OpenDroneMap/WebODM/pull/1699

PR #1888, opened 2025-07-20
Use kdTree for GSD estimation

pierotofy:
Improves the GSD estimate computation by using a 2D-kdtree over the reconstruction area, rather than averaging a plane over the entire reconstruction. This improves the estimate especially when terrain following is enabled.

### DIFF
--- a/opendm/gsd.py
+++ b/opendm/gsd.py
@@ -5,6 +5,7 @@ import math
 from repoze.lru import lru_cache
 from opendm import log
 from opendm.shots import get_origin
+from scipy import spatial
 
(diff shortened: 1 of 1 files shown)


smathermather:
Nice. Any particular testing needed?

pierotofy:
Nop, pretty straightforward.

smathermather:
Seems it. Cool. Test already started running on a cliff. I'll be interested to see how it differs, but full agree this is an improved approach.

smathermather:
Improved estimate for these data: https://community.opendronemap.org/t/3d-model-of-cliff-face/24551 from 9cm to 2cm.

PR #1895, opened 2025-07-24
Minor updates for GPU

### LINKED ISSUE #1894: Nvidia 5 series GPU results in crash
> > if [[ "$UBUNTU_VERSION" == _"20.04"_ ]]; then echo "Enabling PPA for Ubuntu GIS" sudo $APT_GET install -y -qq --no-install-recommends software-properties-common sudo add-apt-repository -y ppa:ubuntugis/ubuntugis-unstable sudo $APT_GET update fi
> > This makes me wonder if the package on repository has changed?
> 
> Hmm. Good catch. Forgot that logic was in there. But with that pull request (now with the correct number...) the build completes now, which it didn't before, right? Does it not run?
> 
> no it has always built, just didn't support GOG format. The current gpu container processes the aukerman fine, although turns out doesn't run SIFT on the GPU anymore! Not sure why, as the logging isn't great, I've got 5070Ti now though. The local one I built while trying fix the COG does though, I'm just trying to pin down why (cuda versions?).  But this is a separate  issue to this one.  I also noticed the readme is wrong, needs to have feature-type sift in to use gpu (will submit at PR).  

 _Originally posted by @ehallein in [#1864](https://github.com/OpenDroneMap/ODM/issues/1864#issuecomment-3111562856)_

edit: yep the GPU  container needs cuda 11.8 to run in 5 series GPU's:

/code/SuperBuild/build/pypopsift/_deps/popsift-src/src/popsift/gauss_filter.cu:245
    cudaMemcpyToSymbol failed for Gauss kernel initialization: unrecognized error code

ehallein:
updated readme for GPU
refactored GPU logging
updated to cuda 11.8 container to be compatible with 5 series GPUs fixes #1894. Not sure how relates to #https://github.com/OpenDroneMap/ODM/commit/36897fc0e98c27728bb43e172269a82766e0c9cd as removed that change (was for cuda 11.2 and removed compatibility for previous cuda versions.)
updated gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ run C:\Users\youruser\datasets\project  [--additional --parameters --here]
 ODM has support for doing SIFT feature extraction on a GPU, which is about 2x faster than the CPU on a typical consumer laptop. To use this feature, you need to use the `opendronemap/odm:gpu` docker image instead of `opendronemap/odm` and you need to pass the `--gpus all` flag:
 
 ```
-docker run -ti --rm -v c:/Users/youruser/datasets:/datasets --gpus all opendronemap/odm:gpu --project-path /datasets project
+docker run -ti --rm -v c:/Users/youruser/datasets:/datasets --gpus all opendronemap/odm:gpu --project-path /datasets project --feature-type sift
 ```
 
 When you run ODM, if the GPU is recognized, in the first few lines of output you should see:
--- a/opendm/gpu.py
+++ b/opendm/gpu.py
@@ -19,13 +19,19 @@ def has_popsift_and_can_handle_texsize(width, height):
             log.ODM_INFO("CUDA compute platform is not supported (detected: %s.%s but we need at least 3.5)" % (compute_major, compute_minor))
             return False
     except Exception as e:
-        log.ODM_INFO("Using CPU for feature extraction: %s" % str(e))
+        log.ODM_WARNING(str(e))
         return False
 
     try:
         from opensfm import pypopsift
-        return pypopsift.fits_texture(int(width * 1.02), int(height * 1.02))
+        if pypopsift.fits_texture(int(width * 1.02), int(height * 1.02)):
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -95,7 +95,7 @@ run C:\Users\youruser\datasets\project  [--additional --parameters --here]
 ODM has support for doing SIFT feature extraction on a GPU, which is about 2x faster than the CPU on a typical consumer laptop. To use this feature, you need to use the `opendronemap/odm:gpu` docker image instead of `opendronemap/odm` and you need to pass the `--gpus all` flag:
 
 ```
-docker run -ti --rm -v c:/Users/youruser/datasets:/datasets --gpus all opendronemap/odm:gpu --project-path /datasets project
+docker run -ti --rm -v c:/Users/youruser/datasets:/datasets --gpus all opendronemap/odm:gpu --project-path /datasets project --feature-type sift
```
smathermather, line 98: Ah, nice.

opendm/gpu.py
```diff
@@ -19,13 +19,19 @@ def has_popsift_and_can_handle_texsize(width, height):
...
         from opensfm import pypopsift
-        return pypopsift.fits_texture(int(width * 1.02), int(height * 1.02))
+        if pypopsift.fits_texture(int(width * 1.02), int(height * 1.02)):
+            log.ODM_INFO("popsift can handle texture size %dx%d" % (width, height))
+            return True
+        else:
+            log.ODM_INFO("popsift cannot handle texture size %dx%d" % (width, height))
+            return False
     except (ModuleNotFoundError, ImportError):
+        log.ODM_WARNING(str(e))
```
pierotofy: `e` is not defined here

smathermather: Mm. Probably:
`except (ModuleNotFoundError, ImportError) as e:`

ehallein: good catch, have removed entirely as redundant anyway.

smathermather: Yup. Makes sense.

PR #1915, opened 2025-08-22
add forward option to retain las file scale and offset

originlake:
When creating the georeferenced laz file, the scale is set to 0.001m(or value of spacing if it's smaller), offset is set to origin of the dataset. But the following classification steps do not respect the scale and offset, using the default value of pdal instead, which restricts the precision of point cloud data to 0.01m. 

Fixed the smrf filter that used pdal directly. Still working on fixing OpenPointClass

### DIFF
--- a/opendm/dem/pdal.py
+++ b/opendm/dem/pdal.py
@@ -164,6 +164,7 @@ def run_pdaltranslate_smrf(fin, fout, scalar, slope, threshold, window):
         '--filters.smrf.slope=%s' % slope,
         '--filters.smrf.threshold=%s' % threshold,
         '--filters.smrf.window=%s' % window,
+        '--writers.las.forward=scale,offset'
     ]
 
     system.run(' '.join(cmd))


originlake:
I have one question, how should I prepare the fix for OpenPointClass? It's not a fork and the fix would be a special patch just for ODM I suppose.

originlake:
https://github.com/originlake/OpenPointClass/tree/las-scale fix for openpointclass is here

smathermather:
Thanks for this!

As OpenPointClass (AFAIK) only operates on PLYs, there shouldn't be any floating point issues requiring setting las scale. But, double check my assumptions on that. I haven't looked under the hood of OpenPointClass in any serious way.

Take a look at this pull request to appropriately set the scale instead of fixed:
https://github.com/OpenDroneMap/ODM/pull/1788

originlake:
The classification starts with a smrf filter then followed with a openpointclass step. OPC does use laz file and overwrites the old laz file, leading to this issue.  So both need to be fixed. 

Code for classification steps:
https://github.com/OpenDroneMap/ODM/blob/5d4862e5157fe3dbcc236972018060155aca96e3/opendm/point_cloud.py#L280-L300

Code where the opc is used:
https://github.com/OpenDroneMap/ODM/blob/5d4862e5157fe3dbcc236972018060155aca96e3/opendm/opc.py#L7-L30

The scale and offset are computed dynamically in an early stage, my fix proposed here follows what is being set in the original laz file, which should be expected as the classification step should not alter the point cloud itself. 

Code where the offset and scale are set in the original laz file:
https://github.com/OpenDroneMap/ODM/blob/5d4862e5157fe3dbcc236972018060155aca96e3/stages/odm_georeferencing.py#L131-L160

smathermather:
Got it. Pull request against OpenPointClass and explain your concerns. If rejected and we need to maintain a fork with this functionality under an OpenDroneMap banner, happy to do so.

originlake:
Thanks, created the PR, will see how it goes. https://github.com/uav4geo/OpenPointClass/pull/33

smathermather:
Nice work! Just update SuperBuild/cmake/External-OpenPointClass.cmake to point to the hash of the commit (dd6a560a1d43cb709f7b220b19a436e25a889e3e), since we don't know when the next release will be cut, and this is similar to how we're handling other upstream libraries now, and we should be good to merge.

smathermather:
Awesome. Thanks! I'll update SuperBuild/cmake/External-OpenPointClass.cmake to point to the appropriate commit.

PR #1934, opened 2025-10-07
Add flag `--merge-skip-blending` to skip expensive blending steps during splitmerge workflow

spwoodcock:
- On request from @smathermather, after using splitmerge on a large 80km2 area of imagery in Freetown.
- If used, the flag skips step 2 and 3 of blending, during ortho production, as the step is slower than all the other processing 'by some factors'.
- This is likely a temporary solution - added a note to the code about this:

```python
    # Ideally the blending should be optimised so we don't need this param to skip it, e.g.
    #   1. Only blend in a buffer zone around cutline edges (e.g., 100-200 pixels)
    #   2. Skip blending for interior regions that are far from edges
    #   3. This would require detecting which blocks intersect cutline boundaries
    # We have block-based processing already in place in orthophoto.py, so need logic to determine
    # if a block needs blending based on its proximity to cutlines
```

### DIFF
--- a/contrib/mergepreview/mergepreview.py
+++ b/contrib/mergepreview/mergepreview.py
@@ -116,7 +116,7 @@ if len(all_orthos_and_ortho_cuts) > 1:
         'BIGTIFF': 'IF_SAFER',
         'BLOCKXSIZE': 512,
         'BLOCKYSIZE': 512
-    })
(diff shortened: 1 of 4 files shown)


smathermather:
Thanks Sam!

PR #1935, opened 2025-10-07
Add interleave=pixel and photometric=ycbcr for orthophoto overviews by default

spwoodcock:
- Ortho overview performance in QGIS isn't great for large datasets.
- This can be optimised using params:

```
PHOTOMETRIC_OVERVIEW=YCBCR
INTERLEAVE_OVERVIEW=PIXEL
```

Credit for the trick from @ivangayton.

This PR:
- Adds an optional flag to include these params during processing with GDAL.
- Works for both `--cog` and without that flag - optimising the overviews in both cases.

### DIFF
--- a/opendm/cogeo.py
+++ b/opendm/cogeo.py
@@ -15,7 +15,7 @@ def convert_to_cogeo(src_path, blocksize=256, max_workers=1, compression="DEFLAT
     """
 
     if not os.path.isfile(src_path):
-        logger.warning("Cannot convert to cogeo: %s (file does not exist)" % src_path)
+        log.ODM_WARNING("Cannot convert to cogeo: %s (file does not exist)" % src_path)
         return False
 
     log.ODM_INFO("Optimizing %s as Cloud Optimized GeoTIFF" % src_path)
--- a/opendm/cogeo.py
+++ b/opendm/cogeo.py
@@ -5,7 +5,7 @@ from opendm.concurrency import get_max_memory
 from opendm import io
 from opendm import log
 
-def convert_to_cogeo(src_path, blocksize=256, max_workers=1, compression="DEFLATE"):
+def convert_to_cogeo(src_path, blocksize=256, max_workers=1, compression="DEFLATE", compact_overviews=False):
     """
     Guarantee that the .tif passed as an argument is a Cloud Optimized GeoTIFF (cogeo)
     The file is destructively converted into a cogeo.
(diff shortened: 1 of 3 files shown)


opendm/cogeo.py
```diff
@@ -15,39 +15,49 @@ def convert_to_cogeo(src_path, blocksize=256, max_workers=1, compression="DEFLAT
     """
 
     if not os.path.isfile(src_path):
-        logger.warning("Cannot convert to cogeo: %s (file does not exist)" % src_path)
+        log.ODM_WARNING("Cannot convert to cogeo: %s (file does not exist)" % src_path)
```
spwoodcock, line 18: I also found this bug while updating the code - if the input file doesn't exist, the ODM logger wasn't used correctly.

opendm/cogeo.py
```diff
@@ -15,39 +15,49 @@ def convert_to_cogeo(src_path, blocksize=256, max_workers=1, compression="DEFLAT
...
-        'compress': compression,
-        'predictor': '2' if compression in ['LZW', 'DEFLATE'] else '1',
-    }
+    # Configuration params
+    threads = max_workers if max_workers else 'ALL_CPUS'
+    predictor = '2' if compression in ['LZW', 'DEFLATE'] else '1'
+    max_memory = get_max_memory()
+
+    # Build gdal_translate command
+    cmd = [
```
spwoodcock: Decided to shift from using the `**kwargs` approach to make adding this optional config simpler.

f-strings are also more readable IMO - the codebase could probably do with a refactor to use them.
I doubt anyone is using Python < 3.6 now?

spwoodcock:
Actually, this doesn't work as intended for COGs

```
[INFO]    running gdal_translate -of COG -co NUM_THREADS=16 -co BLOCKSIZE=256 -co COMPRESS=DEFLATE -co PREDICTOR=2 -co BIGTIFF=IF_SAFER -co RESAMPLING=NEAREST -co PHOTOMETRIC_OVERVIEW=YCBCR -co INTERLEAVE_OVERVIEW=PIXEL --config GDAL_CACHEMAX 37.9% --config GDAL_NUM_THREADS 16 /code/data/odm_orthophoto/odm_orthophoto.tif /code/data/odm_orthophoto/odm_orthophoto_cogeo.tif
Warning 6: driver COG does not support creation option PHOTOMETRIC_OVERVIEW
Warning 6: driver COG does not support creation option INTERLEAVE_OVERVIEW
```

It looks like we could set `INTERLEAVE=PIXEL`, but based on this note in the docs

```
JPEG should generally only be used with Byte data (8 bit per channel). But if GDAL is built with internal libtiff and libjpeg, it is possible to read and write TIFF files with 12bit JPEG compressed TIFF files (seen as UInt16 bands with NBITS=12). For the COG driver, JPEG compression for 3 or 4-band images automatically selects the PHOTOMETRIC=YCBCR colorspace with a 4:2:2 subsampling of the Y,Cb,Cr components with the default INTERLEAVE=PIXEL. For a input dataset (single-band or 3-band), plus an alpha band, the alpha band will be converted as a 1-bit DEFLATE compressed mask.
```

it might make more sense to specify a flag `--cog-compression` and allow the user to specify `JPEG`, killing two birds with one stone...

spwoodcock:
I'm not entirely sure what is the preferred approach here.

We could use a flag like `--compact-overviews` to support **both** standard and COG ortho output with a single config.

Or to add `--cog-compression`, so when set to JPEG the configuration we need will be set automatically. This is more flexible when it comes to COG (allowing to specify the compression type), but won't support the standard ortho output, meaning two flags are needed.

spwoodcock:
Simplified this! We didn't need an extra flag:
- For standard GeoTIFF, the overviews were already configured to `COMPRESSION=JPEG`, so I took the liberty to add `INTERLEAVE_OVERVIEW=PIXEL` and `PHOTOMETRIC_OVERVIEW=YCBCR` by default (as these are not set by default).
- For COG, we can already set the `--orthophoto-compression=JPEG` flag, and this is passed through, adding the interleave and photometric settings automatically, see [docs](https://gdal.org/en/stable/drivers/raster/cog.html#general-creation-options)

I kept the Git history via a revert, as I thought it might be useful if needed.

opendm/cogeo.py
smathermather: Ace. Thanks.

opendm/cogeo.py
smathermather: We can probably switch away from kwargs. No longer relevant to this pull request (which is awesome).

PR #1942, opened 2025-10-27
Merge branch `24.04` --> `master` (#1904)

spwoodcock:
As mentioned in #1904, we can merge the latest release into master.
After that I can start #1936
To clarify, this is based on Ubuntu 24.04, but the new ODM version is **3.6.0**.

* Changes to support building on Ubuntu 24.04 and windows-2022 Github runner.
* Update Python to 3.12
    * Install dependencies in virtual environment
    * Run python scripts from the virtual environment)
* Update dependencies
    * Ubuntu dependencies in snap/snapcraft24.yaml
    * Python dependencies in requirements.txt
    * Windows dependencies built in OpenDroneMap/windows-deps repo
* Update CUDA
    * 12.8.1 on Windows
    * 13.0.0 on Ubuntu)
* Run tests as part of docker build
* Use exact commits to specify dependencies that are built from source, instead of branch names
* Use upstream versions of Libraries:
    * PDAL
    * PDAL-Python
    * untwine
    * ExifRead
    * draco
* Build windows builds with -j2
* Use Micasense's latest version of dls.py
* Update failing unit tests to match current behavior

PR #1943, opened 2025-10-27
Replace pip & requirements.txt with uv & pyproject.toml

### LINKED ISSUE #1936: Modernise Python setup for repo
After https://github.com/OpenDroneMap/ODM/pull/1904 is well tested and merged, I would happily contribute a few things:
- Upgrade from using standard `pip` --> `uv`. Running Python via venv is the recommended approach, regardless of container usage. See comment https://github.com/OpenDroneMap/ODM/pull/1904#issuecomment-3172016947. `uv` would add much better dependency resolution and conflict checking. Users could run outside a container too, without trashing their system python install.
- Move most Python metadata to `pyproject.toml`, such as `requirements.txt` that would go under `[dependencies]`.
- Update all links to requirements.txt to first export this via `uv export`, having the resolved `pyproject.toml` and `uv.lock` as a single source of truth.
- Update the dockerfiles to be a bit more efficient with caching / multi-stage builds, where possible.

spwoodcock:
Fixes #1936

Work in progress.
Built in top of #1942, so the commits will look a lot cleaner once merged.

Tasks:
- [x] Migrate requirements.txt --> pyproject.toml (incl platform specific deps)
- [ ] Update context.py and other places that reference local installs of packages like opencv, opensfm
- [ ] Create a `scripts` section in pyproject.toml for CLI usage
- [ ] Update all references to pip / pip3 --> uv
- [ ] Update snapcraft builds (do we still maintain these? If so, I can look at them. If not, no point spending time)
- [ ] Tweak Dockerfiles to use UV
- [ ] Lots of testing for Linux / Windows (+ a community member to test the MacOS builds 🙏)


kmk142789 (approved):

.github/workflows/publish-windows.yml
```diff
@@ -32,12 +32,13 @@ jobs:
...
-      run: |
-        python -m pip install virtualenv
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+      with:
+        version: "0.9.5"
     - name: Build sources
       run: |
-        python configure.py build
+        uv run configure.py build
```
spwoodcock, line 41: I believe this should be `uv run python configure.py`

PR #1948, opened 2025-11-07
Improve camera identification

originlake:
Currently the camera id is created by combining camera make, camera model, width, height, projection type, focal length. However, this is not enough to cover the case when there are two or more drones with the same model and the same capture settings, this can happen on a large survey project where one drone's battery is not enough to cover the whole survey. And if not separating data from multiple drones, a single camera model will be used to optimize across all the images, leading worse accuracy. 

This PR tries to solve this issue by utilizing serial number exists mostly in the drone image metadata. It will attach the serial number at the end of the camera id. Since there could be multiple tags used to represent the serial number from different drone/camera maker, I don't have enough data to cover all the scenario, my implementation is explained below, please suggest and if any other way to read serial number is missing, let me know. 

The serial number will be `unknown` by default, then, when loading metadata, the following exif and xmp tags will be used with the priority in ascending order, camera/len serial number should always be preferred over drone/body serial number, as some drones allow to change sensors. `EXIF BodySerialNumber`, `EXIF LensSerialNumber`, `@drone-dji:DroneSerialNumber`, `@drone-dji:CameraSerialNumber`

### DIFF
--- a/opendm/photo.py
+++ b/opendm/photo.py
@@ -174,6 +174,9 @@ class ODM_Photo:
         self.camera_projection = 'brown'
         self.focal_ratio = 0.85
 
+        # Sensor identifier
+        self.serial_number = 'unknown'
+
         # parse values from metadata
         self.parse_exif_values(path_file)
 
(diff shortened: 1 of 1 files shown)


smathermather:
Love this. Do you have a test dataset I could run it with?

originlake:
It's tested with some data under NDA :cry: , and has no side effect in normal datasets. If we could ask for some help from the community. Sorry I always have to create PR without actual data, my experience mostly comes from dealing issues with datasets under NDA.

smathermather:
> my experience mostly comes from dealing issues with datasets under NDA.

Makes perfect sense. Could you take https://github.com/pierotofy/drone_dataset_brighton_beach and modify the tags on half the files to simulate it?

If not, I can do this in a bit, but haven't had the moment to do so yet.

originlake:
[report.pdf](https://github.com/user-attachments/files/23778609/report.pdf)
Attached the report here, you can see two different camera models were used in the reconstruction. 

You can find the modified images here https://drive.google.com/file/d/1r1sO0zmqtIZ7fG2V2LHWgrqdjW1RV5Wa/view?usp=sharing

PR #1958, opened 2025-11-20
Ensure we use GDAL 3.11.1 in Python bindings after 24.04 upgrade

spwoodcock:
Follow on from #1956 and https://github.com/OpenDroneMap/ODM/pull/1904

Closes #1953

Fix to ensure 24.04 upgrade is fully functional.

After building latest, I tested on the brighton beach dataset:

```bash
cd ODM
docker build . -t opendronemap/odm:latest
cd ..
git clone https://github.com/pierotofy/drone_dataset_brighton_beach.git
cd drone_dataset_brighton_beach

docker run -it --rm -v "$(pwd)/images:/code/images" -v "$(pwd)/odm_orthophoto:/code/odm_orthophoto" -v "$(pwd)/odm_texturing:/code/odm_texturing" docker.io/opendronemap/odm:latest --dtm --rerun-all
```

Output:

```
[INFO]    running gdaldem color-relief "/code/odm_dem/dtm.preview.tif" "/code/opendm/tiles/color_relief.txt" "/code/odm_dem/dtm.previewcolor.tif" -alpha -co ALPHA=YES
0...10...20...30...40...50...60...70...80...90...100 - done.
[INFO]    running gdaldem hillshade "/code/odm_dem/dtm.preview.tif" "/code/odm_dem/dtm.previewhillshade.tif" -z 1.0 -s 1.0 -az 315.0 -alt 45.0
0...10...20...30...40...50...60...70...80...90...100 - done.
[INFO]    running "/code/venv/bin/python3" "/code/opendm/tiles/hsv_merge.py" "/code/odm_dem/dtm.previewcolor.tif" "/code/odm_dem/dtm.previewhillshade.tif" "/code/odm_dem/dtm.previewcolored_hillshade.tif"
/code/venv/lib/python3.12/site-packages/osgeo/gdal.py:330: FutureWarning: Neither gdal.UseExceptions() nor gdal.DontUseExceptions() has been explicitly called. In GDAL 4.0, exceptions will be enabled by default.
  warnings.warn(
0...10...20...30...40...50...60...70...80...90...100 - done.
[INFO]    running gdal_translate -outsize 1400 0 -of png "/code/odm_dem/dtm.previewcolored_hillshade.tif" "/code/opensfm/stats/dtm.png" --config GDAL_CACHEMAX 37.35%
Input file size is 1400, 1379
0...10...20...30...40...50...60...70...80...90...100 - done.
[INFO]    Exporting report to /code/odm_report/report.pdf
[...]

PR #1959, opened 2025-11-20
Add ccache for cmake caching + optimise dockerfile ordering for cache usage

spwoodcock:
### Dockerfile caching

- While debugging https://github.com/OpenDroneMap/ODM/pull/1958 it was a bit of a pain having to do lots of slow rebuilds for testing.
- Builds previously took about 2529.5s with buildkit on my Framework 16. After the PR for GDAL 3.11.1 building, it takes 2823.0s.
- The initial build after this PR took a marginally longer 3212.5s, with subsequent builds taking a much shorter 1976.9s (~46m --> 33m).
- I tested again modifying a version in `requirements.txt` using the ccache cache, giving 1626.3s build time - big improvement!

### CI caching

- I took the opportunity to add some caching to the primary docker publish workflow.
- As a result, I upgraded most of the workflow versions, which are 5yrs outdated.
- The only breaking change I saw was with the metadata action - updated.

> [!NOTE]
> The windows build and GPU build CI workflows also need updating to match this.
> Testing with the single workflow first, then we can propagate across.

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -13,40 +13,46 @@ jobs:
     timeout-minutes: 2880
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
+    
     - name: Set up QEMU
-      uses: docker/setup-qemu-action@v1
+      uses: docker/setup-qemu-action@v3
+    
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v1
+      uses: docker/setup-buildx-action@v3
       with:
         config-inline: |
           [worker.oci]
             max-parallelism = 1
+    
     - name: Login to DockerHub
-      uses: docker/login-action@v1 
+      uses: docker/login-action@v3
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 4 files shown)


originlake:
I'm working on something similar to optimize the docker build steps, but mostly trying to make the start-dev-env script a bit more easier to use. Currently if you delete the odmdev container, you have to rerun everything from scratch, really inconvenient. I'm trying to make it a 3 stages build and the first stage, which only installs the dev dependency can be saved for use of `start-dev-env.sh`. Can also publish the dev stage as an image so it can be faster for anyone wants to setup the dev environment. 

@spwoodcock @smathermather I think I'll continue my work after this PR is merged, but below is my work in progress, if you can provide feedback, it would be much appreciated

https://github.com/originlake/ODM/commit/2f2aba698ec3bea7890b6f8e2975aa68ab8e4ca1

spwoodcock:
Thanks @originlake!

Sorry for the slow reply with the Christmas break 😄

The changes you propose could be useful. I also planned to optimise the builds a bit, after finishing my PR for using `uv` to track dependencies.

Feel free to make a PR that targets this branch if you wish to extend from the ccache optimisation 🙏

spwoodcock:
Likely superseded by https://github.com/OpenDroneMap/ODM/pull/2037

I imagine pixi will handle all the caching

PR #1970, opened 2025-12-22
fix(photo): handle special case for Mavic 3E backward-facing images

thadwald:
DJI Mavic 3 Enterprise has an unintuitive method for recording gimbal angles when rolled past nadir in the backward direction. 

All pitch angles are less than or equal to zero, with zero being nadir. When the gimbal physically rolls past nadir, images are recorded upside-down, roll is recorded as 180, gimbal heading is turned 180º, and pitch is nadir (0) - angle (pitch is consistent with forward gimbal. 

Tested and works with a Mavic 3E.

### DIFF
--- a/opendm/photo.py
+++ b/opendm/photo.py
@@ -471,6 +471,15 @@ class ODM_Photo:
                     # Roll: 0 (assuming gimbal)
                     if self.has_ypr():
                         if self.camera_make.lower() in ['dji', 'hasselblad']:
+                            
+                            # Mavic 3E smart oblique backward-facing images need special treatment
+                            # backward-facing images are identified with a 180º roll
+                            if abs(self.roll) > 90:
+                               # Camera faces opposite direction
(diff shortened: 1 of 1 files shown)


thadwald:
While this appeared to have worked with small gimbal angles, it broke with larger oblique angles.

Saijin-Naib:
Do you think it is possible to solve for both small and large angles?

thadwald:
Yes. I believe I have a fix. But I want to test it with more datasets.

There were two issues:
1) yaw was read from aircraft instead of gimbal. Gimbal is world-stabilized on all three axis so it is correct to use gimbal yaw if there is one. I changed it to prefer gimbal yaw and fall back on on aircraft if it does not exist

2) The transformation from gimbal yaw, pitch, roll to omega phi kappa required by OpenSFM explicitly ignored gimbal roll and assumed it remains zero.  I changed the transformation matrix to include all three axis.

PR #1972, opened 2025-12-23
Fix subprocess venv activation for split-merge submodel execution on Linux/Mac

thadwald:
When running split-merge with custom image groups, submodel processes fail with
`ModuleNotFoundError: No module named 'dateutil'` and other venv-installed packages.

`get_submodel_argv()` in `osfm.py` uses `sys.argv[0]` (typically `/code/run.py`) as the
startup script for spawning submodel processes. The function calls`run.py` directly, bypassing venv activation.

This fix adds handling for Linux/Mac to use `run.sh` when available, mirroring the
existing Windows logic for `run.bat`. This ensures the venv is properly activated
before Python executes, making all installed packages available to subprocesses.

- Tested split-merge with image_groups.txt on WebODM Docker (latest)
- Submodels now complete successfully through all pipeline stages except for report (different issue).

### DIFF
--- a/opendm/osfm.py
+++ b/opendm/osfm.py
@@ -669,6 +669,13 @@ def get_submodel_argv(args, submodels_path = None, submodel_name = None):
     if sys.platform == 'win32':
         startup_script_dir = os.path.dirname(startup_script)
         startup_script = os.path.join(startup_script_dir, "run")
+    else:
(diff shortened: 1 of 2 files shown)


thadwald:
Closing — this issue was resolved by #1958, which solved it by removing the shebang from run.py and ensuring explicit Python invocation for subprocesses.

PR #1977, opened 2026-01-12
Fix camera angle(omega,phi,kappa) calculation

originlake:
This should be a bug in camera angle calculation, leading to bad performance in triangulation reconstruction method. 
The issue is discussed here. I'm not 100% sure if the fix meets the goal of the original design, but at least the fix would make it mathematically correct. Wish to get some input for this. 
https://github.com/mapillary/OpenSfM/issues/1098

### DIFF
--- a/opendm/photo.py
+++ b/opendm/photo.py
@@ -931,7 +931,14 @@ class ODM_Photo:
             # Unit vector pointing north
             xnp /= m
 
-            znp = np.array([0, 0, -1]).T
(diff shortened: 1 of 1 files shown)


YanNoun:
Hi,

Thank you for pushing this ! Maybe it will be better to merge my OpenSfM fork (or cherry pick relevant commits), so that the fixes is directly in OpenSfM.

Cheers,

Yann

Issue #1979, opened 2026-01-13
bug: Error Running ODM with stacked multispectral images


maurerle:
When reconstructing a multi-channel input from a MicaSense RedEdge-P camera, I get the following error creating the texture:
`Image must be RGB or RGBA`

Eventually, it is possible to skip this step or remove the additional channels for the creation of the texture.

Otherwise this upstream issue discusses if the fix should be placed in this repo or in the texrecon tool:
~~https://github.com/simonfuhrmann/mve/issues/584~~

https://github.com/nmoehrle/mvs-texturing/issues/229
This is the log of the odm process:
```
[INFO]    running "/code/SuperBuild/install/bin/OpenMVS/ReconstructMesh" -i "/datasets/images_proc/odm_meshing/odm_25dmesh.dirty.ply" -o "/datasets/images_proc/odm_meshing/odm_25dmesh.ply" --archive-type 3 --remove-spikes 0 --remove-spurious 0 --smooth 0 --target-face-num 400000 -v 0
15:36:43 [App     ] OpenMVS x64 v2.2.0
15:36:43 [App     ] Build date: Jan 13 2026, 15:27:48
15:36:43 [App     ] CPU: AMD Ryzen AI 9 365 w/ Radeon 880M               (20 cores)
15:36:43 [App     ] RAM: 61.93GB Physical Memory 63.12GB Virtual Memory
15:36:43 [App     ] OS: Linux 6.17.13+deb13-amd64 (x86_64)
15:36:43 [App     ] Disk: 
15:36:43 [App     ] SSE & AVX compatible CPU & OS detected
15:36:43 [App     ] Command line: ReconstructMesh -i /datasets/images_proc/odm_meshing/odm_25dmesh.dirty.ply -o /datasets/images_proc/odm_meshing/odm_25dmesh.ply --archive-type 3 --remove-spikes 0 --remove-spurious 0 --smooth 0 --target-face-num 400000 -v 0
15:36:44 [App     ] MEMORYINFO: {
15:36:44 [App     ] 	VmPeak:	  200708 kB
15:36:44 [App     ] 	VmSize:	  117800 kB
15:36:44 [App     ] } ENDINFO
[INFO]    Finished odm_meshing stage
[INFO]    Running mvs_texturing stage
[INFO]    Writing MVS Textured file in: /datasets/images_proc/odm_texturing/odm_textured_model_geo.obj
[...]

PR #1980, opened 2026-01-13
allow images with more than 4 channels in get_image_size and read_image

maurerle:
starting an ODM job with stacked images (tif images containing all 6 channels of a multispectral camera) fails currently as the libraries used to `get_image_size` do not allow usage of more than 4 channels.

PIL does silently fail for some reason, while the fallback cv2 fails with:

`/code/SuperBuild/src/opencv/modules/imgcodecs/src/grfmt_tiff.cpp:148: error: (-215:Assertion failed) channels <= 4 in function 'normalizeChannelsNumber'`

Rasterio does a good job opening the TIF files without actually reading them.

With this fix, we are one step further processing multi-channel images, but then run into the error described in #1979

### DIFF
--- a/opendm/ai.py
+++ b/opendm/ai.py
@@ -6,14 +6,21 @@ import time
 import sys
 import rawpy
 import cv2
+from pathlib import Path
+import rasterio as rio
 
 def read_image(img_path):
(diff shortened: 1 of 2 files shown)


smathermather:
Thanks @maurerle,

Do you have an example dataset for testing?

PR #1995, opened 2026-02-24
Fixing CI - Compiling GDAL for non-windows

DodgySpaniard:


snap/snapcraft.yaml
```diff
@@ -137,12 +126,30 @@ parts:
...
+    build-packages:
+      - libgeotiff-dev
+      - libhdf5-serial-dev
+      - libopenmpi-dev
+      - pybind11-dev
+      - python3-dev
+      - python3-numpy
+      - python3-setuptools
+      - swig
+    stage-packages:      - libcurl3t64-gnutls      - libgeotiff5
```
spwoodcock: Small formatting issue? I don't think we test with the `snapcraft.yaml` and use `snapcraft24.yaml`

gpu.Dockerfile
```diff
@@ -25,11 +25,11 @@ RUN bash configure.sh clean
 
 ### Use a second image for the final asset to reduce the number and
 # size of the layers.
-FROM nvidia/cuda:13.0.0-devel-ubuntu24.04
+FROM nvidia/cuda:12.9.1-devel-ubuntu24.04
```
spwoodcock: I wonder if we can save a bit of space using the `nvidia/cuda:12.9.1-runtime-ubuntu24.04` image instead of devel here

portable.Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:24.04 AS builder
 
 # Env variables
 ENV DEBIAN_FRONTEND=noninteractive \
-    PYTHONPATH="$PYTHONPATH:/code/SuperBuild/install/lib/python3.12/dist-packages:/code/SuperBuild/install/bin/opensfm" \
+    PYTHONPATH="$PYTHONPATH:/code/SuperBuild/install/local/lib/python3.12/dist-packages:/code/SuperBuild/install/lib/python3.12/dist-packages:/code/SuperBuild/install/bin/opensfm" \
```
spwoodcock, line 5: Does the standard `Dockerfile` need to also be updated to use `local/lib/python3.12/dist-packages` too (now that GDAL goes there)?

Looks like:
gdal --> /code/SuperBuild/install/local/lib/python3.12/dist-packages
pdal --> /code/SuperBuild/install/lib/python3.12/dist-packages
opensfm --> /code/SuperBuild/install/bin

spwoodcock (approved):
Looks like a much better way that unstable ppas!

No doubt the build is a little longer, but stability is more important.

I managed to build the portable image locally without issue 👍 (on AMD64 arch, Debian machine, final image is 1.62GB)

snap/snapcraft.yaml
DodgySpaniard: Good catch, fixing it.

In general, I am a bit puzzled about the whole snapcraft thing. It would seem that it is only used to pull the dependencies during configure.sh, and in our CI only Ubuntu 24.04 is used...

gpu.Dockerfile
DodgySpaniard: Let me try that.

portable.Dockerfile
DodgySpaniard: Probably, I actually forgot about that one. Let me dig around.

snap/snapcraft.yaml
spwoodcock: Me too! 😆

I guess its useful for anyone wishing to use snap, but my personal preference would be Debian base images and apt packages defined in the Dockerfile

snap/snapcraft.yaml
```diff
@@ -137,12 +126,31 @@ parts:
...
+      - libgeotiff-dev
+      - libhdf5-serial-dev
+      - libopenmpi-dev
+      - pybind11-dev
+      - python3-dev
+      - python3-numpy
+      - python3-setuptools
+      - swig
+    stage-packages:      
+      - libcurl3t64-gnutls      - libgeotiff5
```
spwoodcock, line 142: Sorry to be a pain - seems like one was missed!

I wonder if snapcraft is smart enough to still parse the yaml despite the syntax issue

snap/snapcraft.yaml
pierotofy: You can/should remove the snapcraft stuff if you get a chance. The project used to have snap packaging support, but that's no longer the case.

PR #1997, opened 2026-03-08
Installing python dependencies last

DodgySpaniard:
Python dependencies (throug PIP) require gdal, moving python install after superbuild. Keeping Numpy (to compile opnecv).

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -131,9 +131,16 @@ installreqs() {
 
     # edt requires numpy to build
     venv/bin/pip install numpy==2.3.2
+    set +e
+}
+
+installpython() {
+    echo "Installing Python requirements with compiled GDAL"
+    cd /code
+    export GDAL_CONFIG=${RUNPATH}/SuperBuild/install/bin/gdal-config
+    
+    set -e
     venv/bin/pip install -r requirements.txt --ignore-installed
-    #if [ ! -z "$GPU_INSTALL" ]; then
-    #fi
     set +e
 }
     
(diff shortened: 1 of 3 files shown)


DodgySpaniard:
Tested on ARM build: buildx build --file portable.Dockerfile --platform linux/arm64

spwoodcock (approved):
Lgtm 👍

smathermather:
Maybe an extant bug from https://github.com/OpenDroneMap/ODM/pull/1995, PDAL needs GEOS support or else split merge breaks as PDAL cannot generate the hexbin-based boundary:

```
pdal info --boundary --filters.hexbin.edge_length=1 --filters.hexbin.threshold=1 "/datasets/code/submodels/submodel_0000/odm_georeferencing/odm_georeferenced_model
.decimated.las" > "/datasets/code/submodels/submodel_0000/odm_georeferencing/odm_georeferenced_model.boundary.json"
PDAL: PDAL must be using a version of GDAL built with GEOS support to use this function.

[WARNING] Cannot calculate crop bounds! We will skip cropping!
```

Otherwise, I don't see any issues.

To replicate, download [Brighton Beach](https://github.com/pierotofy/drone_dataset_brighton_beach_split_merge/archive/master.zip), then run with the following options: `--split 8 --split-overlap 20`.

smathermather (approved):

DodgySpaniard:
I added the dependencies geos and expat to the system. GDAL CMake uses find_package and uses them by default if found.

As a side note, it is a good gotcha using the split settings, taking notes as we need to define a test process for releases.

PR #1998, opened 2026-03-09
Add Azure Artifact Signing support

sylveon:
This should work, in theory. Not tested because I do not have the setup for GitHub Actions auth to Azure on my side.

Before this PR is merged, it requires adding a new secret beyond the existing ones: AZURESIGNING_METADATA, which contains a base64 encoded version of the Artifact Signing metadata.json file. See [here](https://learn.microsoft.com/en-us/azure/artifact-signing/how-to-signing-integrations#create-a-json-file) for how to create such a file.

### DIFF
--- a/.github/workflows/publish-windows.yml
+++ b/.github/workflows/publish-windows.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     runs-on: windows-2022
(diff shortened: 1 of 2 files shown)


Saijin-Naib (approved):
Looks great.

PR #2008, opened 2026-03-20
Keep dense reconstruction topocentric while applying exact final georeferencing

lupionpe:
## Summary
- keep the dense reconstruction pipeline in topocentric coordinates instead of swapping the projected reconstruction into `reconstruction.json`
- export geocoords side artifacts separately and apply the exact local-to-projected conversion to the final LAZ, textured OBJ, and `shots.geojson` outputs
- avoid the OpenMVS regressions caused by feeding projected reconstruction coordinates into the dense pipeline while preserving the improved georeferencing in the final outputs

## Companion OpenSfM PR
This PR is the ODM-side integration for the companion OpenSfM fix here:
- https://github.com/OpenDroneMap/OpenSfM/pull/48

The core georeferencing solution was proposed and implemented by Yann Noutary in:
- https://github.com/YanNoun/OpenSfM/commit/446a4b06624f38f7dc9922cf737fd3e806e75636

Credit for the root fix is his. This PR is the downstream ODM integration needed to make that corrected transform reach the final orthoimage / point cloud outputs without breaking dense reconstruction.

I have some basic coding experience, but most of the porting / validation / integration work for this contribution was done with AI (Codex).

## Context
Related discussions:
- https://github.com/OpenDroneMap/ODM/pull/1851
- https://community.opendronemap.org/t/what-is-your-biggest-pain-with-odm-opensfm/25791/22
- https://community.opendronemap.org/t/gcp-show-very-good-results-but-is-not-what-i-see-in-reality/25541

## Validation
My conclusion after downstream validation is that this does seem to solve the root cause.

I ran two real tests on two different areas with roughly:
- ~1,600,000 m² survey area
- ~2000 images
- `gcp.txt` + `geo.txt`
- RTK on the drone (Mavic 3E)

In both cases the observed precision improved, as expected.

Checks I could run locally here:
- [x] `python3 -m py_compile opendm/types.py opendm/osfm.py opendm/shots.py opendm/align.py stages/run_opensfm.py stages/odm_georeferencing.py stages/odm_report.py`
- [x] downstream end-to-end validation in a custom ODM / NodeODM deployment on Debian 12
- [ ] native ODM unit tests are not runnable in this shell because the local environment here is missing ODM's Python dependencies

### DIFF
--- a/opendm/align.py
+++ b/opendm/align.py
@@ -5,6 +5,7 @@ import codem
 import dataclasses
 import pdal
 import numpy as np
+from numpy.lib import recfunctions as rfn
 import rasterio
 from rasterio.crs import CRS
 from opendm.utils import double_quote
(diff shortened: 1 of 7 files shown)


jrstear:
This works for me — great work!

End-to-end test on a 6 km aztec highway corridor (1385 DJI M3E images,
41 GNSS control points, 3° off UTM central meridian). Compared v3.6.0
stock against v3.6.0 with this PR + companion
[OpenDroneMap/OpenSfM#48](https://github.com/OpenDroneMap/OpenSfM/pull/48):

- **CHK RMS_H: 1.44 ft → 0.28 ft** (~5× better)
- **GCP RMS_H: 1.62 ft → 0.08 ft** (~20× better)
- **Exceeds Pix4D on the same dataset** (0.28 vs 0.72 ft CHK)
- **Reconstruction RMS_H unchanged** — the keep-dense-topocentric design works;
  only the final projection changes, OpenMVS/texturing/meshing are unaffected

Full report with charts + per-point data:
https://github.com/jrstear/geo-samples/tree/main/odm-ortho-error

Thanks to @YanNoun for the core solution this builds on.
Strongly recommend merging.

PR #2013, opened 2026-03-26
Minimum dev container setup

MJohnson459:
A [devcontainer](https://containers.dev/) is a small wrapper around Docker to make it easy to set up a clean development environment. Most IDE's and code editors support these now as well as many cloud environments.

I've included a very simple environment to start with which we can extend as necessary later to streamline the build process.

To test this I:
- pulled the repo and changed to this branch
- "opened in devcontainer" when prompted (works in vscode, zed, cursor)
- ran `./configure.sh install`
- pulled the `odm_data_aukerman` dataset into `datasets/odm_data_aukerman`
- ran `./run.sh --project-path datasets odm_data_aukerman`

This all ran successfully with no critical errors.

(note I split out the fixes to the build into https://github.com/OpenDroneMap/ODM/pull/2022 to keep feedback separate)


MJohnson459:
Sorry I moved this back to a draft until I have

1. Added GPU support
2. Updated documentation for dev workflows

MJohnson459:
Ok, I've updated this PR to include both a CPU and GPU version of the devcontainer, both of which now point to the existing respective dockerfiles. I've tried to keep changes on the docker front to a minimum to avoid interfering with existing workflows, as such the final docker images should be unchanged.

As this is intended to be a new and simplified developer setup, I've removed the start-dev-env script which basically did a similar thing. I've also updated the README.md with details on how to use this.

I've tested this on both Linux and Windows but it should work on Mac as well. As a known issue, I did find it took significantly longer to build on Windows due to Docker on Windows and the mixed Windows/WSL filesystem, but I don't believe this is a regression from the previous docker based method.

I haven't updated the [hosted docs](https://docs.opendronemap.org/tutorials/#development-and-testing-of-odm) nor looked at NodeODM as part of this work. This should be a standalone improvement for ODM which we can then copy to NodeODM workflows.

smathermather:
Sounds good. So we'll have to T-up NodeODM workflows and docs update and merge in one go then. I'll dive in on testing.

smathermather:
Followed the directions here on devcontainer cli install: https://github.com/devcontainers/cli

Then:
```bash
git clone https://github.com/MJohnson459/ODM
cd ODM
git checkout devcontainer
devcontainer up --workspace-folder .
```
I do get one warning:
``` => WARN: InvalidDefaultArgInFrom: Default value for ARG $BASE_IMA  0.0s```

I'll also give it a run with vscode and take a look on a Mac.

PR #2022, opened 2026-04-20
fix: prevent failures on clean Linux builds

MJohnson459:
While trying to build the project from clean there were a few issues I encountered:

- Ceres didn't build due to the `git patch` command failing.
- GDAL didn't install the python bindings in the venv
- When not using the Docker build the `/code` directory doesn't exist which was causing the configure.sh script to fail

With these changes I can consistently build the project from scratch. I've tested this by using a clean Docker build and also with a [devcontainer](https://github.com/OpenDroneMap/ODM/pull/2013) (both using ubuntu 24.04).

I believe these changes shouldn't affect windows and mac builds but I don't have machines to test those at the moment.


smathermather:
Shouldn't affect Windows and Mac, but double-checking. Built on linux and ran through oats as a quick check. Will take a look at outputs in detail later tonight, but cursory checks are good.

sylveon (approved):
Tested on Windows, everything still works there!

smathermather:
Cool. Works better on linux than 2.5.6 as well.

PR #2033, opened 2026-06-17
chore: contrib code cleanup

maurerle:
this cleans up some code smells I saw when looking at the contrib scripts


smathermather:
Looks good. Thanks!

PR #2035, opened 2026-06-19
orthophoto merge: parallelize the block loop (in-order writer)

Chouffe:
## Summary

Parallelizes the per-block blend loop in `merge()`, behind a `max_workers` parameter (default `1` = unchanged serial behavior), wired from `stages/splitmerge.py` as `args.max_concurrency`.

> **Stacked on #2036** (gated reads). This PR currently shows both commits; once #2036 merges it reduces to just the parallelization. The gating is a prerequisite — see below.

## Why it needs the gated reads (#2036)

The block loop is embarrassingly parallel, but a naive thread pool stalls: rasterio's `boundless=True` reads serialize a VRT (`_serialize_xml`) on every read, which under concurrency dominates and effectively hangs the merge (workers park in `_serialize_xml`). #2036 replaces those with gated reads (no VRT for in-bounds windows), which makes a parallel loop viable.

## Parallelization

- **Parallel compute + single in-order writer.** Blocks are computed in a `ThreadPoolExecutor`; one thread writes them in strict block order with a bounded look-ahead (`cap = 2 * max_workers`). A naive "write each block as it finishes" version can't flush incrementally — GDAL needs writes in row-major (block) order — so dirty blocks accumulate in RAM until it OOMs; the in-order writer keeps writes sequential/flushable and memory small.
- **Bounded GDAL block cache during the merge** (restored on exit) so dirty-tile eviction/flush under GDAL's global lock stays prompt.
- **Per-thread source handles** — GDAL/rasterio datasets are not thread-safe.
- Preserves `--merge-skip-blending`.

`max_workers <= 1` is a plain serial compute-then-write path, byte-for-byte identical to the original loop.

## Correctness

- **Deterministic across worker counts.** On a real 15.9 Gpx survey (3 submodels, ~61k blocks), the merged orthophoto is **byte-for-byte identical** at `max_workers=1` (serial, ~28 min) and `max_workers=16` (~8 min) — the in-order writer makes the result independent of worker count, and the serial/default path completes cleanly with no read↔write stall.
- **End-to-end.** Verified on the same survey: the merge completes with **zero stalled blocks** and produces a valid orthophoto, pixel-identical to the serial baseline (confirmed for both LZW-compressed and uncompressed output).

## Notes

- No new dependencies.
- Default behavior unchanged (`max_workers=1`).

### DIFF
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -265,6 +265,28 @@ def feather_raster(input_raster, output_raster, blend_distance=20):
 
         return output_raster
 
+def _read_window_gated(ds, src_window, dst_shape, dtype):
+    """Read ``src_window`` into a ``dst_shape`` array — equivalent to a
+    ``boundless=True`` read with 0 nodata fill, but avoiding rasterio's boundless
+    VRT path for the common cases. ``boundless=True`` builds a VRT and serializes
+    it via Python's ElementTree (``_serialize_xml``) on every read, which is a
+    large per-read overhead on big merges. Behaviour:
+      - window fully outside the dataset -> all zeros (no read), matching the 0
+        nodata fill a boundless read would produce here;
+      - window fully inside -> plain non-boundless read (no VRT), identical to a
+        boundless read when no out-of-bounds padding is needed;
+      - window partially overlapping the edge -> fall back to boundless (rare;
+        only the true border blocks), preserving exact fill behaviour.
+    """
+    (r0, r1), (c0, c1) = src_window
+    out = np.zeros(dst_shape, dtype=dtype)
+    height, width = ds.height, ds.width
+    if r1 <= 0 or c1 <= 0 or r0 >= height or c0 >= width:
+        return out
+    if r0 >= 0 and c0 >= 0 and r1 <= height and c1 <= width:
+        return ds.read(out=out, window=src_window, boundless=False, masked=False)
+    return ds.read(out=out, window=src_window, boundless=True, masked=False)
+
 def merge(input_ortho_and_ortho_cuts, output_orthophoto, orthophoto_vars={}, merge_skip_blending=False):
     """
     Based on https://github.com/mapbox/rio-merge-rgba/
@@ -361,10 +383,7 @@ def merge(input_ortho_and_ortho_cuts, output_orthophoto, orthophoto_vars={}, mer
                         src.transform, right, bottom, op=round, precision=precision
                     )))
 
-                temp = np.zeros(dst_shape, dtype=dtype)
-                temp = src.read(
-                    out=temp, window=src_window, boundless=True, masked=False
-                )
+                temp = _read_window_gated(src, src_window, dst_shape, dtype)
 
                 # pixels without data yet are available to write
                 write_region = np.logical_and(
@@ -390,10 +409,7 @@ def merge(input_ortho_and_ortho_cuts, output_orthophoto, orthophoto_vars={}, mer
                         src.transform, right, bottom, op=round, precision=precision
                     )))
 
-                temp = np.zeros(dst_shape, dtype=dtype)
-                temp = src.read(
-                    out=temp, window=src_window, boundless=True, masked=False
-                )
+                temp = _read_window_gated(src, src_window, dst_shape, dtype)
 
                 where = temp[-1] != 0
                 for b in range(0, num_bands):
@@ -414,10 +430,7 @@ def merge(input_ortho_and_ortho_cuts, output_orthophoto, orthophoto_vars={}, mer
                         cut.transform, right, bottom, op=round, precision=precision
                     )))
 
-                temp = np.zeros(dst_shape, dtype=dtype)
-                temp = cut.read(
-                    out=temp, window=src_window, boundless=True, masked=False
-                )
+                temp = _read_window_gated(cut, src_window, dst_shape, dtype)
 
                 # For each band, average alpha values between
                 # destination raster and cut raster
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -27,7 +27,7 @@ def _bounded_gdal_cache(nbytes):
     """Temporarily cap GDAL's global block cache, restoring it on exit.
 
     A small cache keeps the parallel merge's output-tile flushes prompt and
-    cheap. Restoring on the way out — even if the merge raises — avoids leaving
+    cheap. Restoring on the way out (even if the merge raises) avoids leaving
     the rest of the pipeline (notably the COG conversion) with a shrunken cache.
     """
     prev = gdal.GetCacheMax()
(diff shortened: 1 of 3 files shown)


spwoodcock:
Thanks @Chouffe - this could be a really nice change! 😁

Before I do a full review, could you help me with two things please:

1. You say the outputs are unchanged from before. Which test dataset did you use to verify? Could you provide any logs of the successful run?

2. I don't think ODM has a policy set for LLM-assisted PRs needing to be declared (yet). For full transparency, was every line written by you, or was this assisted?

Thanks a lot for the contribution! 🙌

Saijin-Naib:
Our policy, as it stands currently, id here:  
https://github.com/OpenDroneMap/documents/blob/master/CONTRIBUTING.md

spwoodcock:
> Our policy, as it stands currently, id here:  
> https://github.com/OpenDroneMap/documents/blob/master/CONTRIBUTING.md

Thank you @Saijin-Naib! 🙏

I might make a little PR to ensure thats linked in the pull request template, so devs are definitely made aware when they create a new PR =)

Chouffe:
@spwoodcock thanks for taking the time to review this PR!

For context: we’re using ODM on large drone surveys and letting users upload flight data to an online platform. We’ve already tuned ODM for our bird-monitoring domain so it runs reliably on AWS ECS Fargate. Even so, we hit one remaining issue: a large stress-test survey would deadlock after ~7 hours of compute.

To address this, I ran ODM from a pinned base Docker image (v3.6.0) with the relevant runtime patches baked in. With these changes, our very large surveys complete successfully without the earlier hiccups.

After extended investigation, the deadlock turned out to be caused by a combination of compounding factors:

1. rasterio “boundless” mode becoming pathologically slow as the GDAL cache fills (including occasional read stalls). The fix is in a companion ODM PR: https://github.com/OpenDroneMap/ODM/pull/2036

2. The merge step running single-threaded, which is a significant bottleneck when multiple cores are available. We can speed it up with parallelism, while still ensuring GDAL writes blocks in the correct order to avoid write stalls.


To validate the merge changes, I reproduced the merge path in isolation by generating three orthophotos (via ODM’s split phase) and then running the merge step across those three outputs. I also confirmed the merged results are byte-identical whether run serially or with a high worker count.

The sub-orthophotos are large (≈5–10GB each), so if you’d like to verify that the read stalls occur in the problematic scenario, tell me what’s the best way to share artifacts with you.

Finally, this work was supported by an LLM, and the investigation itself took about a week of iterative runs to reproduce the exact prod behavior.

opendm/orthophoto.py
```diff
@@ -338,33 +401,70 @@ def merge(input_ortho_and_ortho_cuts, output_orthophoto, orthophoto_vars={}, mer
...
+        def get_sources():
+            """Return this thread's (ortho, cut) rasterio dataset handles.
+
+            GDAL/rasterio handles are not safe to share across threads, so each
+            worker thread lazily opens and caches its own set on first use and
+            registers it in opened_sources for cleanup after the parallel run.
+            """
+            srcs = getattr(tls, "sources", None)
+            if srcs is None:
+                srcs = [(rasterio.open(o), rasterio.open(c)) for o, c in inputs]
```
spwoodcock: This will work perfectly for smaller projects, but larger projects could potentially run into file handle issues.

Each worker will open an ortho + cut pair, so file opens scale as `2 * pairs * max_workers`.

Since splitmege.py now passes args.max_concurrency, **split-merge now becomes concurrent by default for end users**.

An example would be approx 20 submodels to merge, on a machine where 32 workers are possible. This would result in 1280 file handles, and cause an EMFILE error (too many open).

I consdered capping the max_workers` based on RLIMIT_NOFILE from the filesystem, and the number of input pairs. But this would not be cross platform to other OSes like Windows...

There are probably diminishing returns to the parallelism of a merge (I'm making an assumption here), so perhaps we could just implement a hard cap to the number of parallel workers? This would reduce risk of issues 🙏 

Something like:

```python
# Number of submodels to merge at once is bounded by GDAL's global cache lock,
# so parallelism only goes so far.
# We cap workers to still retain a meaningful speedup,
# but stay well under typical RLIMIT_NOFILE of many distros                                                                                                                 
MERGE_WORKER_CAP = 16

if max_workers > MERGE_WORKER_CAP:
    log.ODM_INFO(
                "Capping orthophoto merge workers %d -> %d to limit open file handles"
		% (max_workers, MERGE_WORKER_CAP)
	)
	max_workers = MERGE_WORKER_CAP
```

What do you think?

spwoodcock: One other small change, for the same line, we could also update to `(rasterio.open(o), None if merge_skip_blending else rasterio.open(c))` to avoid opening cut rasters in that mode (and guard the `c.close()` in the cleanup loop for the None case) 😄

spwoodcock (changes requested):
This is really nice - I only left two comments, for the same line of code - hopefully useful 😄

Chouffe:
Thanks for the review @spwoodcock, both are good catches. Addressed in the latest push:

**1. File-handle bound.** Agreed on the EMFILE risk now that split-merge passes `max_concurrency` by default. I went with your hard-cap suggestion rather than an `RLIMIT_NOFILE` probe, for the cross-platform reason you gave:

```python
MERGE_WORKER_CAP = 16

if max_workers > MERGE_WORKER_CAP:
    log.ODM_INFO("Capping orthophoto merge workers %d -> %d to limit open file handles"
                 % (max_workers, MERGE_WORKER_CAP))
    max_workers = MERGE_WORKER_CAP
```

The cap is applied once inside `merge()`, so it protects every caller. This is a pure concurrency bound, so it doesn't change the output (blocks are computed independently and written in strict order regardless of worker count) and the byte-identical guarantee still holds.

**2. Skip cut opens in skip-blending mode.** Done. The cut rasters are only read in the third (blend) pass, which is skipped in that mode, so there's no reason to open them:

```python
srcs = [(rasterio.open(o), None if merge_skip_blending else rasterio.open(c))
        for o, c in inputs]
```

The cleanup loop now guards the `None` handle.

Chouffe:
@spwoodcock let me know if there is anything else I can do for this PR to get merged.

spwoodcock (approved):
Good to merge imo! Thanks again 🎉

Chouffe:
Great to hear! I can't merge myself as I don't have the permissions on this repo. Looking forward to seeing this change land in ODM 🎉

Chouffe:
Hey ODM team, is there anything left to do to merge this PR? Happy to support you in any way.

spwoodcock:
I'm just a bit unsure on who should merge PRs in the repo - I only recently onboarded and am helping out ad-hoc, so don't want to overstep.

I have reached out via DM to someone that will know - will get back to you soon!

smathermather:
> Hey ODM team, is there anything left to do to merge this PR? Happy to support you in any way.

Looks good on my end. Just confirming -- AI use was an augmentation of your work, but it was not prompt engineering? I realize we have a long way to go in communicating various levels of AI use, so just some detail on how AI was used would help.

Sorry for the delay. I really appreciate this pull request -- definitely resolves the mystery of the problems with merges.

spwoodcock:
From the latest https://responsibleai.guide/managing-ai-contributions/

You could use this template, which we could consider adding to the PR template 🙏

## AI Tool Usage

How much of this PR was AI-assisted? (check one)

- [ ] **0** - No AI at any point
- [ ] **1** - AI helped me think, but I wrote all the code myself
- [ ] **2** - I planned the change and decided the approach; AI helped write the code
- [ ] **3** - AI planned and wrote it; I checked and approved each step as it went
- [ ] **4** - I set the AI going and left it to it; I reviewed the finished result
- [ ] **5** - I set the AI going and left it to it; nobody has read the result - no review, or AI review only

Chouffe:
AI use was an augmentation of my work indeed. So I would say number 2 in your template.

Here is some extra context:

I have spent about a week investigating this ODM failure in our production environment. I was able to set up a repro case which would consistently fail. Then I brainstormed with the agent what the problems could be and it helped me navigate and learn the ODM codebase much faster than if I had to do it without augmentation. 

I steered the AI towards the right path and it helped me locate the exact location in the codebase where it would hang. 
After deciding on the approach to fix it, AI helped me update the codebase and I could easily check it worked because I had a repro case at hand that should pass once fixed.

Thanks again for taking the time to look at this PR. We have internally patched the ODM codebase until it lands in your main branch and gets released. It has been running without problems for the past months with this patch.

PR #2036, opened 2026-06-23
orthophoto merge: avoid boundless-read VRT serialization (gated reads)

Chouffe:
## Summary

`merge()` reads every source window with `boundless=True`. rasterio's boundless mode builds an in-memory VRT and serializes it via Python's `ElementTree` (`_serialize_xml`) on **every** read — a large per-read overhead on large merges (a 15.9 Gpx mosaic is ~61k output blocks, each doing 3 passes over the submodel rasters).

`_read_window_gated()` produces identical output but skips the VRT for the common cases:
- window **fully inside** the source → plain non-boundless read (no VRT);
- window **fully outside** → zeros (identical to the `0` nodata fill boundless produces there);
- window **straddling the edge** → fall back to `boundless=True` (rare; only the true border blocks).

## Correctness

**Pixel-identical.** An oracle sampled hundreds of fully-in-bounds windows across a real merge grid and compared a `boundless=True` read against the plain read into a zeroed array — **0 mismatches**.

## Scope

Serial; no behavior change beyond the speedup. This also unblocks parallelizing the block loop (follow-up PR): under thread concurrency the per-read VRT serialization becomes pathological (workers park in `_serialize_xml`), so gating the reads is a prerequisite.

### DIFF
--- a/opendm/orthophoto.py
+++ b/opendm/orthophoto.py
@@ -265,6 +265,28 @@ def feather_raster(input_raster, output_raster, blend_distance=20):
 
         return output_raster
 
+def _read_window_gated(ds, src_window, dst_shape, dtype):
+    """Read ``src_window`` into a ``dst_shape`` array — equivalent to a
+    ``boundless=True`` read with 0 nodata fill, but avoiding rasterio's boundless
+    VRT path for the common cases. ``boundless=True`` builds a VRT and serializes
+    it via Python's ElementTree (``_serialize_xml``) on every read, which is a
+    large per-read overhead on big merges. Behaviour:
+      - window fully outside the dataset -> all zeros (no read), matching the 0
+        nodata fill a boundless read would produce here;
+      - window fully inside -> plain non-boundless read (no VRT), identical to a
+        boundless read when no out-of-bounds padding is needed;
+      - window partially overlapping the edge -> fall back to boundless (rare;
+        only the true border blocks), preserving exact fill behaviour.
+    """
+    (r0, r1), (c0, c1) = src_window
+    out = np.zeros(dst_shape, dtype=dtype)
+    height, width = ds.height, ds.width
+    if r1 <= 0 or c1 <= 0 or r0 >= height or c0 >= width:
+        return out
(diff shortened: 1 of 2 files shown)


spwoodcock (approved):
Nice change, useful docstring as it's not an obvious one 👍 

Some unit tests would be lovely, but overall this is well scoped and looks good to me!

Small thing: as with the other PR, it would also be great if you could indicate how this was tested / results / logs, and also to what extent AI was used to assist 🙏 

A minor related nitpick would be the use of em-dash `—`, a common sign of AI-gen text, for which I don't think we have any in the repo until this point!

Chouffe:
Thanks for the approval. Addressed the two follow-ups:

- Dropped the em-dash from the `_read_window_gated` docstring (`0737c13`). It was the only one in the repo, so good spot.
- Added `tests/test_orthophoto.py` (`6773d00`): 9 unit tests asserting the gated read is pixel-identical to a `boundless=True` read across all three branches (fully inside, fully outside each edge, and partial-edge overlaps). They run under the ODM image's rasterio/GDAL with `python3 -m unittest tests.test_orthophoto`.

Chouffe:
For context on where this came from: this change originally lived inside #2035 (the parallel orthophoto merge). The gated-read fix stands on its own, though. It avoids the boundless-read VRT serialization and speeds up the existing serial merge without changing its output, independent of any parallelization. So I pulled it out into this smaller, standalone PR to make it easier to review, and #2035 now stacks on top of it.

Suggested review/merge order is this one first, then #2035.

Chouffe:
@spwoodcock let me know if there is anything left I can address for this PR to get in.

spwoodcock (approved):
Thanks for the tests 👍 

Looks good to me!

Chouffe:
Great to hear! I can't merge myself as I don't have the permissions on this repo. Looking forward to seeing this change land in ODM 🎉

Chouffe:
Hey ODM team, is there anything left to do to merge this PR? Happy to support you in any way.

Issue #1960, opened 2025-11-26
Do we need to maintain support for old x86_64 CPUs (pre-2010)? i.e. Should we require AVX?


spwoodcock:
## The current state of things

- We currently use bundled `gcc` and `g++` wrappers.
- The wrapper replaces the system gcc/g++ with a script that:
  - Strips out any `-march=*` flags coming from the build system.
  - Injects its own `-march=<ARCH>` where:
    - `nehalem` is forced for x86_64 (very old — 2008).
    - `armv8-a` is forced for aarch64.
  - Delegates compilation to `gcc_real`.
- This makes ODM run on very old CPUs without AVX / AVX2.
- The result is a portable, reproducible binary that runs on essentially any CPU.

## The issues this causes

- After upgrading to Ubuntu 24.04 and updating many libraries, the standard Dockerfile builds fine (no wrappers).
- But `portable.Dockerfile` and `gpu.Dockerfile` both use the wrappers - these are the production images.
- These now fail to build. Likely caused by breaking OpenCV, OpenMVS, or OpenSfM.
- The wrapper overrides what CMake detects:
  - CMake sees AVX/AVX2 support on the build machine
  - But the wrapper forces Nehalem
  - Result: build errors, intrinsic mismatches, inline asm failures - making things harder to maintain

## The question

- Do we still need the latest ODM versions to run on pre-2010 CPUs?
- Can we EOL `3.5.6` as the last "legacy CPU" version, and release `4.0.0` with a modern CPU baseline?
- Notes:
  - Nearly all cloud servers, CI runners, and modern machines support at least AVX, often AVX2.
  - Many major projects have already dropped support for older CPUs or are discussing it ([NumPy](https://numpy.org/devdocs/release/2.4.0-notes.html#modulate-dispatched-x86-cpu-features), [TensorFlow](https://github.com/tensorflow/tensorflow/issues/19584), [PyTorch](https://dev-discuss.pytorch.org/t/dropping-avx-support-avx2-only/202), and others!).

### Benefits

- Improved performance
- Lower maintenance burden
- GPU builds are simpler

### Downsides

- Very old machines will no longer run the newest ODM builds
- Pre-2011 CPUs without AVX would need to stay on ODM 3.x

Related issues:
https://github.com/OpenDroneMap/ODM/issues/1944
https://github.com/OpenDroneMap/ODM/issues/1957
Continues from the saga of https://github.com/OpenDroneMap/ODM/pull/1958

[...]

PR #2037, opened 2026-06-24
Moving build system to pixi

### LINKED ISSUE #1936: Modernise Python setup for repo

MJohnson459:
This moves the entire ODM buildsystem to [pixi](https://pixi.prefix.dev/latest/). This allows us to use prebuild binaries for our libraries massively cutting down the build time, while also introducing a single pixi.toml file to store dependencies.

With this change, anyone should be able to build and run ODM like this:

```bash
pixi run build
pixi run odm -- --project-path ~/Projects/datasets/ Block-Michenzani   # e.g.
```

This should work on Linux (x86, aarch64), Windows, Mac. GPU support is also available by adding a `-e gpu` argument to the commands.

## Changes

- All non-forked dependencies that are on conda-forge we now use. As much as possible I have pinned the exact version we used before, although that wasn't always possible. GDAL, PDAL, Ceres, and OpenCV are the main ones.
- In order to preserve versions, there are a few .patch files introduced in SuperBuild. These are mostly lifted from upstream where available or the minimal change I could make to get it working. If we update versions, most of these would go away. I chose to have the patch files here rather than in our forks just to keep this PR consistent and stand-alone. In the future we may prefer to move these patches to our forks.
- I've updated the github actions to primarily use pixi and added a matrix to test different platforms - Linux (x86, aarch64), Windows, OSX.
- The Dockerfiles have changed substantially however the final image should be very close to what they were before.
- The GPU feature enables GPU acceleration for any library that has the option (libarrow, onnxruntime, etc). This might affect behaviour but it definitely affects the size of the installer. There is a trade-off to be made - if the performance difference is worth the size.
- I've used Ninja on both Linux and Windows to have a single build tool. This works, but I'm not sure if it has a performance impact on Windows. It should give a big speedup to iterative build times as well.

## Testing

To ensure this doesn't cause any major regressions, I've been keeping an eye on the following canaries:

1. Build time from cold - my last test had this go from 40 minutes to 7m on my machine. CI it still takes 20m to 30m depending on the platform.
2. Checked the output from two datasets, `Block-Michenzani` and `odm_data_aukerman`. I have been running these with **no additional flags** and to my untrained eye the results look pretty consistent.
3. To maintain feature parity, I've been testing on Linux and Windows with MacOS and ARM64 builds tested in CI. The GPU image I've tested on Windows both natively and in Docker.
4. The Docker image size has been mostly stable. They have increased slightly but not to a worrying degree. 

### Missing testing

While I have tested the basics, there are some areas I'm aware I haven't tested enough yet:

- [x] Various flags and arguments - i.e. non-default features
- [x] Complicated and larger datasets
- [ ] An experts view on differences on output
- [ ] Mac runtime
- [ ] Arm runtime
- [x] Runtime performance on all platforms
- [ ] Windows installer

## Future work

- Conda creates portable binaries by default which is good in one sense but will have a performance cost. We could do a pass to improve this using [optimized packages](https://prefix.dev/blog/building_cpu_optimized_packages) if its an issue.
- For every dependency we had prior, I have tried to tag that exact version in the pixi file. This works but we lose some benefits of pixi such as dependency resolution. We should go through each dependency and bump it as much as we can. We should also loosen the version range as we do this and rely on the `pixi.lock` file to lock the exact versions.
- Create and upload ODM to a [prefix.dev channel](https://prefix.dev/) - this would make it a one liner to install and run on any platform.


smathermather:
Related:
https://github.com/OpenDroneMap/oats/pull/8

MJohnson459:
The patches were trying to apply every time which as they weren't idempotent meant the build failed. Instead I added a small cmake wrapper which checks if the patch was already applied so repeated builds should now work.

spwoodcock:
Before this PR, I had never heard of `pixi`.

I'm using `uv` exclusively for application and library development in Python.

This article gives a good breakdown of the evolution of Python package management, and how to choose between the latest tools `uv` & `pixi`: https://jacobtomlinson.dev/posts/2025/python-package-managers-uv-vs-pixi/

Pretty much:
pip --> uv
conda --> pixi

Overall, seems like `pixi` is a great fit for ODM, requiring both compiled / native dependencies, C++ libs, and CUDA for GPU acceleration - kudos @MJohnson459!

smathermather:
Instructions for installing pixi on mac use curl, but that fails due to some issue with default ssl / curl / mac os x (I think -- will dig in later). So, I installed pixi using Homebrew, and all was pretty smooth until what I think is a memory issue:

```
2026-07-02 21:18:30,023 DEBUG: Undistorting image DJI_0024.JPG
2026-07-02 21:18:30,160 DEBUG: Undistorting image DJI_0033.JPG
[INFO]    running "/Users/user01/Documents/git/ODM/SuperBuild/install/bin/opensfm/bin/opensfm" export_visualsfm --points "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm"
[INFO]    Finished opensfm stage
[INFO]    Running openmvs stage
[INFO]    running "/Users/user01/Documents/git/ODM/SuperBuild/install/bin/opensfm/bin/opensfm" export_openmvs "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm"
[INFO]    Depthmap resolution set to: 500px
[INFO]    Running dense reconstruction. This might take a while.
[INFO]    Estimating depthmaps
[INFO]    No nvidia-smi detected
[INFO]    running "/Users/user01/Documents/git/ODM/SuperBuild/install/bin/OpenMVS/DensifyPointCloud" "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/scene.mvs" --resolution-level 2 --dense-config-file "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/Densify.ini" --max-resolution 2000 --min-resolution 640 --max-threads 10 --number-views-fuse 2 --sub-resolution-levels 2 --archive-type 3 -w "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/depthmaps" -v 0 --cuda-device -2
dyld[57270]: Library not loaded: @rpath/libboost_iostreams.dylib
  Referenced from: <254CEE38-96AE-3994-A254-3C5BDA650E59> /Users/user01/Documents/git/ODM/SuperBuild/install/bin/OpenMVS/DensifyPointCloud
  Reason: tried: '/libboost_iostreams.dylib' (no such file)
/bin/sh: line 1: 57270 Abort trap: 6           "/Users/user01/Documents/git/ODM/SuperBuild/install/bin/OpenMVS/DensifyPointCloud" "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/scene.mvs" --resolution-level 2 --dense-config-file "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/Densify.ini" --max-resolution 2000 --min-resolution 640 --max-threads 10 --number-views-fuse 2 --sub-resolution-levels 2 --archive-type 3 -w "/Users/user01/Desktop/odm_test/drone_dataset_brighton_beach/opensfm/undistorted/openmvs/depthmaps" -v 0 --cuda-device -2

===== Dumping Info for Geeks (developers need this to fix bugs) =====
Child returned 134
  File "/Users/user01/Documents/git/ODM/stages/odm_app.py", line 82, in execute
    self.first_stage.run()
  File "/Users/user01/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  File "/Users/user01/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  File "/Users/user01/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  [Previous line repeated 1 more time]
  File "/Users/user01/Documents/git/ODM/opendm/types.py", line 444, in run
    self.process(self.args, outputs)
  File "/Users/user01/Documents/git/ODM/stages/openmvs.py", line 121, in process
    raise e
  File "/Users/user01/Documents/git/ODM/stages/openmvs.py", line 107, in process
    run_densify()
  File "/Users/user01/Documents/git/ODM/stages/openmvs.py", line 102, in run_densify
    system.run('"%s" "%s" %s' % (context.omvs_densify_path,
  File "/Users/user01/Documents/git/ODM/opendm/system.py", line 112, in run
    raise SubprocessException("Child returned {}".format(retcode), retcode)
opendm.system.SubprocessException: Child returned 134

===== Done, human-readable information to follow... =====

[ERROR]   Uh oh! Processing stopped because of strange values in the reconstruction. This is often a sign that the input data has some issues or the software cannot deal with it. Have you followed best practices for data acquisition? See https://docs.opendronemap.org/flying/
```

MJohnson459:
So I ran the full OATS suite overnight just on my dev machine (no GPU, Linux) and the results were fairly promising at least on a pass/fail basis - 48 tests, 11 failures, (attached pic of results to preserve colours). Summarising the failures, there were 3 real bugs.

Screenshot_2026-07-03_10-42-17

Cause A - the `latest` image has a broken NumPy (6 failures). All of the tests that use `--dsm` or `--dtm` failed on the latest image. These passed on the `pixi` branch. I either fixed this when sorting out dependencies or its fixed in master but not released yet.

Cause B - codem alignment produces no `registration.json` file (3 failures). This applies to both pixi and latest. For some reason at this step we detect 0 ICP points and don't do the alignment. This seems to apply for both GeoTIFF and LAZ alignment. Note that OATS doesn't clean up properly yet so I think this should really be 4 failures as it affects both tags equally.

Cause C - `odm_boruszyn_kap` segfaults in both latest and pixi (2 failures). This seems like an outstanding bug:

```
error: no valid point-cloud for the ROI estimation
Segmentation fault (core dumped)
→ Child returned 139
```

So in summary, we have at least 2 outstanding bugs before we even look at the output files although neither of these are exclusive to the pixi change. And the latest image is broken for DEM generation.

spwoodcock:
> Instructions for installing pixi on mac use curl, but that fails due to some issue with default ssl / curl / mac os x (I think -- will dig in later). So, I installed pixi using Homebrew, and all was pretty smooth until what I think is a memory issue:
> 
> ```
> dyld[57270]: Library not loaded: @rpath/libboost_iostreams.dylib
>   Referenced from: <254CEE38-96AE-3994-A254-3C5BDA650E59> /Users/user01/Documents/git/ODM/SuperBuild/install/bin/OpenMVS/DensifyPointCloud
>   Reason: tried: '/libboost_iostreams.dylib' (no such file)
> /bin/sh: line 1: 57270 Abort trap: 6
> ```

@smathermather
Looks like a missing lib for OpenMVS:
`@rpath/libboost_iostreams.dylib`

So next step would be to find:

1. If the lib exists. Something like this:

```bash
cd /Users/user01/Documents/git/ODM
pixi shell
find .pixi -name "libboost_iostreams*.dylib"
```

2. If yes, find why it's not read. If not, find why its not installed on Mac.

MJohnson459:
> @smathermather Looks like a missing lib for OpenMVS: `@rpath/libboost_iostreams.dylib`
> 
> So next step would be to find:
> 
> 1. If the lib exists. Something like this:
> 
> ```shell
> cd /Users/user01/Documents/git/ODM
> pixi shell
> find .pixi -name "libboost_iostreams*.dylib"
> ```
> 
> 2. If yes, find why it's not read. If not, find why its not installed on Mac.

Yea this was it. I don't have a way to test on Mac easily so I was relying on the smoke test which basically just does `./run.py --help`. I've extended it a little to also check OpenMVS binaries.

As a fix, the MacOS binaries should now include the RPATH similar to how Linux does it. Please let me know if this helped.

smathermather:
> As a fix, the MacOS binaries should now include the RPATH similar to how Linux does it. Please let me know if this helped.

Further along:

```

[INFO]    Finished odm_filterpoints stage
[INFO]    Running odm_meshing stage
[INFO]    Writing ODM Mesh file in: /Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.ply
[INFO]    running "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 9 --bType 2 --linearFit 
dyld[94363]: Library not loaded: @rpath/libomp.dylib
  Referenced from: <49A3DD28-A115-34E3-A1CA-8904E5979FE4> /Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon
  Reason: tried: '/libomp.dylib' (no such file)
/bin/sh: line 1: 94363 Abort trap: 6           "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 9 --bType 2 --linearFit
[WARNING] Child returned 134
[WARNING] PoissonRecon failed with 8 threads, let's retry with 4...
[INFO]    running "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 4 --bType 2 --linearFit 
dyld[94366]: Library not loaded: @rpath/libomp.dylib
  Referenced from: <49A3DD28-A115-34E3-A1CA-8904E5979FE4> /Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon
  Reason: tried: '/libomp.dylib' (no such file)
/bin/sh: line 1: 94366 Abort trap: 6           "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 4 --bType 2 --linearFit
[WARNING] Child returned 134
[WARNING] PoissonRecon failed with 4 threads, let's retry with 2...
[INFO]    running "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 2 --bType 2 --linearFit 
dyld[94368]: Library not loaded: @rpath/libomp.dylib
  Referenced from: <49A3DD28-A115-34E3-A1CA-8904E5979FE4> /Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon
  Reason: tried: '/libomp.dylib' (no such file)
/bin/sh: line 1: 94368 Abort trap: 6           "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 2 --bType 2 --linearFit
[WARNING] Child returned 134
[WARNING] PoissonRecon failed with 2 threads, let's retry with 1...
[INFO]    running "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 1 --bType 2 --linearFit 
dyld[94370]: Library not loaded: @rpath/libomp.dylib
  Referenced from: <49A3DD28-A115-34E3-A1CA-8904E5979FE4> /Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon
  Reason: tried: '/libomp.dylib' (no such file)
/bin/sh: line 1: 94370 Abort trap: 6           "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/PoissonRecon" --in "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_filterpoints/point_cloud.ply" --out "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" --depth 11 --pointWeight 4.0 --samplesPerNode 1.0 --threads 1 --bType 2 --linearFit
[WARNING] Child returned 134
[INFO]    running "/Users/bobbobberson/Documents/git/ODM/SuperBuild/install/bin/OpenMVS/ReconstructMesh" -i "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply" -o "/Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.ply" --archive-type 3 --remove-spikes 0 --remove-spurious 20 --smooth 0 --target-face-num 400000 -v 0
13:54:27 [App     ] OpenMVS x32 v2.2.0
13:54:27 [App     ] Build date: Jul  2 2026, 21:07:54
13:54:27 [App     ] CPU:  (10 cores)
13:54:27 [App     ] RAM: 16.00GB Physical Memory 16.00GB Virtual Memory
13:54:27 [App     ] OS: Darwin 25.5.0 (arm64)
13:54:27 [App     ] Disk: 
13:54:27 [App     ] warning: no SSE compatible CPU or OS detected
13:54:27 [App     ] Command line: ReconstructMesh -i /Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.dirty.ply -o /Users/bobbobberson/Desktop/odm_test/drone_dataset_brighton_beach/odm_meshing/odm_mesh.ply --archive-type 3 --remove-spikes 0 --remove-spurious 20 --smooth 0 --target-face-num 400000 -v 0

===== Dumping Info for Geeks (developers need this to fix bugs) =====
Child returned 1
  File "/Users/bobbobberson/Documents/git/ODM/stages/odm_app.py", line 82, in execute
    self.first_stage.run()
  File "/Users/bobbobberson/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  File "/Users/bobbobberson/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  File "/Users/bobbobberson/Documents/git/ODM/opendm/types.py", line 465, in run
    self.next_stage.run(outputs)
  [Previous line repeated 3 more times]
  File "/Users/bobbobberson/Documents/git/ODM/opendm/types.py", line 444, in run
    self.process(self.args, outputs)
  File "/Users/bobbobberson/Documents/git/ODM/stages/odm_meshing.py", line 25, in process
    mesh.screened_poisson_reconstruction(tree.filtered_point_cloud,
  File "/Users/bobbobberson/Documents/git/ODM/opendm/mesh.py", line 201, in screened_poisson_reconstruction
    system.run('"{reconstructmesh}" -i "{infile}" '
  File "/Users/bobbobberson/Documents/git/ODM/opendm/system.py", line 112, in run
    raise SubprocessException("Child returned {}".format(retcode), retcode)
opendm.system.SubprocessException: Child returned 1

===== Done, human-readable information to follow... =====

[ERROR]   Uh oh! Processing stopped because of strange values in the reconstruction. This is often a sign that the input data has some issues or the software cannot deal with it. Have you followed best practices for data acquisition? See https://docs.opendronemap.org/flying/

```

```
bobbobberson% pixi shell
 WARN Encountered 1 warning while parsing the manifest:
  ⚠ the `[system-requirements]` table is deprecated in favor of virtual packages on `platforms`
    ╭─[/Users/bobbobberson/Documents/git/ODM/pixi.toml:70:1]
 69 │     
 70 │ ╭─▶ [feature.gpu.system-requirements]
 71 │ ├─▶ cuda = "12.0"
    · ╰──── declare these on the `platforms` entries instead
 72 │     
    ╰────
  help: e.g. platforms = [{ platform = "linux-64", cuda = "12" }]


(odm) bobbobberson% find .pixi -name "libomp.dylib"
.pixi/envs/default/lib/libomp.dylib
```

spwoodcock:
Oh one other thing to note: I think this PR answers the question in https://github.com/OpenDroneMap/ODM/issues/1960

If we use pre-built packages, then we don't run the build through the ODM build wrapper with `-march=nehalem` every time.

This would mean we likely drop support for very old CPUs without AVX / AVX2 (as the prebuilt packages will likely include optimised versions for newer architectures).

What do you think @MJohnson459?

MJohnson459:
@spwoodcock It's a good point, there's a lot going on in this PR so thanks for bringing it up. I'll try to explain what I've done and you can check me, I spent a bit of time trying to work this out myself.

## Conda

For the first piece of the puzzle, Conda compiles any C++ on Linux with these flags (trimmed down for clarity):

```cmake
CXXFLAGS=-std=c++17 -march=nocona -mtune=haswell -O2
```

(I will note [the documentation](https://github.com/conda-forge/conda-forge.github.io/issues/2251) on this is lacking right now)

The `-march=nocona` is circa 2004 CPU which I believe is actually even older than the `nehalem` one ODM used. This means the pre-compiled libs won't use AVX, etc.

However, unlike the previous ODM build it specifies the `-mtune` flag which allows certain optimisations for a `haswell` to be used, without breaking backwards compatibility. This combined with `-O2` means I don't really know how the performance compares - I've at least run it locally and the difference isn't noticeable but I haven't measured the performance yet (I'm extending oats to capture more info for this).

Finally, this applies to the generic `conda-forge` build farm. It's not clear to me if individual libraries can override this or how strict they are about enforcing these defaults.

## Superbuild

For the second piece, there are still a few libraries we compile because we have forks. Two of these have `-march=native` set in their cmake files which overrides anything we do - `openmvs` and `OpenPointClass`. The former we patch out in `SuperBuild/cmake/mvstexturing.patch`. The latter has a `PORTABLE_BUILD` flag we set in `SuperBuild/cmake/External-OpenPointClass.cmake`. 

I've not explicitly set an `-march` or `-mtune` for our Superbuild yet (default for gcc is normally x86-64 on ubuntu) but with `OpenPointClass` setting `nehalem` that becomes our effective base.

## Summary

The best-practice for C++/cmake libraries is to omit -`march` entirely and leave it up to the end application to supply flags. I've tried to catch the ones that don't do that and patch them where I can.

Conda sets `-march=nocona`, Superbuild sets `-march=nehalem` at most. Overall that means `nehalem` is likely our baseline, the same as it was before.

It also means we are likely missing some optimisations, but its not as clear cut how much the `-mtune` makes up for the older instruction set. I mentioned as a next step but [there is a way](https://prefix.dev/blog/building_cpu_optimized_packages) for `conda-forge` libs to be optimised for architecture but I haven't looked into it much yet. I was holding off until we have a way to measure it.

Does that all make (some) sense?

spwoodcock:
> @spwoodcock It's a good point, there's a lot going on in this PR so thanks for bringing it up. I'll try to explain what I've done and you can check me, I spent a bit of time trying to work this out myself.
> ## Conda
> 
> For the first piece of the puzzle, Conda compiles any C++ on Linux with these flags (trimmed down for clarity):
> 
> ```cmake
> CXXFLAGS=-std=c++17 -march=nocona -mtune=haswell -O2
> ```
> 
> (I will note [the documentation](https://github.com/conda-forge/conda-forge.github.io/issues/2251) on this is lacking right now)
> 
> The `-march=nocona` is circa 2004 CPU which I believe is actually even older than the `nehalem` one ODM used. This means the pre-compiled libs won't use AVX, etc.
> 
> However, unlike the previous ODM build it specifies the `-mtune` flag which allows certain optimisations for a `haswell` to be used, without breaking backwards compatibility. This combined with `-O2` means I don't really know how the performance compares - I've at least run it locally and the difference isn't noticeable but I haven't measured the performance yet (I'm extending oats to capture more info for this).
> 
> Finally, this applies to the generic `conda-forge` build farm. It's not clear to me if individual libraries can override this or how strict they are about enforcing these defaults.
> ## Superbuild
> 
> For the second piece, there are still a few libraries we compile because we have forks. Two of these have `-march=native` set in their cmake files which overrides anything we do - `openmvs` and `OpenPointClass`. The former we patch out in `SuperBuild/cmake/mvstexturing.patch`. The latter has a `PORTABLE_BUILD` flag we set in `SuperBuild/cmake/External-OpenPointClass.cmake`.
> 
> I've not explicitly set an `-march` or `-mtune` for our Superbuild yet (default for gcc is normally x86-64 on ubuntu) but with `OpenPointClass` setting `nehalem` that becomes our effective base.
> ## Summary
> 
> The best-practice for C++/cmake libraries is to omit -`march` entirely and leave it up to the end application to supply flags. I've tried to catch the ones that don't do that and patch them where I can.
> 
> Conda sets `-march=nocona`, Superbuild sets `-march=nehalem` at most. Overall that means `nehalem` is likely our baseline, the same as it was before.
> 
> It also means we are likely missing some optimisations, but its not as clear cut how much the `-mtune` makes up for the older instruction set. I mentioned as a next step but [there is a way](https://prefix.dev/blog/building_cpu_optimized_packages) for `conda-forge` libs to be optimised for architecture but I haven't looked into it much yet. I was holding off until we have a way to measure it.
> 
> Does that all make (some) sense?

It certainly does! That's much clearer 😄 Thank you - I should probably close the linked issue then, as I think it's handled well.

smathermather (approved):

MJohnson459:
Sorry the CI jobs are causing problems due to disk space but I reverted to the last full successful commit. That gives us a good baseline. I want to do a full CI review anyway but I don't want to hold this branch up. Testing of it has been good with no noticable drop in performance. It passes the full oats suite with no obvious differences to the existing master.

PR #2040, opened 2026-06-25
Fix --boundary crash on fiona 1.10 (CRSError: Invalid PROJ string: EPSG:4326)

### LINKED ISSUE #2039: --boundary crashes with CRSError: Invalid PROJ string: EPSG:4326 (fiona 1.10 crs.to_string regression in load_boundary)
## How did you install ODM?
Docker — `opendronemap/odm:gpu`

## What is the problem?
Passing any WGS84 GeoJSON polygon to `--boundary` aborts the run during the dataset stage with:

```
File "/code/opendm/boundary.py", line 67, in load_boundary
    t = transformer(CRS.from_proj4(fiona.crs.to_string(src.crs)),
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/venv/lib/python3.12/site-packages/pyproj/crs/crs.py", line 413, in from_proj4
    raise CRSError(f"Invalid PROJ string: {in_proj_string}")
pyproj.exceptions.CRSError: Invalid PROJ string: EPSG:4326
```

In `opendm/boundary.py`, `load_boundary()` reads the source CRS of the boundary with `fiona.crs.to_string(src.crs)` and feeds the result to pyproj's `CRS.from_proj4()`.

With **fiona 1.10**, `fiona.crs.to_string()` is deprecated and now returns the authority string `"EPSG:4326"` instead of a PROJ4 string. `CRS.from_proj4("EPSG:4326")` is invalid (that is an EPSG code, not a PROJ4 string), so it raises `CRSError`.

The minimal repro confirms the deprecation that drives the regression:

```
FionaDeprecationWarning: This function will be removed in version 2.0. Please use CRS.to_string() instead.
```

## How can we reproduce this? (minimal, no images needed)

```bash
docker run --rm --entrypoint /code/venv/bin/python opendronemap/odm:gpu -c "
from opendm import boundary
geo = {'type':'Feature','properties':{'name':'boundary'},
       'geometry':{'type':'Polygon','coordinates':[[[29.02,41.10],[29.03,41.10],[29.03,41.11],[29.02,41.11],[29.02,41.10]]]}}
boundary.load_boundary(geo, '+proj=utm +zone=35 +datum=WGS84 +units=m +no_defs')
"
```

This raises the exact traceback above, with no dataset required. In a normal run it is reached via `stages/dataset.py` → `boundary.load_boundary(args.boundary, reconstruction.get_proj_srs())`.

## Environment
- ODM version: **3.6.0** (also current `master`)
- Image: `opendronemap/odm:gpu`, digest `sha256:365bf752b230...e940`
- fiona: **1.10.1**
- pyproj: **3.7.1**

## Proposed fix
GeoJSON coordinates are, by spec (RFC 7946), always WGS84 (CRS84 / EPSG:4326), so the source CRS does not need to be re-parsed from the file. Replacing the deprecated, now-broken `from_proj4(fiona.crs.to_string(src.crs))` with an explicit `CRS.from_epsg(4326)` fixes it:

```diff
-            t = transformer(CRS.from_proj4(fiona.crs.to_string(src.crs)),
-                            CRS.from_proj4(reproject_to_proj4))
+            t = transformer(CRS.from_epsg(4326),
+                            CRS.from_proj4(reproject_to_proj4))
```

Verified inside the same image that this produces correct results (the polygon above maps to UTM zone 35N: `669632.9, 4551824.3`, …).

I'll open a PR with this change.

### LINKED ISSUE #2042: Add test case for boundary
Should add test that includes geojson boundary to catch regressions like https://github.com/OpenDroneMap/ODM/issues/2039.

EmirTD:
## What
`load_boundary()` re-parsed the boundary's source CRS via `CRS.from_proj4(fiona.crs.to_string(src.crs))`. On **fiona 1.10**, `fiona.crs.to_string()` is deprecated and returns the authority string `"EPSG:4326"` instead of a PROJ4 string, so pyproj's `CRS.from_proj4("EPSG:4326")` raises `CRSError: Invalid PROJ string: EPSG:4326` and any `--boundary` run aborts during the dataset stage.

GeoJSON coordinates are WGS84 by specification (RFC 7946), so the source CRS is always EPSG:4326 and does not need to be read back from the file. This replaces the deprecated, broken call with an explicit `CRS.from_epsg(4326)`.

## Related
Ref #2039 (auto-closed by the bot, but this is a genuine reproducible bug — details and root cause are there).

## Reproduction (before this PR)
```bash
docker run --rm --entrypoint /code/venv/bin/python opendronemap/odm:gpu -c "
from opendm import boundary
geo = {'type':'Feature','properties':{},'geometry':{'type':'Polygon',
       'coordinates':[[[29.02,41.10],[29.03,41.10],[29.03,41.11],[29.02,41.11],[29.02,41.10]]]}}
boundary.load_boundary(geo, '+proj=utm +zone=35 +datum=WGS84 +units=m +no_defs')
"
# -> pyproj.exceptions.CRSError: Invalid PROJ string: EPSG:4326
```

## Change
```diff
         if reproject_to_proj4 is not None:
-            t = transformer(CRS.from_proj4(fiona.crs.to_string(src.crs)),
-                            CRS.from_proj4(reproject_to_proj4))
+            t = transformer(CRS.from_epsg(4326),
+                            CRS.from_proj4(reproject_to_proj4))
             coords = [t.TransformPoint(*c)[:dimensions] for c in coords]
```

## Verification
With the patch, the same input transforms correctly to UTM zone 35N (`669632.9, 4551824.3`, …), and a full `--boundary` run completes past the dataset stage.

## Environment
ODM 3.6.0, `opendronemap/odm:gpu`, fiona 1.10.1, pyproj 3.7.1.

### DIFF
--- a/tests/test_boundary.py
+++ b/tests/test_boundary.py
@@ -77,5 +77,3 @@ class TestBoundary(unittest.TestCase):
 
 if __name__ == '__main__':
     unittest.main()
-if __name__ == '__main__':
-    unittest.main()
--- a/tests/test_boundary.py
+++ b/tests/test_boundary.py
@@ -32,6 +32,85 @@ class TestBoundary(unittest.TestCase):
         # Regression test for https://github.com/OpenDroneMap/ODM/issues/2039
         # On fiona >= 1.10, fiona.crs.to_string() returns "EPSG:4326" instead of a
         # PROJ4 string, which made CRS.from_proj4() raise
+        # "CRSError: Invalid PROJ string: EPSG:4326". load_boundary must reproject
+        # a WGS84 GeoJSON boundary without raising.
+        utm15n = '+proj=utm +zone=15 +datum=WGS84 +units=m +no_defs'
+        coords = load_boundary(self.boundary, utm15n)
+        self.assertEqual(len(coords), 5)
+        # Reprojected coordinates are in meters (UTM), not degrees
+        for x, y in coords:
+            self.assertGreater(abs(x), 1000)
+            self.assertGreater(abs(y), 1000)
+
+    def test_load_boundary_honors_non_wgs84_source_crs(self):
+        # GeoJSON may declare a CRS other than WGS84 (allowed by RFC 7946 and
+        # written in practice by tools like QGIS). load_boundary must reproject
+        # from the file's declared CRS, not assume EPSG:4326.
+        boundary_utm = {
+            'type': 'FeatureCollection',
+            'crs': {'type': 'name',
+                    'properties': {'name': 'urn:ogc:def:crs:EPSG::32635'}},
+            'features': [{
+                'type': 'Feature',
+                'properties': {},
+                'geometry': {
+                    'type': 'Polygon',
+                    'coordinates': [[
+                        [500000, 4540000],
+                        [501000, 4540000],
+                        [501000, 4541000],
+                        [500000, 4541000],
+                        [500000, 4540000],
+                    ]]
+                }
+            }]
+        }
+        wgs84 = '+proj=longlat +datum=WGS84 +no_defs'
+        coords = load_boundary(boundary_utm, wgs84)
+        self.assertEqual(len(coords), 5)
+        # EPSG:32635 (UTM 35N) easting 500000 / northing ~4540000 is ~27E, ~41N.
+        # If the source CRS were wrongly assumed to be WGS84, these would not land here.
+        for lon, lat in coords:
+            self.assertTrue(26 < lon < 28, "unexpected lon: %s" % lon)
+            self.assertTrue(40 < lat < 42, "unexpected lat: %s" % lat)
+
+
+if __name__ == '__main__':
+    unittest.main()
+import unittest
+
+from opendm.boundary import load_boundary
+
+
+class TestBoundary(unittest.TestCase):
+    def setUp(self):
+        # A simple WGS84 (EPSG:4326) GeoJSON polygon
+        self.boundary = {
+            'type': 'Feature',
+            'properties': {'name': 'boundary'},
+            'geometry': {
+                'type': 'Polygon',
+                'coordinates': [[
+                    [-91.99544, 46.84260],
+                    [-91.99417, 46.84260],
+                    [-91.99417, 46.84337],
+                    [-91.99544, 46.84337],
+                    [-91.99544, 46.84260],
+                ]]
+            }
+        }
+
+    def test_load_boundary_without_reprojection(self):
+        # Without a target CRS the coordinates are returned unchanged
+        coords = load_boundary(self.boundary)
+        self.assertEqual(len(coords), 5)
+        self.assertAlmostEqual(coords[0][0], -91.99544, places=5)
+        self.assertAlmostEqual(coords[0][1], 46.84260, places=5)
+
+    def test_load_boundary_reprojects_to_utm(self):
+        # Regression test for https://github.com/OpenDroneMap/ODM/issues/2039
+        # On fiona >= 1.10, fiona.crs.to_string() returns "EPSG:4326" instead of a
+        # PROJ4 string, which made CRS.from_proj4() raise
         # "CRSError: Invalid PROJ string: EPSG:4326". GeoJSON is always WGS84
         # (RFC 7946), so load_boundary must reproject without raising.
         utm15n = '+proj=utm +zone=15 +datum=WGS84 +units=m +no_defs'
--- a/tests/test_boundary.py
+++ b/tests/test_boundary.py
@@ -77,50 +77,5 @@ class TestBoundary(unittest.TestCase):
 
 if __name__ == '__main__':
     unittest.main()
-import unittest
-
-from opendm.boundary import load_boundary
-
-
-class TestBoundary(unittest.TestCase):
-    def setUp(self):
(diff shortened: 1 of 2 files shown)


EmirTD:
Additional confirmation that this is a **fiona-version regression**, not a code change — `opendm/boundary.py` has been unchanged for ~4 years, yet the same `--boundary` call works on an older ODM and crashes on the current one:

| ODM   | fiona  | pyproj | `--boundary`                                            |
|-------|--------|--------|---------------------------------------------------------|
| 3.5.6 | 1.8.17 | 3.3.1  | ✅ works                                                 |
| 3.6.0 | 1.10.1 | 3.7.1  | ❌ `CRSError: Invalid PROJ string: EPSG:4326`            |

In fiona 1.8.x, `fiona.crs.to_string(src.crs)` returned a PROJ4 string that `CRS.from_proj4()` accepts. In fiona 1.10 it returns the authority string `"EPSG:4326"` (and emits a `FionaDeprecationWarning` pointing to `CRS.to_string()`), which is not a valid PROJ4 string — hence the crash.

Since GeoJSON coordinates are always WGS84 by spec (RFC 7946), reading the source CRS back from the file is unnecessary, and `CRS.from_epsg(4326)` is both correct and version-independent.

EmirTD:
Added a regression test in `tests/test_boundary.py` covering this exact case
(`load_boundary` with a GeoJSON boundary + a UTM reprojection target). It fails
on the current code with `CRSError: Invalid PROJ string: EPSG:4326` and passes
with this fix — also addresses the request in #9.

smathermather:
> GeoJSON coordinates are WGS84 by specification (RFC 7946), so the source CRS is always EPSG:4326 and does not need to be read back from the file. This replaces the deprecated, broken call with an explicit `CRS.from_epsg(4326)`.

I would double check the spec. I'm afraid we cannot assume CRS of WGS84. The draft spec officially allowed it:
https://geojson.org/geojson-spec.html#coordinate-reference-system-objects

And the final spec, strictly speaking, allows for it:
https://www.rfc-editor.org/info/rfc7946/#section-4

Moreover, in practice, many prominent tools, like QGIS.org, will allow one to write to an alternative CRS.

If this were a new feature, I would be really comfortable with requiring WGS84 CRS. But as this is an existing feature with possible use cases, we unfortunately probably need to continue to support other CRSs.

smathermather:
> Added a regression test in `tests/test_boundary.py` covering this exact case
> (`load_boundary` with a GeoJSON boundary + a UTM reprojection target). It fails
> on the current code with `CRSError: Invalid PROJ string: EPSG:4326` and passes
> with this fix — also addresses the request in #9.

Excellent! I have migrated that issue here so it can close on merge. It does make more sense here as a proper regression test.

EmirTD:
> > GeoJSON coordinates are WGS84 by specification (RFC 7946), so the source CRS is always EPSG:4326 and does not need to be read back from the file. This replaces the deprecated, broken call with an explicit `CRS.from_epsg(4326)`.
> 
> I would double check the spec. I'm afraid we cannot assume CRS of WGS84. The draft spec officially allowed it: https://geojson.org/geojson-spec.html#coordinate-reference-system-objects
> 
> And the final spec, strictly speaking, allows for it: https://www.rfc-editor.org/info/rfc7946/#section-4
> 
> Moreover, in practice, many prominent tools, like QGIS.org, will allow one to write to an alternative CRS.
> 
> If this were a new feature, I would be really comfortable with requiring WGS84 CRS. But as this is an existing feature with possible use cases, we unfortunately probably need to continue to support other CRSs.

You're right, we shouldn't assume WGS 84. I updated the fix to keep reading
the source CRS from the file, just via a fiona-1.10-compatible call:

    t = transformer(CRS.from_user_input(src.crs),
                    CRS.from_proj4(reproject_to_proj4))

`CRS.from_user_input(src.crs)` takes the fiona CRS object directly, so it works for both
the WGS 84 default and an explicitly declared CRS (verified with a boundary in EPSG:32635).

I also extended `tests/test_boundary.py` with a case that loads a non-WGS 84 boundary
(EPSG:32635) and checks it reprojects from that CRS rather than assuming WGS 84.

smathermather:
Yes. That's the better route. We don't need it to be backwards compatible, so best to use the newer API.

smathermather (approved):
Code change looks great. @MJohnson459 -- can you double-check the tests? They look fine to me, but it would be better to get better eyes on them.

tests/test_boundary.py
```diff
@@ -0,0 +1,126 @@
...
+        coords = load_boundary(self.boundary, utm15n)
+        self.assertEqual(len(coords), 5)
+        # Reprojected coordinates are in meters (UTM), not degrees
+        for x, y in coords:
+            self.assertGreater(abs(x), 1000)
+            self.assertGreater(abs(y), 1000)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
MJohnson459: This seems to be a duplicate of the above?

MJohnson459 (changes requested):
Looks good apart from the duplicate test. Once that is removed I will accept.

tests/test_boundary.py
EmirTD: Removed the duplicated test block, thanks for catching it.

MJohnson459 (approved):

PR #2048, opened 2026-07-15
Fix Linux GDAL build and harden Windows publish

MJohnson459:
In preparation for a release, I am going through the different builds to try and make them reliable. This PR includes three related fixes that fix the Docker and WSL builds.

- configure.sh: expose the SuperBuild gdal-config so Fiona and rasterio build against the compiled GDAL 3.11.1 instead of failing to find gdal-config on Linux/Docker. (e.g https://github.com/OpenDroneMap/ODM/actions/runs/28655904272)
- publish-windows.yml: upload the installer artifact with if: always() and add a concurrency group so a tagged build is not superseded.
- Add RELEASE.md documenting the release process and artifact checklist.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -138,6 +138,10 @@ installpython() {
     echo "Installing Python requirements"
     cd /code
     set -e
+    # Fiona and rasterio build from source and locate GDAL through gdal-config,
+    # which SuperBuild installs under SuperBuild/install/bin.
+    export GDAL_CONFIG="${RUNPATH}/SuperBuild/install/bin/gdal-config"
+    export PATH="${RUNPATH}/SuperBuild/install/bin:$PATH"
     venv/bin/pip install -r requirements.txt --ignore-installed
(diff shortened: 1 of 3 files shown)


configure.sh
```diff
@@ -138,6 +138,10 @@ installpython() {
     echo "Installing Python requirements"
     cd /code
     set -e
+    # Fiona and rasterio build from source and locate GDAL through gdal-config,
+    # which SuperBuild installs under SuperBuild/install/bin.
+    export GDAL_CONFIG="${RUNPATH}/SuperBuild/install/bin/gdal-config"
+    export PATH="${RUNPATH}/SuperBuild/install/bin:$PATH"
```
DodgySpaniard, line 144: I could swear I dealt with this one before; maybe it was in configure.py or the Docker?

DodgySpaniard (approved):
LGTM

PR #2049, opened 2026-07-15
Add OCI image labels carrying git commit SHA to Docker images

MJohnson459:
Declare the [standard OCI labels](https://specs.opencontainers.org/image-spec/annotations/) (revision, source, version, created) in the final image stage of `portable.Dockerfile`, `gpu.Dockerfile` and `Dockerfile` via build args.

These defaulting to unknown as we can't automatically add them from within Docker but we inject the values from the publish workflows so published images can be mapped back to their source commit.

This means the generated docker images will link to the exact version that was built which will make comparisons easier.

### DIFF
--- a/.github/workflows/publish-docker-gpu.yaml
+++ b/.github/workflows/publish-docker-gpu.yaml
@@ -23,6 +23,11 @@ jobs:
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
+    - name: Compute build metadata
+      id: buildmeta
+      run: |
+        echo "date=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> "$GITHUB_OUTPUT"
+        echo "version=$(cat VERSION)" >> "$GITHUB_OUTPUT"
     - name: Build and push Docker image
       id: docker_build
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-docker-gpu.yaml
```diff
@@ -31,6 +36,10 @@ jobs:
         platforms: linux/amd64
         push: true
         no-cache: true
+        build-args: |
+          GIT_COMMIT=${{ github.sha }}
```
DodgySpaniard, line 40: sha is probably enough, but for convenience, can we also add "tags"/ commit title?

DodgySpaniard (approved):

.github/workflows/publish-docker-gpu.yaml
MJohnson459: The tag is already attached to the images, just by a different method. A published image will end up as something like `OpenDroneMap/ODM:3.6.1`. These tags are extra values so we have an unmovable reference to what sha was actually built in that. Looking closer at the OCI labels, we might want a few more of them but I don't see which one would obviously suite a commit title - https://specs.opencontainers.org/image-spec/annotations/. Could always do a custom label but I think we might be able to have everything all link together with the sha and url alone.

PR #2051, opened 2026-07-17
Replace issue triage workflow with issue templates

spwoodcock:
## The AI-assisted triage bot

- I would like to propose we remove the issue-triage workflow currently in place.
- It relies on external calls to OpenAI to do text summarisation (I think this is kindly paid by Piero currently, but eventually ODM would need to pay?).
- I think this makes sense based on the AI-assistance stance set in the contribution guide: https://github.com/OpenDroneMap/documents/blob/master/CONTRIBUTING.md#use-of-ai
- A few relevant issues were closed accidentally, but I think have been re-opened if necessary 👍 

## Issue templates

- Github supports issue templates, that prompt the user to do the right thing on submission (either visit the forums, provide test data / logs, etc).
- There is already an [issue template](https://github.com/OpenDroneMap/ODM/blob/master/docs/issue_template.md) but it's in the docs.
- I added a commit to take this issue template and make a bug report & feature request template.


smathermather:
Love it. Aiming to take a deeper look next week.

PR #2062, opened 2026-08-04
Windows build fixes

monch999:


MJohnson459:
Thanks for contributing but we can't accept entirely AI generated output. Some of the fixes I'm sure are valid but others will break Linux etc. If you can untangle these fixes you are welcome to put a PR up of them individually with a description for why they are needed.

MJohnson459:
Something to keep in mind is we will be moving the build system to pixi imminently so you can always try that for the windows build. https://github.com/OpenDroneMap/ODM/pull/2037

Issue #2067, opened 2026-08-14
Differences in irradiance metadata handling between Micasense and ODM


n-grieger:
I took a closer look at the differences in lighting we observed and commented in https://github.com/OpenDroneMap/ODM/issues/2064 using the Bürgewald subset here https://github.com/OpenDroneMap/odm_data_buergewald. While the different lightings turned out to likely be caused by the visualization settings in WebODM, we found that Micasense and ODM use different metadata fields to handle irradiance information.

The Micasense Rededge-P we were using to record the Bürgewald images wrote horizontal irradiance information into the metadata field DLS:HorizontalIrradiance. ODM, however, looks for this information in the Camera:HorizontalIrradiance field: https://github.com/OpenDroneMap/ODM/blob/47f41f1f2d09f7c15dff33dd3c15ab29c8a97d45/opendm/photo.py#L341-L343, which does not exist in our data and causes ODM to fall back to estimating this irradiance from the spectral data: https://github.com/OpenDroneMap/ODM/blob/47f41f1f2d09f7c15dff33dd3c15ab29c8a97d45/opendm/multispectral.py#L131.

Moreover, the Micasense Rededge-P saves the HorizontalIrradiance data with a different scaling (factor x100) than expected by ODM. By default, ODM assumes a scaling of 1.0: https://github.com/OpenDroneMap/ODM/blob/47f41f1f2d09f7c15dff33dd3c15ab29c8a97d45/opendm/photo.py#L729, which can be overwritten using the Camera:IrradianceScaleToSIUnits field. Micasense unfortunately does not fill this field. We could now either manually add this field with a value of 0.01 to the images recorded by the camera, or add a conditional when reading the horizontal irradiance in ODM that sets the scaling to 0.01 when it detects a Micasense camera. We would be very much interested in the second solution, but I would also be interested to hear if this issue has been observed by others with Micasense cameras to prevent any regression errors.

Saijin-Naib:
Thanks for all your investigations surrounding data from this hardware!

Incredibly helpful.

n-grieger:
I should probably also mention that one can observe these scaling issues in the band data of the orthophoto. As you can see in the screenshot below, pixel values roughly vary between 0 and 0.01 with the original metadata:
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/5a91527b-7378-4ed3-92e3-10f59c74f4c0" />

[...]

PR #2069, opened 2026-08-20
Fix horizontal irradiance handling for MicaSense cameras

n-grieger:
Fix for https://github.com/OpenDroneMap/ODM/issues/2067 adapted from https://github.com/WebODM/ODX/pull/36 and https://github.com/WebODM/ODX/pull/38


smathermather:
Thanks! Looks good.

PR #2057, opened 2026-07-31
Opensfm 370

DodgySpaniard:
Current attempt to use latest our ODM/OpenSFM:370 which is synced with OpenSfM/OpenSfM:v1 in addition to our ODM changes.

This branch is sitting on top of pixi and pixi-windows-fix includes commits from Sam. It is currently a WIP, but should contain all require.

Disclaimer: AI was used in a tool capacity to speed up research. (As usual worth mentioning that the use of this tool has been rather helpful to handle all this random build system errors and likely save many hours of stack overflow scouting).

PR #2071, opened 2026-08-25
Opensfm v110

DodgySpaniard:
- Cleaned up and picked up changes from https://github.com/OpenDroneMap/ODM/pull/2057 into a fresh master.
- Updated with OpenSfM V1.1 branch.


DodgySpaniard:
cc @smathermather

smathermather (approved):

pixi.toml
```diff
@@ -46,14 +55,23 @@ codem = "==0.26.1"
 edt = "==3.0.0"
 fiona = "==1.10.1"
 pillow-jxl-plugin = "==1.3.4"
+pyarrow = "==20.0.0"     # Must match conda's libarrow (python-pdal/GDAL) or sklearn dies on import.
```
MJohnson459, line 58: `pyarrow` seems to be on conda-forge, any reason it needs to be a pypi depedency?

pixi.toml
```diff
@@ -46,14 +55,23 @@ codem = "==0.26.1"
...
 pyodm = "==1.5.12"
 pysolar = "==0.13"
 rasterio = "==1.4.3"
 rawpy = "==0.25.0"
+rerun-sdk = "==0.27.2"   # OpenSfM (pinned to match its pyproject)
```
MJohnson459, line 63: `rerun-sdk` should be on conda-forge as well. Is this needed at runtime? If not we could move it to the build dependencies as well.

MJohnson459 (approved):
Looks good other than a couple of deps that could move to the `dependencies` section instead of `pypi-dependencies`. That makes them more robust to different platforms, allows proper version resolution, and avoids any compilation they might have.

PR #2076, opened 2026-09-10
Fix CI failures

MJohnson459:
This fixes three different failures in the current CI system:

**Publish Windows Setup (Windows GPU)** 
https://github.com/OpenDroneMap/ODM/actions/runs/34355323708/job/102478410205#step:5:2153

The failure was immediately caused by the `PATH` var being too long exceeding the windows limit for 8191 bytes. The root cause was our dependencies including two versions of MSVC and adding both to the path. We also added more to the path in `activate-msvc.bat` which I cleaned up. The path should now be well under the limit. I tested this on my personal PC.

As a bonus I also removed the cache here which was using up 11GB of space without providing much benefit.

**Native builds (arm)**
https://github.com/OpenDroneMap/ODM/actions/runs/33495602391/job/99816940025

This one is harder to see directly, but I could confirm by rerunning the build locally. The memory usage during compilation exceeded the runner limits. We can reduce the max memory usage by limiting the allowed parallelism in cmake. I tested a few values and 3 seemed the best. This capped memory at <12GB while not affecting build time.

**Native builds**
https://github.com/OpenDroneMap/ODM/actions/runs/33495602391

When one build failed, the rest were stopped. This was deliberately in place to reduce usage and avoid contention especially while the CI was unreliable. We can flip this flag so all builds finish and revisit later if we hit issues.

### DIFF
--- a/.github/workflows/native-builds.yaml
+++ b/.github/workflows/native-builds.yaml
@@ -28,6 +28,11 @@ jobs:
     - name: Install environment
       run: pixi install --locked
     - name: Build SuperBuild
+      # The SuperBuild nests ninja inside ninja. Uncapped, each level runs
+      # cores+2 jobs and ~30 compilers exhaust the 16 GB runner. cmake
+      # --build honours this at both levels, so 3 bounds the total at 9.
+      env:
+        CMAKE_BUILD_PARALLEL_LEVEL: 3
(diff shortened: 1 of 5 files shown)


smathermather (approved):
Looks great!

